PR #27, opened 2022-02-01
Update website messaging

ZJvandeWeg:
Minor edits on the messaging, hope to make more in the future as the company story becomes more focussed.

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -11,7 +11,7 @@
     <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#aa4444">
     <meta name="msapplication-TileColor" content="#00aba9">
     <meta name="theme-color" content="#ffffff">
-    <title>FlowForge | An open source low-code development platform</title>
+    <title>FlowForge | An open source low-code workflow hub for IoT and IIoT</title>
 
     {% if tags and 'posts' in tags %}
     <meta name="twitter:card" content="summary" />
@@ -57,7 +57,7 @@
         <div class="max-w-screen-xl xl:mx-auto lg:mx-24 py-8 px-10 text-white">
             <div class="flex flex-col md:flex-row justify-center md:space-x-8">
                 <div class="w-full md:w-96 mb-4 text-sm flex flex-col self-end order-last md:order-first">
-                    <div>Copyright 2021 FlowForge Inc.</div>
+                    <div>Copyright 2022 FlowForge Inc.</div>
                     <div>All Rights Reserved</div>
                     <div class="mt-2 text-xs">This site uses cookies. Click <a href="#" class="underline" onclick="initCookieConsent().showSettings()">here</a> to manage your preferences</a>.</div>
                 </div>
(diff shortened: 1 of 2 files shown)


sammachin:
This narrows our scope to IoT/IIoT? which is the opposite of the direction Node-RED has gone with its tagline over the years, it started off as IoT and is now "Event Driven Applications" 
I think we need to position FlowForge as the solution for all of the Node-RED use cases rather than a specific vertical that may already be served by other NR platforms too.
We should be capitalising on our unique selling point which is that the company is also at the heart of the node-red platform, something like "The Node-RED platform from the people behind Node-RED"

sammachin:
Although there is also some work to do to position Node-RED for those that don't know about it, thats where we should then look at verticals and use-cases in our messaging. We should have a 2 step approach, one is to "sell people on node-red for their use cases" and then to sell them that FlowForge is the best solution for running Node-RED.

ZJvandeWeg:
> Although there is also some work to do to position Node-RED for those that don't know about it, thats where we should then look at verticals and use-cases in our messaging.

Created an issue for this: https://github.com/flowforge/website/issues/28

sammachin:
`The solution for low-code integration`
Using `The` instead of `A/An` projects more confidence. 
It equally could apply to Node-RED so its setting up to demonstrate how node-red solves use-case specific problems in the solutions pages. 
I dropped the open-source part here because it feels like it confuses things, the open-source flavour forms part of our product offering in the top level menus.
Screen Shot 2022-02-01 at 1 43 59 PM

knolleary:
> Open-source low-code integration for applications, data and devices

 - having the three things (apps, data, devices) gives hooks to talk about verticals. Devices (and data) being the IoT hook.

src/_includes/layouts/base.njk
```diff
@@ -11,7 +11,7 @@
     <link rel="mask-icon" href="/safari-pinned-tab.svg" color="#aa4444">
     <meta name="msapplication-TileColor" content="#00aba9">
     <meta name="theme-color" content="#ffffff">
-    <title>FlowForge | An open source low-code development platform</title>
+    <title>FlowForge | An open source low-code workflow hub for IoT and IIoT</title>
```
ZJvandeWeg, line 14:
```suggestion
    <title>FlowForge | Open-source low-code integration for applications, data and devices</title>
```

sammachin: works for me

Steve-Mcl:
Sorry if this is too much - would rather inspire everyone instead of being shy :) 

My ideas here were to try and inspire and invoke thoughts of the possibilities that can be achieved with subtle hints (i.e. the text I added to the drawing and words like power, fast, rapid, integration etc) while appearing to be focused on integration (inclusive of IoT) but not too narrow as to limit it to IoT.

Here goes...

image


image

image


image

image

knolleary:
@Steve-Mcl some good thoughts in there. In the interests of moving things forward, we'll merge the latest proposal and we can iterate on the sub-heading text separately.

Thank you everyone. An important discussion to have - and no doubt it will continue to evolve.

knolleary:
I made a mistake. I failed to apply the proposed change before merging the commit.

Apologies to everyone - will PR the agreed wording shortly.

PR #41, opened 2022-02-17
team: Add country flag to attributes

ZJvandeWeg:
To show a bit more colour, and because it was fun to implement, the team
page now features a flag for each team member. Currently only NL and GB,
but someday hopefully more.

The allignment of the flag isn't quite right in the Y axis, and
potentially needs fixing.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -2,6 +2,7 @@ const markdownIt = require("markdown-it")
 const markdownItAnchor = require("markdown-it-anchor")
 const spacetime = require("spacetime");
 const heroGen = require("./lib/post-hero-gen.js");
+const countryFlag = require("./lib/country-flag-emoji");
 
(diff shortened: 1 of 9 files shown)


joepavitt:
I really like this idea, although not a fan of the 3d effect on the emoji's being used. I'll dig around to see if there are any 2d/flat alternatives and fit those in instead

joepavitt:
Then toying with one of these two layouts. In both cases, I actually now don't mind the 3d icon.

Screenshot 2022-02-17 at 14 40 41

Screenshot 2022-02-17 at 14 41 32

hardillb:
Second one as a preference from me (flag next to name)

knolleary:
> Second one as a preference from me (flag next to name)

Me too.

joepavitt (approved):

PR #43, opened 2022-02-21
copy: Open up the waitlist for our SaaS

ZJvandeWeg:
As we're working towards opening up our platform, let's open the
waitlist fast to gauge pull and start modeling sales etc. That in turn
migth already allow us to refine our messaging.

### DIFF
--- a/src/index.njk
+++ b/src/index.njk
@@ -3,23 +3,23 @@ layout: default
 ---
 <!--Left Col-->
 <div class="flex flex-col w-full md:w-1/2 justify-center lg:items-start overflow-y-hidden bg-white  px-3 py-2 rounded shadow-xl">
-    <h1 class="my-4 text-3xl md:text-4xl text-black-hero-heading font-medium leading-tight text-center md:text-left slide-in-bottom-h1">Open-source <span class="text-blue-hero">low-code</span> integration for applications, data and devices</h1>
+    <h1 class="my-4 text-3xl md:text-4xl text-black-hero-heading font-medium leading-tight text-center md:text-left slide-in-bottom-h1">Enterprise-ready <span class="text-red-hero">Node-RED</span> for your business.</h1>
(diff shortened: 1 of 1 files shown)


src/index.njk
```diff
@@ -3,23 +3,23 @@ layout: default
 ---
 <!--Left Col-->
 <div class="flex flex-col w-full md:w-1/2 justify-center lg:items-start overflow-y-hidden bg-white  px-3 py-2 rounded shadow-xl">
-    <h1 class="my-4 text-3xl md:text-4xl text-black-hero-heading font-medium leading-tight text-center md:text-left slide-in-bottom-h1">Open-source <span class="text-blue-hero">low-code</span> integration for applications, data and devices</h1>
+    <h1 class="my-4 text-3xl md:text-4xl text-black-hero-heading font-medium leading-tight text-center md:text-left slide-in-bottom-h1">Enterprise-ready <span class="text-red-hero">Node-RED</span> for your business.</h1>
     <p class="text-black-hero-body md:text-xl mb-8 text-center md:text-left slide-in-bottom-subtitle">
-        Give your users the power to create their own workflows in a secure, scalable, collaborative environment built around the open source <a class="text-red-hero underline" target="_blank" href="https://nodered.org">Node-RED project</a>.
+        Give your users the power to create their own workflows in a secure, scalable, collaborative environment.
```
sammachin, line 8: We don't mention that its a hosted/cloud service here anywhere

sammachin:
We should turn off RSS subscription for the waitlist buttondown account too so it doesn't show on the signup page

sammachin:
have turned off RSS for the waitlist

src/index.njk
sammachin: how about `Create your own workflows in our secure, scalable, collaborative environment` subtle change to target the user signing up directly and saying our environment implies that we are managing it.

PR #44, opened 2022-02-22
analytics: Add Google Analytics

ZJvandeWeg:
Currently we don't have any tracking at all on our website, and this
seemed like the fastest way to get started with it. Default Google
analytics script is used on each page now.

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -94,4 +94,13 @@ document.getElementById('nav-toggle').onclick = function(){
     classes.toggle("hidden");
 }
 </script>
+<!-- Global site tag (gtag.js) - Google Analytics -->
+<script async src="https://www.googletagmanager.com/gtag/js?id=G-2ZH9W82QL8"></script>
+<script>
+  window.dataLayer = window.dataLayer || [];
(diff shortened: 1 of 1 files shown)


hardillb:
I know Nick was looking at a different analytics provider that didn't track individual users and only provided aggregates for both flowforge.com and nodered.org. Let me see if I can find it before we pull the trigger on this.

hardillb:
Found it, https://plausible.io/ is being loaded in the current flowforge.com

https://github.com/flowforge/website/blob/7166c431dbaef3ae8da6602ad473eaf1361ed489/src/_includes/layouts/base.njk#L24

hardillb:
But I don't see the login details for it on 1Password, so we might have to ask Nick for them.

ZJvandeWeg:
> But I don't see the login details for it on 1Password, so we might have to ask Nick for them.

Will do. Thanks for checking!

knolleary:
I moved to Plausible on a free trial to see how it compared to GA. This includes the Node-RED site as well.

The number of page views we get on the NR site means it costs $50/month. But I think that is worth it.

I have moved the login details to the shared 1password vault.

PR #45, opened 2022-02-23
blog: Accounce flowforge cloud

ZJvandeWeg:
While a SaaS offering was eluded too, and is hinted on our home-page,
this change includes a blog post to announce it. And as call to action
includes the waitlist email registration form.


ZJvandeWeg:
@sammachin Could you proof-read? Also wonder if I explained well what our offering actually does, or do I assume too much?

sammachin (commented):
@zj I assume you have tested the embedded html for the form on the blog?

PR #46, opened 2022-02-23
Sam announce cloud

sammachin:
Some edits

### DIFF
--- a/src/blog/announcing-flowforge-cloud.md
+++ b/src/blog/announcing-flowforge-cloud.md
@@ -7,7 +7,7 @@ authors: ["zeger-jan-van-de-weg"]
 ---
 
 As an open core company, anyone is free to [download and install][install-docs]
-our Community Edition. In many cases this is a great solution, it
+our platform. In many cases this is a great solution, it
 allows for custom setups in your own environment. We know this isn't for everyone
 though, some people just want to start building with Node-RED without having to manage their servers.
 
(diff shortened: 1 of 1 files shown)


src/blog/announcing-flowforge-cloud.md
```diff
@@ -1,27 +1,24 @@
...
 authors: ["zeger-jan-van-de-weg"]
 ---
 
-As an open core company, anyone is open to [download and install][install-docs]
-our [latest release][release-v02]. In many cases this is a great solution, it
-allows for custom setups in your own environment. This isn't for everyone
-however, many use-cases are will not need custom firewall configuration, or want
-to maintain and monitor a bespoke solution for managing their Node-RED instances.
+As an open core company, anyone is free to [download and install][install-docs]
+our Community Edition. In many cases this is a great solution, it
```
ZJvandeWeg, line 10: Have discussed the 'Community Edition' before? It implies there's an 'Enterprise edition', which also doesn't exist. Think it's better to not get into the details too much? And either it's self managed like you recommended, or FlowForge managed?

```suggestion
our platform. In many cases this is a great solution, it
```

ZJvandeWeg (commented):
One minor item, but agree this is much better than my proposal. Thanks!

src/blog/announcing-flowforge-cloud.md
sammachin: yes good point really the difference here is "our hosted platform" vs "self-managed" Community vs Enterprise is a different comparison (there may even be 'enterprise' hosted at some point in the future

PR #56, opened 2022-03-17
blogs: Restructure files

ZJvandeWeg:
As the deploy works as hoped, I'm adding a bit of structure to the files
in `src/blog`. Hopefully makes it a bit more manageble.


ZJvandeWeg:
Self merging as it seems no-contentious.

knolleary:
This has changed the URL of all blog entries and broken any links people had.

PR #58, opened 2022-03-17
Revert "blogs: Restructure files"

ZJvandeWeg:
Reverts flowforge/website#56

I was right this was a decent idea, wrong to assume I could move current files. For SEO reasons, linking online, etc. Let's move things back for now.


knolleary:
Right - I agree with the principle, I don't know why I didn't start out with a more organised approach. But to do this will require adding redirects for the existing blog posts. Doable, but needs a bit more work.

PR #77, opened 2022-04-05
navigation: Sign in link to the FlowForge app

ZJvandeWeg:
Since yesterday external users might user the flowforge app. As such the
app should be more discoverable. Adding a "Sign in" item to the top
navigation should aid in that.

Closes https://github.com/flowforge/website/issues/76


joepavitt (approved):
We may want to re-style this in the form of a button, or at least distinguish this from the other links in some way, but for now, this is good.

PR #85, opened 2022-04-06
Integrate /docs into website from flowforge/docs

### LINKED ISSUE #40: Include generated platform docs on website
We maintain the docs for the core platform in the main repo here: https://github.com/flowforge/flowforge/tree/main/docs

We should have a means to incorporate those docs on the website to make them easier to browse and more accessible.

 - They should be versioned against individual releases
 - By default, users should be taken to the current stable release
 - There should be the 'latest' docs taken from the `main` branch (ie, in the under-development version)

This should be largely automated. As with the handbook, a GitHub action on the main repo could trigger an action here (but only if anything under the docs dir has changed... otherwise that's a lot of wasted free Action minutes). That would update the 'latest' docs.

Another action on the main repo could, when a new tag is applied, trigger an action that pulls in that tagged version of the docs under a suitable subdir and update an index of the versions available.

The more pragmatic approach would be to make that a manual task in the first instance once the basic scaffolding of the docs pages is in place.

joepavitt:
Closes #40

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -66,6 +66,7 @@ module.exports = function(eleventyConfig) {
     eleventyConfig.addFilter("rewriteHandbookLinks", (str) => {
         str = str.replace(/href="\.\/([^/]*?)\.md(#.*)?"/g,'href="../$1/$2"')
         str = str.replace(/href="(.*?)\.md(#.*)?"/g,'href="$1/$2"')
+        str = str.replace(/href="(.*?)README\/?"/g, 'href="$1"')
         return str;
     })
     eleventyConfig.addFilter("handbookMapOriginalPath", (str) => {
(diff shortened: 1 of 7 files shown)


joepavitt:
@knolleary @hardillb we now read version number for the docs from `package.json`, but require `jq` in order to do so. We will also need some clever GitHub action to update the `website` repo when relevant (on push of new tag? or updates to docs?)

joepavitt:
@hardillb don't know my way around the `build.yml` to know why this is tripping now? Looks fine to me?
https://github.com/flowforge/website/actions/runs/2109556154/workflow

hardillb:
It's a whitespace thing, looks like I inserted an extra space. I'll fix it

knolleary (approved):

PR #95, opened 2022-04-19
Add security.txt

hardillb:
This adds a basic minimal security.txt file (https://securitytxt.org/)

We should also decide if we should sign this with a PGP key and include a public key for people to use to encrypt security reports when they send them via email?

This file will need to be updated with a new date 1 year from today.

Fixes https://github.com/flowforge/CloudProject/issues/44


sammachin:
I've added an item to the shared calendar to update this on 3rd April next year

knolleary (approved):

PR #117, opened 2022-05-19
team: Add email to some members

ZJvandeWeg:
To quickly reach someone, an email address might be nice. Downside is
spam and cold emails in. For now I find that downside acceptable and
would like my email listed.

### DIFF
--- a/src/team.njk
+++ b/src/team.njk
@@ -23,6 +23,13 @@ nohero: true
                         {{ team[member].country | countryFlag }}
                     </span>
                     {% endif %} #}
+                    {% if team[member].email %}
+                    <span class="">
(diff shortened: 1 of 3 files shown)


src/team.njk
```diff
@@ -23,6 +23,13 @@ nohero: true
                         {{ team[member].country | countryFlag }}
                     </span>
                     {% endif %} #}
+                    {% if team[member].email %}
+                    <span class="">
+                        <a href="mailto:{{ team[member].email }}">
+                          @
```
ZJvandeWeg, line 29: This component doesn't exist I think?

joepavitt (approved):

PR #123, opened 2022-05-24
Remove CTAs for the waitlist

ZJvandeWeg:
As users can directly sign up and the waitlist is phased out, the website needs new CTA's and disallow waitlist sign ups.

### DIFF
--- a/src/product.njk
+++ b/src/product.njk
@@ -4,20 +4,7 @@ layout: layouts/page.njk
 subtitle: Team-based Low-Code Development
 description:
     <p>Start building your low-code workflows on FlowForge Cloud with your team.<p>Get started with just a few clicks.</p>
-    <form
-        action="https://buttondown.email/api/emails/embed-subscribe/flowforge-waitlist"
-        method="post"
-        target="popupwindow"
-        onsubmit="window.open('https://buttondown.email/flowforge-waitlist', 'popupwindow')"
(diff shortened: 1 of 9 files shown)


src/product.njk
```diff
@@ -4,20 +4,7 @@ layout: layouts/page.njk
...
-        class="embeddable-buttondown-form p-1 my-1 "
-    >
-        <div class="max-w-md m-auto">
-            <h5 class="mt-4 mb-2">Join the Waitlist</h5>
-            <input type="email" name="email" id="bd-email" placeholder="Enter your email" class="ff-input ff-text-input" />
-            <input type="hidden" value="1" name="embed" />
-            <button type="submit" class='ff-btn ff-btn--secondary mt-4'>Sign Up</button>
-        </div>
-    </form>
+    <a class="ff-btn ff-btn--primary" href="https://app.flowforge.com/account/create">Sign up</a>
```
ZJvandeWeg, line 7: This button turned out very wide. Can you take a look @joepavitt?

joepavitt:
It should be a teal button, as the red clashes with the dark background, but other than that, not sure what your issue is:

Screenshot 2022-05-24 at 12 19 36

joepavitt:
Just noticed I was looking on the home page, not /product. Have pushed the fix

PR #127, opened 2022-05-29
Page: Contact us

ZJvandeWeg:
It's too hard to reach us right now, and this change adds a contact us
page. The details end up in SalesForce to ease tracking.

### DIFF
--- a/src/contact-us.njk
+++ b/src/contact-us.njk
@@ -1,0 +1,66 @@
+---
+layout: layouts/page.njk
+title: Contact us
+---
+<div class="pricing-tiles">
+<div class="flex justify-between items-center mb-4 px-4 md:-mb-0 md:px-0 md:block">
+<div class="product px-6 pt-20 pb-24 md:px-0 md:pt-32 md:pb-48">
+
+<form action="https://webto.salesforce.com/servlet/servlet.WebToLead?encoding=UTF-8" method="POST">
+  <input type=hidden name="oid" value="00D8a0000028InL">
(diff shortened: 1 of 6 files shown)


joepavitt:
@ZJvandeWeg should we have an open multi-line text input as part of this form? Seems strange to just have contact details, without the ability to ask us questions, or give details on their reason for contacting us?

joepavitt:
Is there a way in SF to configure the "Thanks for submitting" page? Not a fan that the user sees this after submitting a contact us form request:

Screenshot 2022-06-01 at 10 38 42

src/contact-us.njk
```diff
@@ -0,0 +1,80 @@
...
+title: Contact us
+---
+<div>
+  <div class="flex justify-between items-center mb-4 md:px-4 md:-mb-0 md:px-0 md:block">
+    <div class="ff-bg-light product px-4 pb-24 pt-4 md:pt-12 md:p-12 md:grid grid-cols-2">
+      <img class="hidden md:w-72 md:m-auto md:block md:mt-0" src="../images/pictograms/browser_red.png" />
+
+      <form action="https://webto.salesforce.com/servlet/servlet.WebToLead?encoding=UTF-8" method="POST">
+        <input type=hidden name="oid" value="00D8a0000028InL">
+        <input type=hidden name="retURL" value="http://flowforge.com/contact-us?submit=true">
```
ZJvandeWeg, line 12: @joepavitt you should be redirected here after submitting. But let me check on the Salesforce side

joepavitt: The form is submitted to Salesforce successfully, but SF then send back a redirect to the page above

PR #129, opened 2022-06-07
contact us: Try and fix the form

ZJvandeWeg:
The form doesn't work for me right now and could impact potential
customers. After regenerating the HTML I've found this change as only
difference. While the form seems to work fine on the website, I hope
this change 'magically' fixes it.


ZJvandeWeg:
@joepavitt The semantical delta is very small here, and I'm not sure if this fixes anything sadly.

PR #139, opened 2022-06-12
pricing: Add Air Gap as feature

ZJvandeWeg:


sammachin:
Have we checked that we can do this, eg can we install a platform without any reliance on external repositories like NPM or DockerHub?

hardillb:
I believe it's to be theoretically possible (using a private docker container repository and either disabling the palette manager or running a private npm registry/catalogue) as in it has not actually been done end to end.

Their will be boot strapping things to sort (e.g. the base alpine/nodejs containers will need to be included in the repos)

PR #145, opened 2022-06-27
contact us: Add support email to the page

ZJvandeWeg:
As some folks had support questions that ended up in salesforce, let's
add the support email to the contact page to resolve that.


sammachin:
this is fine as a first iteration but I suspect having a form and then just an email address will result in people filling out the form for support still.
Would be better to have a single form with a selection box for, sales, support, etc that sends to the relevent application eg salesforce or freshdesk

sammachin (approved):

PR #152, opened 2022-07-27
add switch for sales/support

sammachin:
Adds a new field for customer to select their topic and based on this it then changes the form action between salesforce or a node-red flow (on ff cloud) that routes it to Freshdesk.
The submit button is hidden until the field is selected to make it mandatory

### DIFF
--- a/src/contact-us.njk
+++ b/src/contact-us.njk
@@ -4,7 +4,6 @@ title: Contact us
 ---
 <div>
   <div class="flex justify-between items-center mb-4 md:px-4 md:-mb-0 md:px-0 md:block">
-    <p>For support queries please email <a href="mailto:support@flowforge.com">support@flowforge.com</a>.</p>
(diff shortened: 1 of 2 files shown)


joepavitt:
Have improved the styling of the button and moved it to the top, as I feel like it's sensible to have as the first option.

Screenshot 2022-07-28 at 10 14 23

joepavitt (approved):

sammachin:
I did initially put it as the first option but then I moved it to the bottom as in my mind its better UX to let the user dump their thoughts into the form initially and then ask them to make the choices as to how that fits into the companies internals?

Improved styling is good

joepavitt:
Fairly sure the user will know upfront if they're doing a "Support Request" or "Sales Enquiry". We should probably have an "other" as a catch all?

sammachin:
Other would require additional handling and routing on the backend, right now I think we can run with this and if we see a lot of non sales or support messages arriving we can then add something else to catch those.

joepavitt:
@sammachin have all of the requests coming in so far fit into one of those two categories? If not, then we need an "other"

PR #160, opened 2022-08-15
sales: Reduce sticker price for Enterprise license

ZJvandeWeg:
As we're experimenting with the sticker price to get the right
conversation. This change reduces the price to get more input.


sammachin:
Is it worth adding the `From` label above like we've done with Cloud Edition? Just helps to frame the conversation then?

PR #173, opened 2022-09-23
pricing: Check version history

ZJvandeWeg:
With snapshots released the version history checkbox is ticked, but there's room to improve with a Git integration. That integration is added to the list.

### DIFF
--- a/src/_includes/feature_lists/cloud.njk
+++ b/src/_includes/feature_lists/cloud.njk
@@ -26,6 +26,9 @@
             label: "Fully Browser-Based",
             values: [null, 'check', 'check']
         }, {
+            label: "Version History",
+            values: ['check', 'check', 'check']
+        }, {
             label: "Inter Project Communication",
             values: [null, 'check', 'check']
         }, {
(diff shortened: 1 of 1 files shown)


src/_includes/feature_lists/cloud.njk
```diff
@@ -25,6 +25,9 @@
         }, {
             label: "Fully Browser-Based",
             values: [null, 'check', 'check']
+        }, {
+            label: "Version History",
```
sammachin, line 29:
```suggestion
            label: "Basic Version History",
```

sammachin: Better to manage expectations, especially as its in all editions leaving scope for a more advanced versioning feature (with or without Git) in future iterations

sammachin (commented):
Looks good otherwise, feel free to accept or reject that suggestion then good to merge

src/_includes/feature_lists/cloud.njk
ZJvandeWeg: No need to prepend basic in the marketing material I believe. On most features in the list we can prepend "Basic" as there's always room for improvement?

PR #185, opened 2022-10-06
product: Security statement

ZJvandeWeg:
As we've had the same type of questions a couple of times now I figured we should start documenting our security practises in a statement.


ZJvandeWeg:
@hardillb I'll merge those, I knew I'd written something a while ago, but was unable to locate it.

PR #190, opened 2022-10-10
Add Pez to the team page

Pezmc:

### DIFF
--- a/src/_data/team/pez-cuckow.json
+++ b/src/_data/team/pez-cuckow.json
@@ -1,0 +1,8 @@
+{
+    "name": "Pez Cuckow",
+    "title": "Senior Software Engineer",
+    "twitter": "Pezmc",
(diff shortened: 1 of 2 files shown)


src/_data/team/pez-cuckow.json
```diff
@@ -0,0 +1,8 @@
+{
+    "name": "Pez Cuckow",
+    "title": "Senior Software Engineer",
+    "twitter": "Pezmc",
+    "github": "Pezmc",
+    "blog": "https://www.pezcuckow.com",
+    "country": "BE"
```
Pezmc, line 7: Nationality, or current country? Chosen BE here which gives the page more flags.

ZJvandeWeg: Current country for the timezone etc.

PR #200, opened 2022-10-16
Subtle H1 messaging change

joepavitt:
before:
Screenshot 2022-10-16 at 17 52 40

after:
Screenshot 2022-10-16 at 17 58 51

Think this is more natural English and reads cleaner, and the increased line height is easier on the eye.


joepavitt:
and increase line-height:

PR #206, opened 2022-10-18
Homepage Redesign & "What is Node-RED" page

### LINKED ISSUE #184: Add product imagery/screenshots/videos to homepage
### Story

_No response_

### Description

Currently, whilst we flaunt our features & integrations, we lack visual insights into what Flowforge is and does. I always feel like companies that don't show what their product looks like have something to hide - we do not.

Even pseudo-screenshots as per https://fleetdm.com/ would work well, as it also then means we do not need to update the website every time our UI updates.

joepavitt:
Closes #184 
- Messaging targeting those that already know about Node-RED
    - with link off to "What is Node-RED" for those that need extra context.
- Calling out what FF is adding to the table/value

**Changes:**
- Primary tagline changed to `An open source platform, helping you run Node-RED in production
in a secure, scalable and team-based environment.`
- `4,000+ Integrations` tagline moved to a standalone "What is Node-RED" page. 
- Replace this section with `Run Node-RED in Production` section, with explicit callout to value added to Node-RED.


joepavitt:
**Existing Homepage:**

https://user-images.githubusercontent.com/99246719/196739851-1f74c8c2-bedc-4007-8762-a4bd0a53391b.mov


**New Homepage & "What is Node-RED?" page:**

https://user-images.githubusercontent.com/99246719/196740139-af21d843-2da7-4594-8311-e9af984ce7c0.mov

src/node-red.njk
```diff
@@ -0,0 +1,91 @@
+---
```
ZJvandeWeg, line 1: should we create a directory `src/product` and move the current product page to `index.njk`, allowing this page to fall under `product/node-red`?

joepavitt: I'm hesitant as it implies Node-RED is a product of ours?

knolleary (approved):
Looks great 🚀

PR #211, opened 2022-10-21
Pricing

joepavitt:
**Desktop:**

https://user-images.githubusercontent.com/99246719/197178596-8eff4f6c-e5da-40ac-9ee7-758d412e3ea2.mov

**Mobile:**

https://user-images.githubusercontent.com/99246719/197178790-7b13e897-6494-4d05-b939-ddc051c943f1.mov

### DIFF
--- a/src/_includes/layouts/page.njk
+++ b/src/_includes/layouts/page.njk
@@ -6,9 +6,8 @@ layout: layouts/base.njk
     {% if title %}
         {% if nohero %}
         <div class="container m-auto text-center flex mt-6 p-6 mb-6 md:max-w-4xl md:mt-12">
-        {% else %}
+        {% elif heroimg %}
         <div class="hero {{ 'hero-lg' if heroLg else ''}} container m-auto text-center flex flex-wrap mt-6 p-6 mb-32 md:flex-nowrap md:max-w-4xl md:mt-12">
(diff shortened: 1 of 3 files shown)


joepavitt:
Need to finalise the taglines for each option

joepavitt:
Latest version:

Screenshot 2022-10-21 at 11 50 57

knolleary:
A slight rework of the Community box to propose. It felt a bit unbalanced to not have a sub heading in the box... WDYT?

image

joepavitt:
I was fine with this, but then implemented and looked at it again. Now I'm not so sure. Whilst i does offer more balance, it's not part of the choice to be made by the user, which is what I'm trying to lead here.

In add "Open Source", I also feel like it then conflicts with the "self managed" side, as well, they're both self-managed.

Even though the "Community" and "Self Managed" are both self managed, the fact they're at different levels in the decision and on the screen, feels cleaner to me.

knolleary:
@joepavitt good points well made.

I see you've removed the 'on your own' text which was the only other bit I wasn't a fan of as it read a bit Billy-No-Mates.

And TIL that `K8s` is the proper capitalisation and not `K8S` as I almost proposed.

knolleary (approved):

PR #213, opened 2022-10-21
Temporarily hide "Product" page

joepavitt:
Realise this may be a contentious decision, btu I would like to _temporarily_ hide the "Product" page as an option for navigation.

PostHog is showing us that we've had no bites on CTAs on that page, and it needs a considerable re-design given the updates to the Homepage and Pricing, that, whilst I can probably get done by 1.0, I do not want to 100% commit to, and i do not want misleading content on there that isn't necessarily adding value at this stage.

https://github.com/flowforge/website/issues/210 will contain the redesign work, and my thoughts on what the Product page should become.


knolleary:
I'm not oppose to this if the page isn't adding value in its current state. But would like @ZJvandeWeg to sign off as well before we do so.

joepavitt:
Looking around at competitor and other tech startups - I am favouring the "Product" page being re-labelled as "Features". I only really see "Products" being used as a navigation header when there are multiple offerings in-play

PR #228, opened 2022-10-28
Add a /careers page which will redirect to Greenhouse

joepavitt:


joepavitt:
Merging without review to test if provides us with a URL that uses our `og:image`, whilst retaining the smaller, square logo on the Job Board.

PR #229, opened 2022-10-30
jobs: Redirect /jobs to Greenhouse board

ZJvandeWeg:
Our URL for jobs is based on an externally hosted board, which makes linking harder. I want to be able to type `flowforge.com/jobs` and be done with it. This change creates a redirect from there to the Greenhouse board.


joepavitt:
I have added a flowforge.com/careers page which does the redirect too, but happy to have both as they're interchangeable imo

joepavitt (approved):

PR #241, opened 2022-11-11
Updates to Team Page

joepavitt:
https://user-images.githubusercontent.com/99246719/201387679-ef6ef271-4fc0-4696-a4e7-7106e0881d05.mov


ZJvandeWeg:
Looks good to me, one nit might be to add a link to the jobs page. Though that's already at the bottom of the screen.

joepavitt:
> Looks good to me, one nit might be to add a link to the jobs page. Though that's already at the bottom of the screen. 

Stage two of the Team Page will be a "Working at FlowForge" section, which will include values, lin to handbook, etc. and link to job/careers page.

Cut it at this for now so it's at least better than what we have.

PR #243, opened 2022-11-17
Update hero copy

ZJvandeWeg:
In short, I'm more fan of the messaging in our seed announcement. As such I've updated the hero copy to take the same line.

### DIFF
--- a/src/index.njk
+++ b/src/index.njk
@@ -8,15 +8,15 @@ layout: default
             Node-RED for the <span class="text-teal-500">Enterprise</span>.
         </h1>
         <h3 class="mt-8 text-gray-400 text-lg font-normal">
-            An open source platform, helping you run Node-RED<br>in a secure, scalable and team-based environment.
+            Integrate your IT and OT in a secure and team-based environment
         </h3>
         <div class="flex flex-col">
           <div class="m-auto mt-6 flex flex-col gap-4 sm:flex-row items-center justify-center">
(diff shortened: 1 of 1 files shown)


joepavitt:
just flagging that you have two PRs open for the same thing here, I reviewed the other one then thought I was going crazy when I saw this too, both change the hero copy: https://github.com/flowforge/website/pull/244

joepavitt:
My feedback on the hero copy can be found in #244

ZJvandeWeg:
@joepavitt Thanks for flagging, staged a few too many changes. Should've used `git add -p`.

src/index.njk
```diff
@@ -33,7 +33,7 @@ layout: default
 <div class="w-full ff-bg-light px-6 pt-12 pb-48 sm:pt-32 sm:px-0">
     <div class="container m-auto text-center max-w-4xl">
         <div class="m-auto mt-36 mb-12 sm:mt-64 sm:mb-16 w-full text-center">
-            <h2>Run Node-RED in Production</h2>
+            <h2>What FlowForge brings to Node-RED</h2>
```
ZJvandeWeg, line 36: @joepavitt recommended the verb `add`. I'm not a fan of `add`, as I'd rather amplify and extend Node-RED. Add seems too much like Node-RED is forked IMO

knolleary (approved):

PR #245, opened 2022-11-17
analytics: Add HubSpot

ZJvandeWeg:
To analyse how our content and pages work with our targeted users tracking is added from HubSpot.

The content type feature of HubSpot is not yet implemented, and will have to wait until a future iteration.

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -114,4 +114,5 @@ document.getElementById('nav-toggle').onclick = function(){
     classes.toggle("hidden");
 }
 </script>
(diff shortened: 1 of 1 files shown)


joepavitt:
what does this give us that PH doesn't?

ZJvandeWeg:
> what does this give us that PH doesn't?
Segmentation based on buyer persona.

knolleary:
Am wary of adding multiple analytics - but taking it somewhat at face value that this is needed now. Will merge so we're capturing data asap, but would like a discussion on the different roles the tools play

joepavitt:
> Segmentation based on buyer persona.

What does that actually mean in the context of HubSpot? PostHog enables segmentation by any of the features we record, and enables the definition of "Cohorts" which feature particular traits.

How is HubSpot defining a persona?

PR #256, opened 2022-11-23
Reduce size of headshot images

joepavitt:
- Noticed that the team page can be particularly slow to load due to the size of the png images (each 1.7MB)
- Have reduced each image to 512px wide, could probably go even further, but this will do for now.


Pezmc (approved):

knolleary:
With #257 merged, is this PR still needed?

joepavitt:
#257 merged into this PR/branch, so this PR needs review & merging

Pezmc:
I'd approved above, so merging!

PR #279, opened 2022-12-09
misc: Add copy code button

ZJvandeWeg:
There's often a case on the website where a blog post adds a code block which users need copy. This change adds a button in the code block to allow the code block contents to be copied into their clipboard.

This change does add two JS libraries loaded based on a URL, which is something we might not want to do.

Further; there's a choice of styling. I've chosen the red as it has more contrast on both the white and dark code block background.

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -1,6 +1,7 @@
 const markdownIt = require("markdown-it")
 const markdownItAnchor = require("markdown-it-anchor")
 const markdownItFootnote = require("markdown-it-footnote")
+const codeClipboard = require("eleventy-plugin-code-clipboard");
 const spacetime = require("spacetime");
 const heroGen = require("./lib/post-hero-gen.js");
 const countryFlag = require("./lib/country-flag-emoji");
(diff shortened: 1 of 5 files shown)


ZJvandeWeg:
Screenshot 2022-12-09 at 14 49 57

joepavitt:
Not this only renders the copy button where a code style has been applied, e.g. `html`, `js`, so we will need to do a sweep of the docs, which seem to be missing this currently.

The syntax highlighting is also not working currently, investigating as to why.

joepavitt:
Added "tomorrow" theme for syntax highlighting, with more time, and as a low priority, we can customised these themes (other themes are available - [Default Themes](https://unpkg.com/browse/prismjs@1.20.0/themes/) | [Third Party](https://github.com/PrismJS/prism-themes)) to better suit the FlowForge design branding/palette

Screenshot 2022-12-09 at 15 37 41
Screenshot 2022-12-09 at 15 37 57
Screenshot 2022-12-09 at 15 37 49

joepavitt (approved):

PR #281, opened 2022-12-10
blog: Explain the exec nodes

ZJvandeWeg:
Recently I got the question why the exec nodes were disabled on FlowForge Cloud, but there's cases you'd not disable them for each stack. While this blog post doesn't explain that, it explains the basic knowledge I assumed others to have for my explanation.

This blog post builds a basis on which to build in follow ups, for example: why is `stdin` missing? How doe FileDescriptors work, and how does that influence the exec node? And in what cases do we need the exec nodes?


robmarcer:
@ZJvandeWeg I've made some commits, can you review? Otherwise I was happy with the PR.

ZJvandeWeg:
@robmarcer looks good to me, ready to be published when it fits in the schedule

PR #291, opened 2022-12-15
Streamline/improve the terminology on the homepage headline features

joepavitt:


src/index.njk
```diff
@@ -5,10 +5,10 @@ layout: default
 <div class="w-full px-6 pb-24 mt-12 md:px-0">
     <div class="container m-auto text-center max-w-2xl">
         <h1 class="text-gray-50 max-w-lg m-auto">
-            Node-RED for the <span class="text-teal-500">Enterprise</span>.
+            Run Node-RED in <span class="text-teal-500">Production</span>.
         </h1>
         <h3 class="mt-8 text-gray-400 text-lg font-normal">
-            Integrate your IT and OT in a secure and team-based environment
+            FlowForge is a centralized platform to manage all of your Node-RED instances. Providing version control, remote deployment to edge devices, and much more.
```
ZJvandeWeg, line 11: Disagree, these are tactical reasons. Let's keep it strategic.

joepavitt: We are getting feedback that it's still not clear what FlowForge is or does, even after looking at our website. With one customer saying we need a "single line summary of what FlowForge offers"

"Integrate your IT and OT in a secure and team-based environment" is too vague in my opinion.

Whilst I appreciate you want the strategic view, it doesn't answer what FlowForge is, which I think that tagline should achieve.

joepavitt: I'm sure we also recently discussed what our main points of value are in FF - "Secure" and "Team-Based" are not the highest value points for FF, yet it's what we lead with?

joepavitt: Given that our personas are mostly existing Node-RED users - tagline should be addressing "what does FlowForge offer me, that my existing Node-RD infra doesn't"

Our key metric at the moment is FF instances - of our existing personas - how many of them would understand what OT means? I don't think it's many.

I'm happy to have that convergence statement on the site, I just don't think we are doing ourselves justice with it as our primary tagline.

ZJvandeWeg: > Our key metric at the moment is FF instances - of our existing personas - how many of them would understand what OT means? I don't think it's many.

I think you're overfitting for your own experience though. Factories are usually split in a IT and OT team too.

joepavitt: > I think you're overfitting for your own experience though. Factories are usually split in a IT and OT team too.

Very plausible, the rest of my arguments still stand though.

ZJvandeWeg: > I'm sure we also recently discussed what our main points of value are in FF - "Secure" and "Team-Based" are not the highest value points for FF, yet it's what we lead with?

Disagree, again this is on a tactical level. _Why_ people are adopting Node-RED and _why_ they need FlowForge isn't because of "Secure" or "Team based". They don't care. They want to make sure the data the OT team accesses is integrated with their ERP. They want to track production output. They want data previously unobtainable to be useful to make their production more agile.

joepavitt: > why they need FlowForge isn't because of "Secure" or "Team based".

That's the point I was making - you're agreeing with me. I'm saying we shouldn't be leading with those terms, but currently, we do.

ZJvandeWeg: Mate, I do agree. But the current suggestion makes the same mistake than right? 😄 

So we help you scale your Node-RED deployment so you can access your data and make better decisions, faster. Would that work?

joepavitt: > But the current suggestion makes the same mistake than right?
I removed "Secure" and "Team-Based" in my proposal...

> Scale your Node-RED deployment
I like this as an opener to that tagline. Don't feel like the biggest value of FF is about accessing data and making decisions faster though? It's about reduced operational overheads, frictionless devops and deployments, team/collaboration tools & scaling Node-RED

ZJvandeWeg (approved):

PR #308, opened 2023-01-05
Enable windows

hardillb:
## Description

Replaces all the unix bash scripts with nodejs

## Checklist

    - [ ] Upgrade instructions
    - [ ] Configuration details
    - [ ] Concepts

## Labels

 - [ ] Backport needed? -> add the `backport` label
 - [ ] Includes a DB migration? -> add the `area:migration` label


joepavitt:
@iskerrett let's connect when you get online (ping me on Slack) and we can check this works okay. Some new GitHub workflow to introduce to you here too in order for you to test the PR

scripts/copy_docs.js
```diff
@@ -0,0 +1,57 @@
...
+                } else {
+                const { stdout } = await exec(`git log -1 --pretty=format:%ci ${file.name}`, {
+                    cwd: src 
+                })
+                const header = '---\n' +
+                `originPath: ${path.join(src, file.name)}\n` +
+                `updated: ${stdout}\n` +
+                `version: ${version}\n` +
+                '---\n'
+                const content = await fs.readFile(srcFile)
```
Pezmc, line 29: This doesn't include the changes from https://github.com/flowforge/website/pull/304/files#diff-87fcd60654beba05681c9de3d7c5fcf12e464f4396566688b33f4ecf48ffc45cR34

https://github.com/flowforge/website/blob/05b4534c93acd3f92fd10e36cf43660bdeb39523/scripts/copy_handbook.sh#L34

Pezmc: Same for the handbook!

Pezmc (changes requested):
I'd understood that most windows users nowadays install node as part of WSL, which comes with bash.

Is the intension of this PR to allow windows users running node directly within powershell to build the website? 

Blocking due to comment above.

joepavitt:
> I'd understood that most windows users nowadays install node as part of WSL, which comes with bash.

Whilst developers may operate that way, non-technical people coming on board to FlowForge, that want to edit the Handbook & Website, will not. They'll just go to the NodeJS website and click "Download".

scripts/copy_docs.js
hardillb: The sed line is not in the original `copy_docs.sh`, but there is an equivalent in the `copy_handbook.js` on line 29

```
const body = header + content.toString().replace('---\n---\n', '')
```

I think I've understood what the sed line does and this should be the same because it's acting on the same file.

iskerrett (approved):
Works on my machine

scripts/copy_handbook.js
```diff
@@ -0,0 +1,55 @@
...
+                } else {
+                const { stdout } = await exec(`git log -1 --pretty=format:%ci ${file.name}`, {
+                    cwd: src 
+                })
+                const header = '---\n' +
+                `originPath: ${path.join(src, file.name)}\n` +
+                `updated: ${stdout}\n` +
+                '---\n'
+                const content = await fs.readFile(srcFile)
+                const body = header + content.toString().replace('---\n---\n', '')
```
Pezmc, line 29: @hardillb @joepavitt I don't think this works the same as the SED command.

`header + content.toString().replace('---\n---\n', '')` will find and replace sets of `---` that are next to each other in the _content_ only, but it should be finding and replacing after the header and content have been concatenated.

joepavitt: This was fixed by Nick last night - thanks for spotting to Pez

PR #309, opened 2023-01-05
Handbook Navigation Groups & Optional Ordering

joepavitt:
## Description

https://user-images.githubusercontent.com/99246719/210894092-630efccd-412e-41ef-8ef2-c2fae7cebbaa.mov

- Adds top-level navigation groups to make navigation clearer.
- Adds `order` to a group in order to force "Other" to the bottom if rendered. 
    - Currently doesn't get controlled by any data at the `.md` level, would be nice to control ordering this way too, but it's excessive for now
- Removes old imports

## Related Issue(s)

Follow on to https://github.com/flowforge/website/pull/304

## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -123,7 +123,6 @@ module.exports = function(eleventyConfig) {
         nestedChildrenToArray(nav)
 
         // add functionality to group to-level items for better navigation.
-        // TODO: Not currently shown in UI, but here for future use as will improve nav
         let groups = {
             'Other': {
                 name: 'Other',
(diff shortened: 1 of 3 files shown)


.eleventy.js
```diff
@@ -126,21 +124,48 @@ module.exports = function(eleventyConfig) {
 
         // add functionality to group to-level items for better navigation.
         // TODO: Not currently shown in UI, but here for future use as will improve nav
```
Pezmc, line 126: This comment is out of date now, right?

Pezmc (approved):

.eleventy.js
joepavitt: right you are - will remove, thanks for spotting

PR #311, opened 2023-01-06
sitemap: Add full URL, not only path

ZJvandeWeg:
As in our `_data/site.json` we called URL baseURL the sitemap generating script only generated the path element. That created errors for consumers of the sitemap.

This change make sure the baseURL is used, and is prepended to the path. As such it should improve indexing.


ZJvandeWeg:
Merging myself as I than can trigger Google to reindex the sitemap

PR #312, opened 2023-01-06
Update site description & keywords for better SEO

joepavitt:
## Description

Having reviewed content in the [Google Search Console](https://search.google.com/u/1/search-console/performance/), it's clear we need improvements in our Website description and keywords to better match the terms we expect users to be searching, when we want them to find us.

Previously, `Integrate your IT and Operations in a secure, scalable and team-based environment` contained no terms that we expect users to be googling, in order to find us. This is more direct, addresses the target of users trying to run Node-RED in production, or those trying to run Node-RED on k8s, Docker or in the Cloud.

## Related Issue(s)

None

## Checklist


iskerrett (commented):
Good stuff.  We need to work on changing the title since that is key for SEO.

iskerrett (approved):

PR #343, opened 2023-01-19
Update Analytics

ZJvandeWeg:
Updates the website analytics stack in two commits. The second one adds GA, which might not be what we want. As such I'm fine with dropping that commit if we decide against it.

There's a case to be made to merge as is, see also: https://flowforgeworkspace.slack.com/archives/C02TG4WH5PC/p1674139511144249?thread_ts=1674138549.251439&cid=C02TG4WH5PC


joepavitt (approved):

joepavitt:
@iskerrett fyi - GA is now in the website

PR #345, opened 2023-01-19
Remove 404.html from site index

### LINKED ISSUE #335: SEO: Soft 404
### Current Behavior

> Page indexing issues detected in submitted URLs for flowforge.com
> 
> To the owner of flowforge.com:
> 
> Search Console has identified that your site is affected by 1 Page indexing issue(s). The following issues were found on your site.
> 
> Top Issues
> 
> Soft 404
> We recommend that you fix these issues when possible to enable the best experience and coverage in Google Search.

This is our `404.html` page. 

### Expected Behavior

Need to ensure we return HTTP 404 response too.

### Steps To Reproduce

_No response_

### Environment

_No response_

knolleary:
Fixes #335

Remove 404.html from siteindex as we don't want it indexing our 404 page and causing upset with Google Site Indexing thinking that page should have a 404 status code to match.

### DIFF
--- a/src/sitemap.njk
+++ b/src/sitemap.njk
@@ -5,10 +5,12 @@ eleventyExcludeFromCollections: true
 <?xml version="1.0" encoding="utf-8"?>
 <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
     {% for page in collections.all %}
+        {% if page.url != '/404.html' %}
         <url>
             <loc>{{ site.baseURL }}{{ page.url | url }}</loc>
             <lastmod>{{ page.date.toISOString() }}</lastmod>
             <priority>{{ page.data.sitemapPriority | default(0.6) }}</priority>
         </url>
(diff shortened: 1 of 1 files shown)


joepavitt:
would this suffice from excluding it though? The sitemap is an optional file that we include, but doesn't dictate which parts of our website get indexed, and which do not?

Seems like we in fact want to include

```html
<meta name=”robots” content=”noindex”>
```

into the `404.njk` from reading https://www.ilfusion.com/how-to-prevent-google-from-indexing-certain-web-pages

Pezmc (changes requested):
I'd second Robs suggestion of setting a no-index flag instead.

Equally, the 404 page isn't really part of our sitemap, so perhaps a no index (in case it's found from another route) AND this PR is a good combination!

joepavitt:
Closed by https://github.com/flowforge/website/pull/387

PR #353, opened 2023-01-24
352 the monthly newsletters are not showing on the blog

robmarcer:


ZJvandeWeg:
`collections.all` isn't available for `blob.njk`. That's surprising. @joepavitt If we merge it, it will break the (unfindable) newsletter page. What do you think we should do?

joepavitt:
@ZJvandeWeg is the newsletter page on this branch? I can't find it?

joepavitt:
Fixed - the blog articles are automatically tagged by eleventy using the `/src/blog/blog.json` file. If `tags` are included in the `.md` file, then Eleventy assumes that we do not want to populate the tags with the defaults, and leaves hte newly defined `tags` (in this case `newsletter`) in place.

The solution is to include explicitly both the `posts` and `newsletter` tags. This will also be useful to know for when we [introduce grouping.filtering in our blog too](https://github.com/flowforge/website/issues/303)

joepavitt (approved):

PR #375, opened 2023-02-01
Handle windows line endings when copying content from hb/docs

knolleary:
Handle Windows line endings when copying content from handbook/docs.

### DIFF
--- a/scripts/copy_docs.js
+++ b/scripts/copy_docs.js
@@ -31,7 +31,7 @@ async function copyFiles (src, dest, version) {
                     if (/^---/.test(content)) {
                         // The original file starts with yaml front-matter, so
                         // remove the double-delimter we've just introduced
-                        body = body.replace('---\n---\n', '')
+                        body = body.replace(/---\r?\n---\r?\n/s, '')
                     }
                     await fs.writeFile(destFile, body)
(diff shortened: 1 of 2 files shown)


joepavitt:
this still isn't fixing the issue unfortunately - this is what I had tried locally, but the particular file in question isn't collapsing the `---` as we'd expect

knolleary:
Has the file in question been fixed to have the expected `---` at the start?

joepavitt:
> Has the file in question been fixed to have the expected `---` at the start?

yes.

Although, looking in detail, it may be where I've mixed the line endings with my own editing:

```
---$
originalPath: marketing/webinars.md$
updated: 2023-01-31 13:45:55 +0700$
---$
---^M$
navTitle: Marketing^M$
---^M$
```

Unix line endings (\n or LF) as $ and Windows line endings (\r\n or CRLF) as ^M$.

knolleary:
I've just merged https://github.com/flowforge/handbook/pull/206 to fix the file in git.

knolleary:
Can confirm that has fixed it for me locally. A clean checkout of handbook + this branch builds correctly.

joepavitt (approved):

PR #380, opened 2023-02-02
Add posthog.identify call on registration

joepavitt:
## Description

Provides us richer usage data on conversions and returning visitors.

### DIFF
--- a/src/_includes/layouts/webinar.njk
+++ b/src/_includes/layouts/webinar.njk
@@ -45,7 +45,12 @@ layout: layouts/base.njk
                                 region: "eu1",
                                 portalId: "26586079",
                                 formId: formId,
-                                onFormSubmit: function($form) {
+                                onFormSubmit: function($form, data) {
+                                    // identify user 
+                                    const formData = new Map(data.map((obj) => [obj.name, obj.value]));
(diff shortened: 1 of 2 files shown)


src/_includes/layouts/webinar.njk
```diff
@@ -45,7 +45,12 @@ layout: layouts/base.njk
                                 region: "eu1",
                                 portalId: "26586079",
                                 formId: formId,
-                                onFormSubmit: function($form) {
+                                onFormSubmit: function($form, data) {
+                                    // identify user 
+                                    const formData = new Map(data.map((obj) => [obj.name, obj.value]));
```
Pezmc, line 50: [nb] I'm not sure exactly how HubSpot implements this callback, but there is a native [FormData](https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData) object now.

Pezmc (approved):

Pezmc (changes requested):
I think this might not be GDPR/CCPA compliant, since the user isn't explicitly agreeing for their data to be sent to PostHog, lets look into this further.

PostHogs own summary page: https://posthog.com/docs/privacy/gdpr-compliance

PR #400, opened 2023-02-07
Webinar Page Improvements

joepavitt:
## Description

- Adds Feb 2023 Webinar Tile
- Adds "X Days Away" for upcoming events
- Sorts the upcoming AMA/Webinar depending on their relative dates
- Change existing tile to `.jpg` and resize for better OpenGraph ratio

Screenshot 2023-02-07 at 13 30 58

## Related Issue(s)

- Addresses sorting issue raised here: https://github.com/flowforge/content/issues/47#issuecomment-1419335849
- Adds Webinar tile requested here: https://github.com/flowforge/content/issues/47#issuecomment-1419138464

## Checklist


.eleventy.js
```diff
@@ -81,6 +81,19 @@ module.exports = function(eleventyConfig) {
         return postDate.isAfter(spacetime.today()) || postDate.isSame(spacetime.today())
     });
 
+    eleventyConfig.addFilter('countDays', (date) => {
```
ZJvandeWeg, line 84: Do we want to use: https://momentjs.com/?

joepavitt: not sure we need to include it given that we're already able to calculate date differences effectively?

joepavitt: also `spacetime`z = 398kB, `Moment.js`z = 4MB

joepavitt:
> I don't think the countdown is working correctly since the AMA session is not tomorrow but on Thursday.

good spot, the image for the Webinar also isn't loading for me either:

Screenshot 2023-02-07 at 17 15 27

Trying to work out the problems now

Issue #401, opened 2023-02-07
Improve General Performance of Website


Pezmc:
### Description

Connected to flowforge/website#349 

The discussion in flowforge/website#347 around use of a facade for the YouTube page prompted me to run a Lighthouse report of the website (set to throttle internet speed to "Slow 4G"), there's plenty of space for improvement, particularly surrounding the number of third-party resources that are blocking the page load (most of these can be deferred).

<img width="869" alt="Screenshot 2023-02-07 at 16 22 19" src="https://user-images.githubusercontent.com/507155/217287228-c4429da9-2808-4e96-83a9-2af5946b9d04.png">

In rough order of impact it'll likely have from most to least

```[tasklist]
- [x] Defer (or remove from pages it's not in use) hubspot forms script flowforge/website#433
- [x] Defer (or remove from pages it's not in use) syntax highlighting from [prism-tomorrow.css](https://unpkg.com/prismjs@1.20.0/themes/prism-tomorrow.css)  flowforge/website#433
- [x] Defer Google Tag manager script  flowforge/website#433
- [x] Use a fascade to defer the YouTube script (see flowforge/website#347)  flowforge/website#433
- [x] Properly size screenshots (and add retina versions?) for the size they're displayed on the page https://github.com/flowforge/website/pull/446
- [ ] Set a longer cache TTL on all images (currently set at 10 minutes) - Not possible on GitHub pages
- [x] Remove unused [materialdesignicons](https://cdn.jsdelivr.net/npm/@mdi/font@6.5.95/css/materialdesignicons.min.css) icons (custom build?)  flowforge/website#433
- [x] Lossless image compression
```

Misc:

```[tasklist]
 - [x] Add alt tag to `flowforge-logo-wordmark-white.svg`
 - [x] Add alt tag to the YouTube iframe / facade image
 - [ ] Add alt tag to the image in `div.w-full > header.ff-header > nav.relative > a.ff-wordmark`
 - [x] Add lang=en to the html tag (done through b4ada313a8455a59ca2c861c7524d4d0737eef01)
 - [ ] https://github.com/flowforge/handbook/issues/217
- [x] Automate image compression - https://github.com/flowforge/website/pull/447
```

Pezmc:
Progress after #432 and #433, notable improvement: 

![Screenshot 2023-02-22 at 10 45 11](https://user-images.githubusercontent.com/507155/220582989-08c3069b-3ede-49ec-a74f-688b2989b286.png)

[...]

PR #402, opened 2023-02-07
Chore tweak pull request template

Pezmc:
This imports the pull request template from [.github/PULL_REQUEST_TEMPLATE.md](https://github.com/flowforge/.github/blob/main/.github/PULL_REQUEST_TEMPLATE.md), which means it's deviating from the standard. What are your thoughts on this @knolleary?

Two objectives from @joepavitt and I:
 - [ ] Remove the prompts that aren't relevant to the website repo
 - [ ] Explicit prompt to consider performance impact of changes (from #401)


knolleary:
Makes sense 👍🏻

joepavitt:
> Although we may need a bit more guidance to back that up. Some do's/don'ts etc. Such as 'don't add a 10Mb image when a 100kb version will do'.

Section in the Handbook under contributions maybe? A "Best Practices" guide?

Pezmc:
Let's add that to the to-dos for the website improvement work @knolleary: #401

PR #419, opened 2023-02-16
security: Mention MQTT endpoint

ZJvandeWeg:
Our security document should explicitly mention MQTT, now it does.


src/product/security.md
```diff
@@ -75,4 +75,6 @@ corporate website is hosted at [flowforge.com](https://flowforge.com). The
 application is served at [app.flowforge.com](https://app.flowforge.com), while
 the Node-RED instances are served from \*.flowforge.cloud.
 
+MQTT traffic is served from mqtt.flowforge.cloud.
```
hardillb, line 78:
```suggestion
MQTT traffic is served from mqtt.flowforge.cloud as MQTT over WebSockets.
```

PR #421, opened 2023-02-16
Fix blog image rendering

joepavitt:
## Description

Noticed that og:image tile wasn't rendering for our latest blog.

Upon investigation, it's linked to how image tiles are rendered in the blog. The blogs should require absolute path, rather than relative path image pointers in the meta data.

e.g.

```
---
image: /blog/2023/02/images/ff-r14-image.png
---
```

not 
```
---
image: ff-r14-image.png
---
```

This was missed because the first/latest blog post was rendering with the relative path, but the historical blogs were not. The `og:image` also required an absolute path and assumed images stored in the top-tier `/images` folder, which has now been removed.

Verified via socialsharepreview.com that these all now work correctly:

Screenshot 2023-02-16 at 16 25 06

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -109,8 +109,6 @@ module.exports = function(eleventyConfig) {
     })
 
     eleventyConfig.addFilter("toAbsoluteUrl", function(url) {
-        console.log(url)
-        console.log(new URL(url, site.baseURL).href)
         return new URL(url, site.baseURL).href;
     })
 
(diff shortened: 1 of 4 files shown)


.eleventy.js
```diff
@@ -109,6 +109,8 @@ module.exports = function(eleventyConfig) {
     })
 
     eleventyConfig.addFilter("toAbsoluteUrl", function(url) {
+        console.log(url)
```
ZJvandeWeg, line 112: Should be removed

ZJvandeWeg (approved):
If you manually checked, please merge after removing the logging

PR #433, opened 2023-02-21
Improve Website Performance

Pezmc:
## Description

Several enhancements and simplification to the website all with the goal of reducing page load time and improving performance.

In this PR:

 - All directly included third party javascript and CSS marked `async` (lazy load)
 - Adds a MutationObserver that  also defers any third party _injected_ scripts (e.g. posthog/hubspot/gtm) - observer needed as there's no direct control of this injection
 - Swapped external scripts to share `jsdelivr` CDN to reduce DNS lookups
 - Use a facade to load the youtube video
 - Replaced the whole of `materialdesignicons` (392 KB font + a 53.1 KB CSS file) used to load a single icon with inline CSS
 
## Related Issue(s)

For https://github.com/flowforge/website/issues/401 and #349 

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [-] Suitable unit/system level tests have been added and they pass
 - [x] Documentation has been updated

PR #436, opened 2023-02-22
force hbspt to load prior to form

joepavitt:
## Description

Fixes HubSpot forms (broken in #433) - the solution is detrimental to performance as I require a load of the JS upfront, but given we have a webinar in 3 hours, and we've just shouted on socials for people to sign up and no forms currently work, thought this was worth pushing in ASAP.

## Checklist

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -78,6 +78,10 @@
     <meta name="twitter:site" content="@flowforgeinc" />
     {% endif %}    
 
+    <!-- HubSpot -->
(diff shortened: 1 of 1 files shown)


knolleary (approved):

joepavitt:
@Pezmc fyi - if we can work out a fix for this properly that'd be grand. 

This should have been picked up as part of the dev effort, or at least the PR review before merging. Worth ensuring we have at least one PR reviewer that tests the website functionality locally, not just eyeballing the changes before approving.

Pezmc:
When click testing locally there was no problem locally, not sure why it would be different when hosted!

joepavitt:
For what it's worth - testing locally - it was breaking for me, but a definite race condition that we need to address "properly"

Issue #427, opened 2023-02-17
Reduce how long the page is blocked from responding to user input


ZJvandeWeg:
### Epic

426

### Description

**As a:** Website visitor

**I want to:** not be blocked by JS to input data

**So that:** I can move on with what I'd like to do

### Acceptance Criteria

- [ ] All effected pages on: https://app-eu1.hubspot.com/content-strategy/26586079/recommendations/domain/16160539/hostname/16711611/category/PERFORMANCE/type/REDUCE_TOTAL_BLOCKING_TIME report no issue

Pezmc:
Effectively a duplicate of #433, ironically, HubSpot is one of the main causes of page load delay.

Pezmc:
Solved as much as possible as part of #401, times are significantly reduced across the board, but not below HubSpot's (somewhat arbitrary) threshold of 200ms.
<img width="637" alt="Screenshot 2023-03-08 at 13 06 09" src="https://user-images.githubusercontent.com/507155/223709202-1d4aa7d0-8182-48ef-a3ee-0e5eeaf19a82.png">

PR #438, opened 2023-02-22
Pin mermaid to 9.4.0

Pezmc:
V10 removed the file included by the eleventy-plugin-mermaid plugin (see https://github.com/KevinGimbel/eleventy-plugin-mermaid/issues/2), shipping instead only the esm version.

This fixes Mermaid JS while the lovely @KevinGimbel cuts a new release.

## Related Issue(s)

#427 #401 

## Checklist


KevinGimbel:
The issue is fixed, you can update to version 2.0.0 which uses the ESM modules. 

Thanks for using my plugin! 🎉

PR #443, opened 2023-02-23
seo: Set canonical for all blog posts

ZJvandeWeg:
Early on, when the blog wasn't scoped to YYYY/MM yet, we published some content. That content is now redirected to the right page, but the new ones need a canonical too.

This change is just a canon to kill a fly, but it works!

The high level effect of this change, make sure search engines know what the true content page should be.

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -26,6 +26,8 @@
     <title>FlowForge &#x2022; DevOps for Node-RED</title>
     {% endif %}
 
+    <meta rel="canonical" href="{{ page.url }}">
+
     <!-- Browser Description -->
     {% if meta and meta.description %}
     <meta name="description" content="{{ meta.description }}">


src/_includes/layouts/base.njk
```diff
@@ -26,6 +26,8 @@
     <title>FlowForge &#x2022; DevOps for Node-RED</title>
     {% endif %}
 
+    <meta rel="canonical" href="{{ page.url }}">
```
Pezmc, line 29: PR title says "all blog posts", but this will impact all pages on the website, which is the intension?

ZJvandeWeg: @Pezmc I'm not aware of tricks to do so just for blogs. I don't think it matters either, it's not a bad practise this way.

But there's a few duplicate pages, only as blogs, which is why it's specifically about blogs.

Pezmc (approved):

PR #444, opened 2023-02-23
seo: Add title element everywhere

ZJvandeWeg:
Part of the SEO changes is to have a title element on every page. This has been a bit of an uphill battle given it's been generated instead of user supplied. This commit starts with providing a specific title in the frontmatter.

The handbook repo and docs in the flowforge repo will follow.

### DIFF
--- a/src/redirect.md
+++ b/src/redirect.md
@@ -27,5 +27,6 @@ permalink: "{{ redirect.from }}"
 # The "redirect" layout just has a small html header with the meta tags that do redirection.
 layout: redirect
 skipIndex: true
+title: "You're being redirected"
 ---
 (the content can be left blank; it's entirely the frontmatter doing the work here.)


joepavitt:
This isn't what I had in mind when you asked me about this. The `title` element is used in the rendering of the page too, and will mean that the pages all now render differently to how they currently do as the `nohero` layout uses the `title` in rendering - you should see this if you're testing these changes locally (which I hope that you are)

The reason I introduced `meta.title` is so that we can define the HTML title (in browser and for SEO) without the need for rendering any specific text on the page.

In the examples you've changed here, it would be better to include these manually to ensure formatting is appropriate.

In the docs and handbook, their layout template does not render the `title` so it should be fine.

ZJvandeWeg:
@joepavitt Updated to your preference.

joepavitt (approved):

PR #454, opened 2023-02-23
Explicit Third Party Script Blocking

Pezmc:
## Description

Adds support for explicitly blocking a third party included script that is not configurable (in this case Hubspot's cookie banner).

Additionally, improves the deferral of scripts to be done before they're even added to the page.

## Related Issue(s)

#401 

## Checklist


joepavitt:
Help me understand this a little better please, HS is auto-loading this cookie JS file how? Via the inclusion of the HS JS file for forms?

Pezmc:
The analytics script (https://js-eu1.hs-scripts.com/26586079.js) that is the bottom of every page injects several scripts:

 - https://js-eu1.hscollectedforms.net/collectedforms.js - Used to scrape forms
 - https://js-eu1.hs-banner.com/v2/26586079/banner.js - Used to display a cookie banner
 - https://js-eu1.hs-analytics.net/analytics/1677492300000/26586079.js - For analytics and user tracking
 - https://js-eu1.usemessages.com/conversations-embed.js - The loader for the chat window

Even when certain Hubspot features are not enabled, they're injected into the page regardless. In FlowForge's case, the form scraping, user tracking and chat are all used, but the banner script is not. 

This PR effectively blocks the cookie banner from being loaded, since it's wasted download and file parsing time on the main thread.

joepavitt:
Thanks Pez, this does relate somewhat to us needing to update our own Cookie banner though, which I thought we had an issue for, but can't find?

Pezmc:
There's a separate cookie banner included via our own script, https://github.com/flowforge/website/blob/main/src/js/cookies.js, so any changes to the cookie banner will likely happen there.

You might be thinking of this github issue: https://github.com/flowforge/admin/issues/144

knolleary (approved):

PR #460, opened 2023-02-24
Ensure past webinars show their images too

joepavitt:
## Description

**Before:**

Screenshot 2023-02-24 at 15 51 07

**After:**

Screenshot 2023-02-24 at 15 50 46


Pezmc:
Fix is also in https://github.com/flowforge/website/pull/459

Pezmc (approved):

PR #477, opened 2023-03-01
create blog post to answer the questions we missed in the webinar

robmarcer:
## Description

There were some questions which we didn't get time to answer in the webinar last week. This blog posts answers those questions.


robmarcer:
@iskerrett looks as if the image issues are now sorted so if you could review this PR that would be appreciated.

PR #502, opened 2023-03-06
Merge handbook repo

ZJvandeWeg:
## Description

Contributing to the handbook is harder than it should be, rendering it requires cloning on the right location, and managing multiple terminal windows.

Than there's also the duplication in tooling and development effort.

Merging was done by adding all git objects so log, blame, etc still work.


knolleary (commented):
Looks good. Have built locally and verified links work etc.

Cannot check the build action changes without merging and just seeing what we get.

I do note we've had one PR merged in the handbook repo today that isn't included here - https://github.com/flowforge/handbook/pull/226 - do you want to pull those commits into your branch before we merge this? We should then
1. update handbook repo readme to point users to this repo
2. archive the handbook repo to prevent new issues/prs etc

PR #520, opened 2023-03-14
misc: Remove members.json

ZJvandeWeg:
As Eleventy makes the whole directory available in a map structure this change removes the members.json file. Adding things twice makes it error prone, and it gives us no additional benefit.


src/team.njk
```diff
@@ -5,53 +5,57 @@ layout: layouts/nohero.njk
...
+            {% set member = team[name] %}
+
             <div class="team-headshot w-full max-w-xs m-auto">
-                {% set imageSrc = ["./images/team/headshot-", team[member].headshot ] | join %}
-                {% set imageAlt = ["Photo of ", team[member].name ] | join %}
+                {% set imageSrc = ["./images/team/headshot-", member.headshot ] | join %}
+                {% set imageAlt = ["Photo of ", member.name ] | join %}
+		{{ imageSrc | log }}
+		{{ imageAlt | log }}
                 {% image imageSrc, imageAlt, [256] %}
```
ZJvandeWeg, line 16: @Pezmc This line doesn't work for me now anymore, and I don't understand the why. Do you have an idea?

Pezmc: I tweaked the loop in https://github.com/flowforge/website/pull/520/commits/3b32b4f8838be87ea3a92a9c1529af0f7312dea7, which seems to have fixed your image rendering

joepavitt:
@ZJvandeWeg just noticed a consequence of this - our /team page is now ordered alphabetically, not by seniority/join date: https://flowforge.com/team/

I vote we put this back

Pezmc:
@joepavitt Should we add a "sort/rank" value to the JSON? Then order by that desc?

joepavitt:
Yeah that's sensible

joepavitt:
Although better going ascending, as it grows better

PR #523, opened 2023-03-15
Blog post about function nodes

robmarcer:
## Description

Create a blog post discussing the benefits and drawbacks of using Node-RED function nodes


src/blog/2023/03/why-should-you-use-node-red-function-nodes.md
```diff
@@ -0,0 +1,45 @@
...
+authors: ["rob-marcer"]
+tags:
+    - posts
+---
+
+Function nodes are an essential part of Node-RED. They allow you to write custom JavaScript functions that can be used in your Node-RED flows. In this blog post, I will discuss some of the benefits and drawbacks of using Function nodes in your next project.
+
+<!--more-->
+
+ 5 Benefits of using Function Nodes:
```
iskerrett, line 15:
```suggestion
 ## 5 Benefits of using Function Nodes: ##
```

src/blog/2023/03/why-should-you-use-node-red-function-nodes.md
```diff
@@ -0,0 +1,45 @@
...
+
+1. Customisation: Function nodes allow you to write custom JavaScript functions that can be tailored to your specific needs. You can create complex functions that perform a variety of tasks, the only limit is your programming skills.
```
iskerrett, line 17:
```suggestion
1. **Customisation:** Function nodes allow you to write custom JavaScript functions that can be tailored to your specific needs. You can create complex functions that perform a variety of tasks, the only limit is your programming skills.
```

src/blog/2023/03/why-should-you-use-node-red-function-nodes.md
```diff
@@ -0,0 +1,45 @@
...
+
+2. Reusability: Function nodes can be reused in multiple flows, saving you time and effort. You can create a library of custom functions that can be easily accessed and reused in different flows.
+
+3. Debugging: Function nodes provide an easy way to debug your code. You can use console.log statements to output debug information to the Node-RED debug panel, making it easier to identify and fix issues.
+
+4. Performance: Function nodes can be more performant than using multiple nodes to achieve the same result. By combining multiple tasks into a single function, you can improve performance, assuming your code is efficient.
+
+5. Flexibility: Function nodes provide a high degree of flexibility. You can use them to perform tasks that are not possible using a single, standard Node-RED node, such as complex data manipulation.
+
+5 Benefits of avoiding Function Nodes:
```
iskerrett, line 27:
```suggestion
## 5 Benefits of avoiding Function Nodes:
```

src/blog/2023/03/why-should-you-use-node-red-function-nodes.md
```diff
@@ -0,0 +1,45 @@
...
+
+1. Simplicity: Not using function nodes can make your flows simpler and easier to understand. By using standard Node-RED nodes, you can create flows that are easy to follow and maintain for both you and your team.
```
iskerrett, line 29:
```suggestion
1. **Simplicity:** Not using function nodes can make your flows simpler and easier to understand. By using standard Node-RED nodes, you can create flows that are easy to follow and maintain for both you and your team.
```

src/blog/2023/03/why-should-you-use-node-red-function-nodes.md
```diff
@@ -0,0 +1,45 @@
...
+
+2. Ease of Use: Standard Node-RED nodes are easy to use and require no programming knowledge. This makes it easier for non-technical users to create and maintain flows.
+
+3. Modularity: By using standard Node-RED nodes, you can create modular flows that can be easily modified and extended. This makes it easier to add new functionality to your flows as your needs change.
+
+4. Community Support: Standard Node-RED nodes have a large and active community, providing support and resources for users. This can make it easier to find solutions to common problems and share knowledge with others.
+
+5. Compatibility: Standard Node-RED nodes are usually compatible with all versions of Node-RED, making it easier to migrate flows between different environments.
+
+Conclusion:
```
iskerrett, line 39:
```suggestion
## Conclusion:
```

iskerrett (commented):
I added some headers and bold. I would suggest each benefit you bold since I think it will be easier for the reader to scan.

iskerrett:
Can you add a screen grab of a function node?

src/blog/2023/03/why-should-you-use-node-red-function-nodes.md
joepavitt, line 1: title is "Benefits & Drawbacks" - but then the two sections are "Benefits of using..." and "Benefits of not using...". Is this because we want to avoid the negative connotation to a node? The title feels cleaner than the section headers?

joepavitt:
@iskerrett re: function node screenshot, am thinking this for the blog tile, so that could cover it to a certain extent?

Screenshot 2023-03-15 at 15 20 10

iskerrett:
But doesn't the function node open up to show JavaScript?

joepavitt:
Yes, that's true, at which point, a gif demo would probably be more appropriate

src/blog/2023/03/why-should-you-use-node-red-function-nodes.md
robmarcer: What about...

5 reasons why you should and should not use Node-RED function nodes in your next project

robmarcer:
@iskerrett @joepavitt hopefully I've addressed all your feedback.

src/blog/2023/03/why-should-you-use-node-red-function-nodes.md
joepavitt: I actually quite like the clarity of "Benefits & Drawbacks"

The latter term isn't overly negative, it doesn't dissuade users from using it, etc.

PR #525, opened 2023-03-15
FF 1.5 release notes

iskerrett:
## Description

Draft of FF 1.5 release notes. Comments and feedback are helpful.


iskerrett:
I converted this into a draft PR but I don't think this is what I wanted to do. If someone can convert it back to a regular PR it would be appreciated.

joepavitt:
I have created https://github.com/flowforge/website/pull/526 with my proposal for this to be split into two articles, one focussed on the Terminology changes and another on the Release content (which references the other article)

joepavitt:
verified that this all works locally after the merge

ZJvandeWeg (approved):

Issue #315, opened 2023-01-08
Add E2E Testing for 404s


joepavitt:
### Description

We need to ensure that links on our website are valid. We should create a collection of tests (ideally automatically generated) that tests all `<a></a>` elements on our website (docs/handbook included) and makes sure they successfully link to a valid page.

This ensures potential customers do not get sent to invalid paths, in addition to ensuring that as our Docs and Handbook grow, and are re-reorganised at any point, we do not leave behind invalid links.

### Acceptance Criteria

- [ ] An `npm` script that can be run on the Website to run tests that validate all links included return a 200

joepavitt:
@ZJvandeWeg this is covered now with your new CI tool, right?

PR #536, opened 2023-03-22
CI testing of links and anchors

ZJvandeWeg:
Internal links are broken from time to time, this change adds CI to check before we add them.

Closes https://github.com/flowforge/website/issues/315

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,12 +13,10 @@ jobs:
       - uses: actions/checkout@v3
         with:
           repository: flowforge/flowforge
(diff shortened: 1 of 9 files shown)


ZJvandeWeg:
Bit of rubber ducking 🦆 , but this if failing as it appears I'm not moving the docs correctly in .github/workflows/test.yml.

Maybe I'm trying to be overly clever by cloning a repo to its location and then moving the contents? I think I just need to clone elsewhere, to `tmp` or so.

joepavitt:
Closes #315

PR #562, opened 2023-03-28
Switch to Netlify Hosting & Introduce A/B Testing 

### LINKED ISSUE #475: Investigate Website A/B Testing
### Story

_No response_

### Description

A/B Testing in PostHog is known as "Feature Flags" - https://posthog.com/manual/feature-flags

Given we generate static content through Eleventy, we need to investigate what we can get away with and our our site would render the Feature Flags accordingly.

This task will be scoped to exploring the options, and testing out results in PostHog, to understand what the workflow looks like.

joepavitt:
## Description

Okay, this is a bit of a big one, so will do my best to summarise what's going on:

### Infrastructure Changes:
- We will now need to run the website from Netlify instead of GitHub pages given that we are using [Netlify Edge Functions](https://docs.netlify.com/edge-functions/overview/), I have been testing this works in my own personal account and it does: https://cheerful-biscotti-428b51.netlify.app/

### Running Locally
- We now run our local server via `npx netlify dev`, which has been packaged up as part of the `npm run start` command in our `package.json`. You will need to run an `npm install` before running this, but the `npm run start` command in order to kick off the server shouldn't need to change.

### How to Implement A/B testing
- I will be writing a more detail instruction set into the README. (in-progress as part of this PR), but the TLDR is:
    1. Create a [PostHog Experiment](https://posthog.com/docs/experiments/manual)
    2. Choose content you want A/B tested in the HTML, and write as below, where `funnel-test` is the [Feature Flag](https://posthog.com/docs/feature-flags/manual) associated to the Experiment and `control`/`test` are the variants available. Experiments will _**always**_ contain a `control` value.
    
```njk

<h1 class="text-gray-50 max-w-lg m-auto">
    {% abtesting "funnel-test", "control" %}
        DevOps for Node-RED
    {% endabtesting %}
    {% abtesting "funnel-test", "test" %}
        Run Node-RED in Production
    {% endabtesting %}
</h1>
```
- Results are then available in PostHog for review:
Screenshot 2023-03-28 at 12 14 22

### Outstanding Items:
- [x] I tried and tried to make the PostHog API key driven by an environment variable, but hit many challenges:
    - We have no access to `process.env` when running via `netlify dev`
    - The alternative is to use `Deno.env.get()` ([docs](https://examples.deno.land/environment-variables)) which works fine when deployed to Netlify, but not locally.
- [ ] More efficient use of Edge Functions:
    - In our newly created `netlify.toml` we can define the path on which the Edge Functions run, we have set this to `/*`, as we want to do this for all our endpoints. The challenge here is that this also runs for every _resource_ too, e.g. images, fonts. The `toml` doesn't support negation, but an underlying `manifest.json` does. I've not been able to work out how to define our own one of these unfortunately. I have raised [forum posts](https://answers.netlify.com/t/edge-functions-negate-endpoints-in-netlify-toml/88455), without success.
- [x] More efficient reporting of PostHog Feature Flags:
   - Because on the edge, we cannot access other `npm` modules, we have to call PostHog's API directly. When I call `/decide`, which, for a given user, returns the collection of feature flags associated to them, we then have to report to PostHog the decision through a `.capture("$feature_flag_called")`. Currently, we do this _every_ page view where the {% abtesting %} is rendered. This is excessive as the flag values cannot change for a given user, once we've got them already (and stored in our cookie) we don't need to report a change/update to PostHog, as this could rack up quite a bill, quite quickly.
- [x] Update the Handbook with detail on how to setup new Experiments 
- [x] FlowForge Netlify Account
- [ ] Do we setup our own GitHub Action to push, or do we let Netlify pull (automatically setup)

## Related Issue(s)

Closes #475 

## Checklist


.eleventy.js
```diff
@@ -15,14 +16,20 @@ const markdownItAttrs = require('markdown-it-attrs');
 const spacetime = require("spacetime");
 const { minify } = require("terser");
 
+// const axios = require('axios');
```
ZJvandeWeg, line 19:
```suggestion
```

.eleventy.js
```diff
@@ -15,14 +16,20 @@ const markdownItAttrs = require('markdown-it-attrs');
...
+
 const heroGen = require("./lib/post-hero-gen.js");
 const site = require("./src/_data/site");
 
 const DEV_MODE = process.env.ELEVENTY_RUN_MODE !== "build" // i.e. serve/watch
 
 module.exports = function(eleventyConfig) {
+
+    console.log("SERVER RELOAD")
```
ZJvandeWeg, line 28:
```suggestion
```

joepavitt:
Nick and I in agreement that for now, we'll just rely on Netlify to _pull_ the GH content, rather than using an action to push.

joepavitt:
Currently blocked by GH Action not appearing to be too friendly with Edge Functions (https://github.com/nwtgck/actions-netlify/issues/962)

knolleary:
How about:
1. GH Action builds `main`, pulling in platform docs, and pushes result into a branch called `live`
2. Netlify configured to pull from the `live` branch

That gives us the best of both worlds.

joepavitt:
Okay, so we now have a full end-to-end working pipeline for A/B testing. Now just need to decide how best to merge this all and get this on flowforge.com

Our build process is as follows:

- `build.yml` workflow now checks out both `website` and `flowforge`, runs `docs` cmd to pull docs into the website folder, then adds, commits and pushes this to a newly created `live` branch we have on our website repo.
- Netlify is configured to watch the `live` branch and any updates there will automatically trigger a `netlify build` and `netlify deploy` within Netlify itself.
- With this, we also get Lighthouse reports via the Lighthouse Plugin (see https://github.com/flowforge/website/issues/576) - this is non blocking for now, but will give a very clear picture of when issues are introduced, as it also offers a timeline view in Netlify.

When running locally, first time running since you've pulled in the Netlify work, you'll be asked about setting up VS Code (if that's your editor) for Edge Functions, etc. agree to it all, it's 2 seconds and automatically configures everything you need.

joepavitt:
Closing this as work contiued in a different branch back on the original repo - new PR is #577

PR #564, opened 2023-03-28
Update ama-nodered.md

iskerrett:
New April AMA page. Please don't merge PR until I have tested.


joepavitt:
Are we keeping the time as "16:00 GMT" - given that the UK are now operating on BST?

iskerrett:
> Are we keeping the time as "16:00 GMT" - given that the UK are now operating on BST?

I would prefer if we went with CET. Germany is actually our largest audience in Europe.

joepavitt:
sure, so that'll be the same as GMT now, right? 16:00 CET?

iskerrett:
> sure, so that'll be the same as GMT now, right? 16:00 CEST

The time of the webinar is actually 17:00CET.

iskerrett:
> > The time of the webinar is actually 17:00CET.
> 
> That's correct, but Germany, et. al are all now operating on CEST, which would be 16:00

I don't understand what you are saying here. Can we please specify the time as being 17:00CET?

joepavitt:
Yes, I can make that change, but your point was most of our audience are in Germany, and they are currently operating on CEST, not CET. So whilst I can change the time advertised to 17:00 CET, the clocks in Germany at that time will show 16:00.

src/ask-me-anything/ama-nodered.md
```diff
@@ -1,9 +1,9 @@
 ---
-title: Ask Me Anything about Node-RED (March Edition)
+title: Ask Me Anything about Node-RED (April Edition)
 subtitle: Join Nick O'Leary, co-creator of Node-RED & CTO at FlowForge and Rob Marcer, Developer Educator at FlowForge, for an AMA on Node-RED
 video: P0ZfQXh-mqM
 image: /images/webinars/ama-march.jpg
-date: 2023-03-09
+date: 2023-04-13
 time: 16:00 GMT (11:00 ET)
```
ZJvandeWeg, line 7: 17:00 CEST (XX:XX ET)

iskerrett:
I got a verbal explanation from ZJ and Marian so let's go with 17:00CEST.  I do know the webinar will be at 5:00pm German time.

ZJvandeWeg (approved):

joepavitt:
Seems like the title for the March one was left as "April" - on my phone so can't make the change myself but flagging if anyone is able to.

I'll sort it first thing otherwise

PR #577, opened 2023-03-30
A/B Testing Infrastructure & Automated Netlify Workflow

### LINKED ISSUE #475: Investigate Website A/B Testing

joepavitt:
## Description

Continued work from https://github.com/flowforge/website/pull/562 <- contains a lot of detail about the implementation and process that follows.

I needed to pull content into this repo (away from my personal) so that I could run the build actions/workflows for pushing to Netlify.

Didn't make sense to merge this work back into my personal repo, to then continue with #562 so I'm just making this fresh PR.

We already have a deployed version in Netlify running from the `live` branch: https://flowforge-website.netlify.app/

Netlify Env Var currently points to PostHog's Test Project - would need to switch.

### Recommended next steps _before_ review:

- [x] Remove the use of `{% abtesting %}` from this branch for now, as only contains test experiments.
- [x] Update the Environment Variable in Netlify to point to PostHog "Production" project
- [x] Trigger a manual `build` workflow to get the latest of this branch running on Netlify

### Recommended next steps _after_ review:

- [ ] Switch the DNS Settings of https://flowforge-website.netlify.app/ to point to flowforge.com
- [x] Merge this PR into `main`
- [ ] Verify that the merge to `main` in turn triggers the new `build` workflow and deploys any new changes to the Netlify Site automatically

## Related Issue(s)

Closes #475

## Checklist


joepavitt:
Actually, thinking about it @knolleary - we could merge this _before_ doing the DNS switch, as all it would do is have the slightly more up to date version running in Netlify, merging this would not affect the existing GH Pages build given we've changed that `build.yml` workflow

joepavitt:
When testing locally, @knolleary saw an error upon first time running linked to the edge functions not being available. This was similar to an error I see sometimes when deploying on Netlify within about 20 seconds of the deploy occuring).

We were then unable to reproduced, and with fresh clones and installs on my own machine, I could not reproduce.

I asked @iskerrett to give it a go, and he too hit the same problems as Nick (which were circumvented the second time running). Whilst, for local development it's not too bad, as it just occurs the first ever time someone clones and instals with Edge Functions, it does seem to correlate to the issue I've seen on Netlify too, so this needs further investigation and fixing.

The issues went as follows:

### 1. When first running, 11ty wants to setup an `Edge Functions Environment`
Screenshot 2023-03-31 at 14 20 04

### 2. After this, the server loads up, but the website crashes
Screenshot 2023-03-31 at 14 20 42

### 3. The reported error says that netlify could not find the `generated` netlify functions
Screenshot 2023-03-31 at 14 20 22

My guess here is that because the setup occurs, netlify then doesn't return back to the list of edge functions that we have and run it's auto-generation process (where it creates output into `_generated`. Upon restarting the server, the environment has been configured for Edge Functions, and so works.

joepavitt:
just sanity checked this again with a fresh Netlify deploy, no edge function errors, but it is slow to load the first request after a deploy, but that seems linked to [how Netlify cache](https://answers.netlify.com/t/slow-connection-after-new-deploy-on-first-load-for-every-user/20625/6) rather than anything specific to Edge Functions

src/handbook/development/ab-testing.md
```diff
@@ -0,0 +1,72 @@
+---
+navTitle: A/B Testing
```
knolleary, line 2:
```suggestion
navTitle: Website A/B Testing
```

src/handbook/development/ab-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+## How we run A/B Tests
+
+We run our A/B Testing via [PostHog Experiments](https://posthog.com/docs/experiments/manual). These are easily configured and setup in PostHog, and require that you answer the above questions in [Defining a Test](#defining-a-test)
+
+### Types of PostHog Experiments
+
+#### Trend
+
+If you are purely interested in the raw number of a partcular event taking place, then you'll want a Trend Experiment, e.g:
```
knolleary, line 33:
```suggestion
If you are purely interested in the raw number of a particular event taking place, then you'll want a Trend Experiment, e.g:
```

src/handbook/development/ab-testing.md
```diff
@@ -0,0 +1,72 @@
...
+
+## A/B Testing on our Website
+
+To conduct a new Experiment:
+
+1. Create an Experiment in PostHog
+2. Define the criteria for which users you'd like involved, and the split across the variants
+![Screenshot from PostHog showing a "Trend" Experiment](../images/screenshots/ph-experiment-funnel.png "Title")
+3. Make note of the relevant Feature Flag created, and variants (e.g. `control`, `testA`, `testB`)
+4. Wherever you want to implement the varints, includes the following (equiavlanet) code:
```
knolleary, line 51:
```suggestion
4. Wherever you want to implement the variants, includes the following (equivalanet) code:
```

knolleary (approved):
Some typos but otherwise happy with this.

src/handbook/development/ab-testing.md
joepavitt: Fair suggestion - although the A/B testing isn't limited to website, _at the moment_ we only use it there.

PR #595, opened 2023-04-04
Nav Options: Resources

joepavitt:
## Description

Accidentally created https://github.com/flowforge/website/pull/566 from a branch on my own repo, this PR is the PR to merge within flowforge repo to `main`

## Related Issue(s)

Closes https://github.com/flowforge/website/issues/319

## Checklist


iskerrett:
I don't think this makes it easy for people to use this Resource menu. The menu is too long, there are menu items that aren't accessible. It really isn't the ui I was expecting. I was thinking of a floating menu similar to mongodb.  https://www.mongodb.com/

joepavitt:
> For now, just appending to the long list as it's the most mobile-view compatible option. Will then open a new issue for the re-style which will likely result in something like https://github.com/flowforge/website/issues/319#issuecomment-1483124819.

Taken from: https://github.com/flowforge/website/pull/566#issue-1645285889

iskerrett:
I don't believe this is even a viable short term option. I would prefer to just keep the menu structure as it is now. 

I do hope we use mongodb resource menu as our inspiration once we get to implementing floating menus.

PR #600, opened 2023-04-04
Update to hubspot process in the handbook

iskerrett:
## Description
Update to the hubspot process in the handbook.

### DIFF
--- a/src/handbook/marketing/hubspot.md
+++ b/src/handbook/marketing/hubspot.md
@@ -23,7 +23,7 @@ The folowing is a description of each stage.
 | Subscriber | Contact has signed up to hear more information, when it is available, about FlowForge, e.g. newsletter subscriber |
 | Lead | Contact that has converted on our website, or through some other interaction, with our organisation beyond a subscription sign up. For example, registered for a webinar, signed up for a free trial, etc.|
 | Marketing Qualified Lead | The contact has requested to talk to someone directly at FlowForge about the product/services of FlowForge, either through book a demo, contact us, or direct outreach to a FlowForge employee. Support requests are not considered MQL. |
-| Sales Qualified Lead | After the initial discussion with an MQL, the sales team will qualify if the contact is a potential customer by changing the lifecyle stage to SQL.  This status of this stage is then further qualified in the [Lead Status](#lead-status) property |
+| Sales Qualified Lead | After the initial discussion with an MQL, the sales team will qualify if the contact is a potential customer by changing the lifecycle stage to SQL. This status of this stage is then further qualified in the [Lead Status](#lead-status) property. |
 | Opportunity | A contact becomes an Opportunity once a Deal has been opened. A deal is opened once the sales qualification questions have been answered.  |
 | Customer | An active, paying, user of FlowForge. |
 | Evangelist | FlowForge will not actively use Evangelist lifecycle stage. |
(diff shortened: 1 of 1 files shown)


src/handbook/marketing/hubspot.md
```diff
@@ -14,18 +14,19 @@ we talk to to ensure no crossover in sales and conversations.
 
 #### Lifecycle Stage
 
-When defining a new contact in HubSpot, an option exists to define the "Lifecycle Stage", this section
-details your options here, and when to use the respective value:
+All contacts in Hubspot need to have a defined lifecycle. It is required the all contacts pass through each lifeycle stage at the appropriate time based on their interaction with FlowFOrge marketing and sales. Contacts should not skip a stage. If is up to the contact owner to ensure their contacts have the appropriate lifecycle stage set.
```
ZJvandeWeg, line 17: What about:

```suggestion
All contacts in HubSpot need to have a defined lifecycle. It is required the all contacts pass through each lifecycle stage at the appropriate time based on their interaction with FlowForge marketing and sales. If is up to the contact owner to ensure their contacts have the appropriate lifecycle stage set.
```

src/handbook/marketing/hubspot.md
```diff
@@ -14,18 +14,19 @@ we talk to to ensure no crossover in sales and conversations.
...
-| Lead | Contact that has converted on our website, or through some other interaction, with our organisation beyond a subscription sign up, e.g. installed FlowForge locally, or signed up for a free trial |
-| Marketing Qualified Lead | Contact that the Marketing Team has qualified as being ready for a Sales Engagement. |
-| Sales Qualified Lead | Contact that the Sales team has qualified as a potential customer. This stage is then further qualified in the [Lead Status](#lead-status) property |
-| Opportunity | Contact with which a member of FlowForge is actively negotiating a deal. |
+| Lead | Contact that has converted on our website, or through some other interaction, with our organisation beyond a subscription sign up. For example, registered for a webinar, signed up for a free trial, etc.|
+| Marketing Qualified Lead | The contact has requested to talk to someone directly at FlowForge about the product/services of FlowForge, either through book a demo, contact us, or direct outreach to a FlowForeg employee. Support requests are not considered MQL. |
```
ZJvandeWeg, line 25:
```suggestion
| Marketing Qualified Lead | The contact has requested to talk to someone directly at FlowForge about the product/services of FlowForge, either through book a demo, contact us, or direct outreach to a FlowForge employee. Support requests are not considered MQL.
```

src/handbook/marketing/hubspot.md
```diff
@@ -14,18 +14,19 @@ we talk to to ensure no crossover in sales and conversations.
...
+| Sales Qualified Lead | After the initial discussion with an MQL, the sales team will qualify if the contact is a potential customer by changing the lifecyle stage to SQL.  This status of this stage is then further qualified in the [Lead Status](#lead-status) property |
```
ZJvandeWeg, line 26:
```suggestion
| Sales Qualified Lead | After the initial discussion with an MQL, the sales team will qualify if the contact is a potential customer by changing the lifecycle stage to SQL. This status of this stage is then further qualified in the [Lead Status](#lead-status) property. |
```

ZJvandeWeg:
@iskerrett A few minor edits, let me know what you think.

src/handbook/marketing/hubspot.md
iskerrett: I would like to keep the point contacts should not skip a stage. We need to be able to report based on when a contact enters a particular stage. If contacts skip a stage this becomes impossible to do the reporting.

ZJvandeWeg: But if a contact comes in through the contact us from, they might go straight into a phase halfway the list? How do you propose to handle that?

iskerrett: It is very rare that a contact comes in halfway through the lifecycle. If it does happen, what we were doing at HiveMQ is saving and changing the lifecycle stage for each step.

If we don't, it becomes difficult to answer the question 'How many MQLs did we create in [insert month]'

PR #619, opened 2023-04-07
Add both content_viewed and content_scroll as each have their value

joepavitt:
## Description

An extra data point alongside https://github.com/flowforge/website/pull/618 which provides a different measurement technique - `content_viewed`.

Paired with `content_scrolled` - this gives valuable insight into the amount o content seen/viewed, as well as the draw of the page to the user such that they scrolled to see more.

## Checklist


joepavitt:
Pushing this directly, thoroughly tested locally and ensures we get some sample data over the weekend in order to assert it's value.

PR #620, opened 2023-04-08
Algolia Search for Handbook & Docs

joepavitt:
## Description

- Utilises the free Netlify Algolia Plugin to create an index of our site on build.
- Have then added a search bar to the docs/handbook pages, although need to refine per the todo list below

https://user-images.githubusercontent.com/99246719/230742521-97b58765-a3ab-4665-8ed9-d2a59086d21b.mov

## Still todo:

- [ ] Limit contents to _just_ handbook in the handbook search
- [ ] Improved styling ([Example in docs](https://www.algolia.com/interface-demos/3a389a38-e15f-4e33-a1e6-be468d43eb51?utm_medium=page_link&utm_source=dashboard) shows OG images in the search preview, would be great if we had similar)
- [ ] Implement for /docs too (need to run on `main` in order for docs to be included in build index)
- [ ] On focus, provide a few recommendations [as per the demo too](https://www.algolia.com/interface-demos/3a389a38-e15f-4e33-a1e6-be468d43eb51?utm_medium=page_link&utm_source=dashboard)

## Related Issue(s)

Closes #555 

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [-] Suitable unit/system level tests have been added and they pass
 - [-] Documentation has been updated

PR #629, opened 2023-04-11
Switch the search bar to use the "live" index, not "branch"

joepavitt:
## Description

Crawler ran successfully, but didn't realise Algolia would create a new index (it seems to do this by git branch), so needed to switch our search bar to point to this new index.

## Related Issue(s)

#620

## Checklist


src/_includes/layouts/base.njk
```diff
@@ -332,7 +332,7 @@ gtag('config', 'G-2ZH9W82QL8');
     appId: 'KWE0727TZS',
     apiKey: '595ee07f069991380e9c3036cfad8e5e',
     siteId: '00f8cf60-997f-4c4d-9427-a97924358648',
-    branch: '555-algolia-search',
```
ZJvandeWeg, line 335: 👍 That's a good change

PR #640, opened 2023-04-12
misc: Move roadmap under /product

ZJvandeWeg:
## Description

Clean up the root path regarding URLs and have more semantical URLs. So the roadmap is the product roadmap, and this change moves it to "/product/roadmap".

## Checklist


joepavitt:
@ZJvandeWeg fyi ^ conflicts. Think this is with Yndira's new OG

MarianRaphael (approved):
Tested it locally
Waiting for @joepavitt. I want to avoid interfering with Yndira's new OG

PR #667, opened 2023-04-20
product: Open-core vs enterprise

ZJvandeWeg:
Rereading it I've made some edits to be more precise in the handbook.

### DIFF
--- a/src/handbook/product/strategy.md
+++ b/src/handbook/product/strategy.md
@@ -101,10 +101,10 @@ See also: [Convention over Configuration on Wikipedia](https://en.wikipedia.org/
 We offer a number of ways that users can run the FlowForge product, we will
 always offer an open source version that has the core features. Features that
 offer higher business value or permit users to share across larger groups will
-be offered as part of our paid enterprise proposition. Our managed FlowForge
-offering is public and will generally include all features that are availble
-but may not offer certain features such as SSO where integration is required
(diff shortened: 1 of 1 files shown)


src/handbook/product/strategy.md
```diff
@@ -101,10 +101,10 @@ See also: [Convention over Configuration on Wikipedia](https://en.wikipedia.org/
...
-be offered as part of our paid enterprise proposition. Our managed FlowForge
-offering is public and will generally include all features that are availble
-but may not offer certain features such as SSO where integration is required
-between the platform and an enterprise. We will offer customers the ability to
-have a dedicated managed instance if thats is a deployment model they require.
+be offered as part of our paid propositions. FlowForge Cloud is public and will
+generally include all open-source features that are availble but may not offer
+certain features such as SSO where integration is required between the platform
+and an enterprise. We offer customers the ability to have a dedicated managed
+instance if thats is a deployment model they require.
```
knolleary, line 108:
```suggestion
be offered as part of our paid propositions. FlowForge Cloud is public and will
generally include all open-source features that are available but may not offer
certain features such as SSO where integration is required between the platform
and an enterprise. We offer customers the ability to have a dedicated managed
instance if that is a deployment model they require.
```

typo fixes

MarianRaphael (approved):

PR #683, opened 2023-04-25
Speed up dev mode image handling of website

Pezmc:
## Description

- Reduces cold build time (on my machine) from ~60 seconds to ~50 seconds
- Reduces all follow up builds (on my machine) from ~13 seconds to ~4.5 seconds
- Disables Netlify auto-opening web browser on first build, leading to confusing 404
    -  **Of Note:** This means developers will now need to click on the link (or open the browser) manually
- Disables clearing of caching on rebuild
    -  **Of Note:** This means that on first load, developer might be seeing the old page, manual refresh will be needed

### Possible Follow Ups

- Disable image pipeline entirely locally (speeds up first cold build) - downside possible differences between dev and prod

## Related Issue(s)

https://github.com/flowforge/website/issues/632

## Checklist

 - [na] Suitable unit/system level tests have been added and they pass
 - [na] Documentation has been updated

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -446,23 +446,20 @@ module.exports = function(eleventyConfig) {
         const imgSrc = token.attrGet('src')
         const imgAlt = token.content
         const imgTitle = token.attrGet('title')
+
+        const folderPath = env.page.inputPath
         
-        const parsedTitle = (imgTitle || '').match(
-            /^(?<skip>@skip ?)?(?<title>.*)/
-        ).groups
-
-        if (parsedTitle.skip || imgSrc.startsWith('http')) {
-            const options = { ...htmlOpts }
-            const metadata = { img: [{ url: imgSrc }] }
-            return eleventyImage.generateHTML(metadata, options)
-        }
-
         const widths = [650] // width of blog prose
         const htmlSizes = null
 
         const async = false // cannot run async inside markdown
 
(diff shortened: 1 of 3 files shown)


Pezmc:
@joepavitt As part of review please check a cold build, re-build and change based rebuild

joepavitt:
I will get to this today 👍

Pezmc:
> For the practicality of auto-opening, I think we keep this as is. All I'm doing anyway is refreshing the page or clicking the localhost link and seeing the 404 anyway until it's ready, so I think the auto-open is worth keeping.

I disagree here, I think seeing a 404 unexpectedly is confusing for team members who don't know why they're seeing that screen. A user should be waiting for the build to display the finished link, before expecting the page to load.

> I'd like to try this and see what performance improvements we get. Tried to get my head around the pipeline work and workout where to disable, but struggled to find it, so will wait on your input here.

I can do further research here, but let's please move this PR forward in the mean time.

joepavitt:
> I disagree here, I think seeing a 404 unexpectedly is confusing for team members who don't know why they're seeing that screen. I believe Ian has previously reported it as a bug. Without the auto-open, I'd expect the user to wait for the build to finish, before loading the page.

I believe @iskerrett's concern was more around the amount of time it takes to show the page, rather than the 404 itself, although please correct me if I'm wrong Ian.

PR #699, opened 2023-04-27
Update event banner with May AMA

iskerrett:
Update top banner to promote May AMA session

### DIFF
--- a/src/_includes/components/events-banner.njk
+++ b/src/_includes/components/events-banner.njk
@@ -2,7 +2,7 @@
 
 </script>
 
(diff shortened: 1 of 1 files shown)


src/_includes/components/events-banner.njk
```diff
@@ -4,10 +4,10 @@
 
 <a data-action="event-banner" href="/webinars/2023/industrial-data-node-red/"
```
joepavitt, line 5: need to update this URL to point to the AMA session

joepavitt (changes requested):

joepavitt (approved):

PR #730, opened 2023-05-10
reduce blog image widths & add ff-prose to blog post

joepavitt:
## Description

- Add `ff-prose` class to get FlowForge prose styling on the blog
- Try reducing img widths in blogs to prevent low resolutions images scaling undesirably.
- Remove the forcing of `img` to fill `picture` if it's isn't big enough (causes pixel inflation)

## Related Issue(s)

Issue first noticed in https://github.com/flowforge/website/issues/721

## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -475,7 +475,7 @@ module.exports = function(eleventyConfig) {
 
         const folderPath = env.page.inputPath
         
-        const widths = [450] // width of blog prose
+        const widths = [650] // maximum width an image can be displayed at as part of blog prose
         const htmlSizes = null
 
         const async = false // cannot run async inside markdown
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -475,7 +475,7 @@ module.exports = function(eleventyConfig) {
 
         const folderPath = env.page.inputPath
         
-        const widths = [650] // width of blog prose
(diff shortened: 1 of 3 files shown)


.eleventy.js
```diff
@@ -475,7 +475,7 @@ module.exports = function(eleventyConfig) {
 
         const folderPath = env.page.inputPath
         
-        const widths = [650] // width of blog prose
+        const widths = [450] // width of blog prose
```
Pezmc, line 478: This should stay at 650, the maximum width an image _can_ display is still 650px wide on non retina (and 1300px on retina).

Pezmc:
We should add a requirement/guideline that images/screenshots for the blog are uploaded at at least double the width they'll be displayed (max 1300px), else we lose retina support.

Pezmc (approved):

.eleventy.js
Pezmc, line 478:
```suggestion
        const widths = [650] // maximum width an image can be displayed at as part of blog prose
```

joepavitt:
> We should add a requirement/guideline that images/screenshots for the blog are uploaded at at least double the width they'll be displayed (max 1300px), else we lose retina support.

@Pezmc could you detail something in the handbook accordingly please? No immediate rush on it

PR #734, opened 2023-05-10
Upgrade Netlify cli version & re-enable edge functions

joepavitt:
## Description

Our support request with Netlify has been updated to suggest that Netlify should now support the edge functions without error'ing. 

This should be fixed in to parts:

1. Fix on Netlify side which means we no longer need the `try/catch`
2. Upgrade Netlify client to v14+ (now set to 15 in our `package.json`)

## Checklist

### DIFF
--- a/netlify/edge-functions/eleventy-edge.js
+++ b/netlify/edge-functions/eleventy-edge.js
@@ -3,17 +3,6 @@ import {
     precompiledAppData,
 } from "./_generated/eleventy-edge-app.js";
 
-import { configAsync } from 'https://deno.land/x/dotenv/mod.ts';
-
-try {
-    // when running locally, we need to call this
-    // it'll eerror in Netlify prod, but that's okay
-    await configAsync({export: true});
-} catch (err) {
-    // ignore the error on netlify Prod
-}
-
-// const POSTHOG_APIKEY = 'phc_yVWfmiJ3eiVd2iuLYJIQROuHUN65z3hkhkGvAjjaTL7'; //Deno.env.get("POSTHOG_APIKEY");
 const POSTHOG_APIKEY = Deno.env.get("POSTHOG_APIKEY");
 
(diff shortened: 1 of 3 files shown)


joepavitt:
Verifying now if a re-build causes any down time of website as it did previously. First build went very smoothly, running second as we speak.

joepavitt:
Confirmed - 0 downtime during second build - seems as though the deployment of edge functions is good to go. No A/B testing implemented yet. Will look at some small experiments soon.

netlify/edge-functions/eleventy-edge.js
```diff
@@ -167,7 +156,7 @@ export default async (request, context) => {
                         flags = getExistingFeatureFlags(context)
                     }
                     // set cookies to pass data to client PostHog for bootstrapping
-                    if (flags[flag] && flags[flag] === value) {
+                    if (flags && flags[flag] && flags[flag] === value) {
```
ZJvandeWeg, line 159: Why wouldn't line 151 be: `let flags = {}`?

ZJvandeWeg: What codePath does actually require this change?

ZJvandeWeg:
@joepavitt Split into 2 commits please next time.

joepavitt:
> Split into 2 commits please next time.

Not sure I understand why?

netlify/edge-functions/eleventy-edge.js
joepavitt: `getExistingFeatureFlags(context)` returns `null` if there are no flags. That was tripping this when I ran it locally, so put this check in as a fix. I generally prefer returning `null` if there is no value, rather than an empty object.

ZJvandeWeg:
When you want to revert one, you can just do so with Git. Not mucking about manually.

PR #776, opened 2023-05-24
values: Add Constructive Candor

ZJvandeWeg:
As part of moving the values to a new format, this value too has been updated and rewritten.

### DIFF
--- a/src/handbook/company/values.md
+++ b/src/handbook/company/values.md
@@ -41,30 +41,21 @@ Allow progress being made by anyone. Keep an agenda for meetings, write down dec
 1. Think big picture: It’s not “your” issue/bug, the whole community can have
 opinions and provide suggestions.
 
+### ⛑️ Constructive Candor
+
+1. Feedback aids growth: Tell others what they can improve.  
+1. Share feedback directly; with candor and consideration how it will be received.
+Don’t be arrogant while sharing what others have to improve; it’s about them, not you.
+1. Assume positive intent when receiving feedback. When you feel defensive it’s OK to
+ respond later.
+1. Be optimistic, treat situations as if a positive outcome is always possible.
+1. Be open and honest about your own mistakes, take ownership of them and their resolve.
+1. Appreciate another. Say thanks often, preferably in public.
+
+
 #### Be Trusting
 
 To move with high velocity, trust needs to be given before one has had the
 opportunity to earn it. People can only grow when they're able to take up
 challenging tasks and project, and are then allowed to go their own route. When
 challenged, ask for feedback and support, not for permission or approval.
-
-### Communication
-
-#### Use Candor
-
-We use truth as a superpower, even when inconvenient to ourselves. We use it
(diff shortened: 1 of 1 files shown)


src/handbook/company/values.md
```diff
@@ -41,30 +41,21 @@ Allow progress being made by anyone. Keep an agenda for meetings, write down dec
...
+1. Feedback aids growth: Tell others what they can improve.  
+1. Share feedback directly; with candor and consideration how it will be received.
+Don’t be arrogant while sharing what others have to improve; it’s about them, not you.
+1. Assume positive intent when receiving feedback. When you feel defensive it’s OK to
+ respond later.
+1. Be optimistic, treat situations as if a positive outcome is always possible.
+1. Be open and honest about your own mistakes, take ownership of them and their resolve.
+1. Appreciate another. Say thanks often, preferably in public.
+
+
```
knolleary, line 55:
```suggestion
### ⛑️ Constructive Candor

1. Feedback aids growth: it reinforces what is good and helps drive improvement where needed
1. Be considerate: feedback is about helping others to improve, not scoring points. Use care and consideration of how it will be received.
1. Assume positive intent when receiving feedback. When you feel defensive it’s OK to
 respond later.
1. Be optimistic, treat situations as if a positive outcome is always possible.
1. Be open and honest about your own mistakes, take ownership of them and their resolve.
1. Appreciate each other. Say thanks often, preferably in public.

```

I think the opening line feels a bit negative - we want feedback to reinforce good work as well as help drive improve.

ZJvandeWeg: @knolleary The only thing I'm missing in the text is the "see something, say something" mentality; I think issues should generally be brought up early. Discussing feedback when it's small is better.

How about the second line as:

> 1. Share feedback sooner rather than later: Small challenges are fast to remedy. Use care and consideration of how it will be received.

knolleary: I agree that is a point worth covering. But I think combining it with the second item is combining two different ideas.

At the risk of making it a longer list, how about we add a *new* item in the 2nd position:

> 1. Provide timely feedback: it is better to address things when they are fresh in our minds and quicker to remedy.

ZJvandeWeg:
```suggestion
### ⛑️ Constructive Candor

1. Feedback aids growth: it reinforces what is good and helps drive improvement where needed.
1. Be considerate: feedback is about helping others to improve, not scoring points. Use care and consideration of how it will be received.
1. Provide timely feedback: it is better to address things when they are fresh in our minds and quicker to remedy.
1. Assume positive intent when receiving feedback. When you feel defensive it’s OK to
 respond later.
1. Be optimistic, treat situations as if a positive outcome is always possible.
1. Be open and honest about your own mistakes, take ownership of them and their resolve.
1. Appreciate each other. Say thanks often, preferably in public.

```

knolleary (approved):

PR #778, opened 2023-05-25
Add pagination blog overview page

ZJvandeWeg:
## Description

The more we write, the bigger the pages get. This change introduces pagination. In the current state, not ready for merging. The navigation between pages is built by me, not someone that actually styles things. @Yndira-FlowForge can you take a look?

## Related Issue(s)

Per: https://github.com/flowforge/website/issues/636

## Checklist

 - [X] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
 - [X] I have considered the performance impact of these changes
    - Yes! Should be better

### DIFF
--- a/src/blog.njk
+++ b/src/blog.njk
@@ -106,14 +106,13 @@ meta:
     {%- endfor -%}
     </ul>
 
-    <nav aria-labelledby="pagination">
-      <h2 id="my-pagination">This is my Pagination</h2>
-        <ol>
-          <li>{% if pagination.href.previous %}<a href="{{ pagination.href.previous }}">Previous</a>{% else %}Previous{% endif %}</li>
-        {%- for pageEntry in pagination.pages %}
-          <li><a href="{{ pagination.hrefs[ loop.index0 ] }}"{% if page.url == pagination.hrefs[ loop.index0 ] %} aria-current="page"{% endif %}>Page {{ loop.index }}</a></li>
-        {%- endfor %}
-         <li>{% if pagination.href.next %}<a href="{{ pagination.href.next }}">Next</a>{% else %}Next{% endif %}</li>
-       </ol>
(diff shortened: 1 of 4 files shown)


Pezmc:
Should this be considered a draft? The pagination doesn't look production ready!

Screenshot 2023-05-25 at 15 03 49

ZJvandeWeg:
@Pezmc Yes it should, explained that clearly in the description too

Yndira-FlowForge:
@ZJvandeWeg I was doing a bit of research first because is not the most common practice nowadays. Almost everything I've seen is either infinite scroll or a "load more" button.

I found [this article](https://www.nngroup.com/articles/infinite-scrolling-tips/#:~:text=Infinite%20scrolling%20is%20a%20listing,content%20up%20into%20multiple%20pages.), in case you'd like to check it out...

At the moment, I think you're right and the pagination option is the best one for us.

For desktop, it should be like this
Screenshot 2023-05-25 at 17 33 34

When we have more pages, like this
Screenshot 2023-05-25 at 17 36 09

For mobile I think this works, not making the user touch a single number (this can be tricky on phones), but still letting them know where they are and how many pages they have ahead.
Screenshot 2023-05-25 at 17 33 49

[Here's](https://www.figma.com/file/hXDN5PnY3TmeQSv3n7Hna5?node-id=817:2776#454697918) the design in figma

I can try to change it in your branch, but I'm not 100% sure I'll be successful at it!

ZJvandeWeg:
@Yndira-FlowForge I'd rather iterate towards infinite scroll, this PR I could create in 10 minutes. If styling costs more than about the same time, let's just close it.

If we can more forward with this PR, maybe the next step is a "Load more" button or so, but that's much more backend and frontend work already.

PR #794, opened 2023-05-31
Fix Edge Functions - exclude ELEVENTYEDGE from minification

joepavitt:
## Description

Excludes the compiled `<!-- ELEVENTYEDGE -->` comments form html minification, which fixes Edge Functions hosted on Netlify.

Also fixes the gitignore to use `/generated/` rather than `/generated` which is generated at build time by Eleventy Edge & Netlify.

## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -500,6 +500,9 @@ module.exports = function(eleventyConfig) {
                     conservativeCollapse: true,
                     preserveLineBreaks: true,
                     removeComments: true,
+                    ignoreCustomComments: [
+                        /^\s+ELEVENTYEDGE/
+                      ]
                     removeEmptyAttributes: true,
                     removeRedundantAttributes: true,
                     useShortDoctype: true,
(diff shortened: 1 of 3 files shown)


.eleventy.js
```diff
@@ -500,6 +500,9 @@ module.exports = function(eleventyConfig) {
                     conservativeCollapse: true,
                     preserveLineBreaks: true,
                     removeComments: true,
+                    ignoreCustomComments: [
+                        /^\s+ELEVENTYEDGE/
```
ZJvandeWeg, line 504: I think more simple would be:

```suggestion
                        /\<!-- ELEVENTYEDGE --\>/
```

joepavitt: If that was the full comment, yes, but an example full comment is: 

`<!--ELEVENTYEDGE_edge "njk" %}EleventyEdgeNunjucksPrecompile:349b99eaaab03aab275b22b91bad65b5a21877cc28a5a6db3f487f3367129c80ELEVENTYEDGE_edge-->`

joepavitt: This regex (as things stand doesn't actually work - was meant to open as Draft)

ZJvandeWeg:
As this is the solution that was proposed by Pez in Slack, I'm going ahead and merge this without his explicit review.

Pezmc:
@ZJvandeWeg "This regex (as things stand doesn't actually work - was meant to open as Draft)"

PR #800, opened 2023-05-31
Link to Pillar Page in Resources heading

### LINKED ISSUE #796: Link to Node-RED Pillar Page from top-level navigation
### Description

We should link to the Node-RED Pillar page from the top level navigation. This will help the search crawlers put more importance on this page. We can add a 'About Node-RED' menu item to the resource menu that will link to this page.

### Epic/Story

_No response_

joepavitt:
## Description

- Link to `/node-red` from the "Node-RED" header text in the Resources navigation
- To ensure consistency, link to "About FlowForge" from the FlowForge header too.

Request was to have an "About Node-RED" item int he list, but this would make it taller, this was an alternative given the header/text is already there. Easy to switch to initial request if preferred.

## Related Issue(s)

Closes #796 

## Checklist

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -140,7 +140,7 @@
                         {% navoption "resources", null, 0 %}
                             <!-- Mobile -->
                             <ul class="block sm:hidden">
-                                {% navoption "FlowForge", "/about", 1, "flowforge", true, "col-span-2" %}
+                                {% navoption "FlowForge", null, 1, "flowforge", true, "col-span-2" %}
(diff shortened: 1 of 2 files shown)


iskerrett:
The FlowForge link should go to the features page not the company page.

FWIW, I don't find it intuitive to lick on the section headers. If you think this is the best UI, then let's go with it. However, I think adding an extra About Node-RED is more intuitive. That might just be me though.

ZJvandeWeg:
I want to disagree and commit so we can iterate. 😀

PR #812, opened 2023-06-01
Add "on the roadmap" tooltip for features table

### LINKED ISSUE #548: Add tooltips across the site
Screenshot 2023-03-23 at 09 53 40

### Description

https://github.com/flowforge/website/commit/8b5b4f3ece9119d407356d47434077b8df4a5e72 added a new "tooltip" component into our website via an Eleventy Paired Shortcode. We have multiple places across our website that would benefit from tooltips, a quick sweep of the site should be undertaken and add these where appropriate.

joepavitt:
## Description

- Adds tooltips to the features table on the /pricing page

Couldn't find any other obvious places for the tooltips to be added, so this can close #548 until we find more.

## Related Issue(s)

Closes #548 

## Checklist

### DIFF
--- a/src/_includes/feature_lists/cloud.njk
+++ b/src/_includes/feature_lists/cloud.njk
@@ -163,7 +163,9 @@
                 {% if value === "check" %}
                 {% include "components/feature-check.svg" %}
                 {% elif value === "time" %}
-                {% include "components/feature-time.svg" %}
+                {% tooltip "On the roadmap" %}
+                    {% include "components/feature-time.svg" %}
+                {% endtooltip %}
                 {% elif value === "optional" %}
                 {% include "components/feature-optional.svg" %}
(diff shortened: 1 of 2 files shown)


Yndira-FlowForge:
The icon is not hiding behind the header as I scroll

Screenshot 2023-06-01 at 18 46 23

joepavitt:
> The icon is not hiding behind the header as I scroll

fixed

joepavitt:
> do you think we need the key if we add the tooltip?

Mobile - can't hover on mobile.

> Could the icons have the link to the roadmap? or would that be too much?

Cautious of causing too much navigation away from the page, if it linked to specifically _when_ in the roadmap the respective item was, then maybe, but think it would be too easy to accidentally navigate away from the screen.

Yndira-FlowForge (approved):

PR #833, opened 2023-06-08
Remove edge function declaration

joepavitt:
## Description

Temporarily remove edge functions as our website is down due to Netlify Edge function problems.


joepavitt:
merging directly in order to get website live again

PR #836, opened 2023-06-08
Update page title to improve click rate.

iskerrett:
Change the title of this page from Introduction to Node-RED to be Node-RED Resources.  Also changed the sub title to be A collection of technical resources that answer your questions about Node-RED.

The goal is to improve the ctr for this page.  As of June 6, the last 28 days, the CTR was 3%, the average position was 20.5.

For the query 'node red', ctr was 0.3, average position 26.4
For the query 'node-red', ctr was 1%, average position 24.8

We should compare the results after 28 days.

### DIFF
--- a/src/node-red.njk
+++ b/src/node-red.njk
@@ -2,7 +2,7 @@
 layout: page
 nohero: true
 meta:
-    title: Introduction to Node-RED
+    title: Node-RED Resources
 description:
     Node-RED is a programming tool for wiring together hardware devices, APIs and online services in new and interesting ways.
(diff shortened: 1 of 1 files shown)


joepavitt:
> For the query 'node red', ctr was 0.3, average position 26.4
For the query 'node-red', ctr was 1%, average position 24.8

Both the same query?

iskerrett:
> > For the query 'node red', ctr was 0.3, average position 26.4
> > For the query 'node-red', ctr was 1%, average position 24.8
> 
> Both the same query?

It is reported as different queries in google search console. This is why I captured it here.

joepavitt:
My bad, totally missed the `-` 🤦‍♂️

joepavitt (approved):

PR #850, opened 2023-06-13
Update labels

MarianRaphael:
## Description

Added "customer request label" description in our handbook


knolleary (commented):
Looks good - only outstanding task is the request to add a link to this from the sales page.

Issue #720, opened 2023-05-08
Improve support page: List SLA's and who gets what support


ZJvandeWeg:
### Description

The support page could do better by setting expectations for our customers when raising an issue. There's a lot of information on: https://flowforge.com/handbook/operations/support that should be on the main website instead.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Yndira-FlowForge:
Hi @ZJvandeWeg what do you think about renaming it to "Help Center" and adding all the ways our users can get support? Not just the paying customers.

Here's an idea for the content and the order in which we present the information (but it's not a finished design proposal).
<img width="939" alt="Screenshot 2023-06-20 at 20 52 12" src="https://github.com/flowforge/website/assets/129537638/06fcbb41-2411-441f-ab68-6ca4924b5e6a">

The "click here" link would take them to the [support section](https://flowforge.com/pricing/#:~:text=Ticket%2DBased%20Support) on the features table in the pricing page.

I'm not so sure if that's how the chat works or if we want to encourage people to use it. [Here's](https://www.figma.com/file/hXDN5PnY3TmeQSv3n7Hna5?node-id=2170:13030#476848476) the figma layout in case you want to edit the text.

Yndira-FlowForge:
Updated design


![Image](https://github.com/flowforge/website/assets/129537638/cb7179fe-d076-4181-b981-f1f5107b216d)


Prototype [here](https://www.figma.com/proto/hXDN5PnY3TmeQSv3n7Hna5/FlowForge-Website?page-id=0%3A1&type=design&node-id=2174-13145&viewport=-13331%2C4601%2C0.5&scaling=min-zoom&starting-point-node-id=804%3A2591&show-proto-sidebar=1)

ZJvandeWeg:
Love the wire frame! :shipit:

PR #883, opened 2023-06-21
Iteration on support page

Yndira-FlowForge:
## Description

Updated content and layout

## Related Issue(s)

#720

## Checklist


src/support.njk
```diff
@@ -1,27 +1,114 @@
 ---
-layout: layouts/nohero.njk
-title: Request Support
+align: center
+layout: layouts/page.njk
+title: We’re here to help!
```
ZJvandeWeg, line 4: Is the title element in the HEAD of the HTML the same? In that case I'd rather have Support

src/support.njk
```diff
@@ -1,27 +1,114 @@
...
+              </div>
+            </h3>
+            <p class="mb-0">
+              Get real-time assistance and quick answers. Chat with us for prompt support and solutions to your queries.
+            </p>
+          </div>
+          <div class="container text-center max-w-4xl flex flex-col md:items-start mt-6">
+            <button onclick="window.HubSpotConversations.widget.open()" class="ff-btn ff-btn--primary flex flex-col">
+              <span class="text-base uppercase no-underline md:px-4">
+                open chat box
```
ZJvandeWeg, line 64: Nit: COMMUNITY FORUM is in all caps too

```suggestion
                OPEN CHAT BOX
```

ZJvandeWeg (approved):

src/support.njk
Yndira-FlowForge: Yeah, it's the same... Will change it back to Support

src/support.njk
Yndira-FlowForge: The "uppercase" class takes care of that ;) 
Will change it anyways just to tidy it up

src/support.njk
ZJvandeWeg: Help center is fine with me

PR #885, opened 2023-06-21
Handbook product metrics

MarianRaphael:
## Description

This PR introduces a new handbook page focused on 'Product Metrics'. This page aims to provide an understanding of the key metrics we track to inform our data-driven decision-making process.

### DIFF
--- a/src/handbook/product/metrics.md
+++ b/src/handbook/product/metrics.md
@@ -4,33 +4,30 @@ navTitle: Metrics
 
 # Product Metrics
 
-Our ongoing mission is to enable data-driven decision-making that facilitates strategic product development and enhances user experience. We achieve this by focusing on specific product metrics that help us understand user behavior, feature adoption, and the overall performance of our product. 
+Our ongoing mission is to enable data-driven decision-making, facilitating strategic product development and enhancing user experience. To achieve this, we focus on specific product metrics that illuminate user behavior, feature adoption, and overall product performance. 
 
-This handbook page provides an overview about the metrics we focus on, and how they contribute to our overall [product strategy](./strategy.md).
+This handbook page provides an overview of the metrics we concentrate on, and how they contribute to our overall [product strategy](./strategy.md).
 
 ## Principles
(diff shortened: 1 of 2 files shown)


src/handbook/product/metrics.md
```diff
@@ -0,0 +1,33 @@
...
+# Product Metrics
+
+Our ongoing mission is to enable data-driven decision-making, facilitating strategic product development and enhancing user experience. To achieve this, we focus on specific product metrics that illuminate user behavior, feature adoption, and overall product performance. 
+
+This handbook page provides an overview of the metrics we concentrate on, and how they contribute to our overall [product strategy](./strategy.md).
+
+## Principles
+
+#### Data-driven decision making 
+Aligned with our value of [Results](../company/values.md#📈-results), we are staunch advocates for data-driven decision-making. Instead of relying on assumptions or rushing decisions, we let data guide our actions to drive results.
```
ZJvandeWeg, line 14: My only issue is here is that lack of data should make is bias towards action, based on our values. Now it could be interpreted we'll just wait until more data is available.

```suggestion
Aligned with our value of [Results](../company/values.md#📈-results), we are staunch advocates for data-driven decision-making. Instead of relying on assumptions or rushing decisions, we let data guide our actions to drive results.
```

ZJvandeWeg:
@MarianRaphael One minor note, afterwards it can be merged

src/handbook/product/metrics.md
MarianRaphael:
```suggestion
Aligned with our value of [Results](../company/values.md#📈-results), we are staunch advocates for data-driven decision-making. Instead of relying on assumptions or rushing decisions, we let data guide our actions to drive results. this doesn't mean we become paralyzed in the absence of data. Our Iterative Improvement and Bias for Action values remind us that it's crucial to keep moving forward, even when perfect data isn't available. We balance careful analysis with swift action, using the best information available to us at any given moment.
```

PR #911, opened 2023-06-28
Structure for NR docs

ZJvandeWeg:
## Checklist


Yndira-FlowForge:
As for the layout, I must admit that the left nav feels empty since all the list items are only one word each, but it is consistent with our docs page. So, as a first iteration, it works.

Regarding the right nav, it doesn't show anything because there are no anchors. Will there be any anchors? If not, I think it would be better to hide the "on this page" section.

I understand this is still a work in progress since the PR is still a draft, so I'm not sure if it's too soon to mention this, but:

- I noticed that in the `/node-red/core-nodes/link/` the "details" title is repeated three times.
- `node-red/core-nodes/tls/` is basically empty.

And, are we replacing the link to the https://nodered.org/docs/ in our nav-bar? and/or adding a link to this new docs in our node-red page? I'm thinking in [this section](https://flowforge.com/node-red/#:~:text=Node%2DRED%20itself.-,Learn%20Node%2DRED,-Quick%20Tips).

src/node-red/core-nodes/index.njk
```diff
@@ -0,0 +1,23 @@
...
+provide the basic building blocks for creating Node-RED flows.</p>
+
+<p>Core nodes include nodes for basic functionality like input/output, processing,
+and control flow. They are the foundation upon which more complex workflows can
+be built, and they are essential to the operation of Node-RED.</p>
+
+<p>The nodes are grouped in a few sections in the nodes sidebar:
+<ol type="1">
+    <li>common - miscellaneous nodes</li>
+    <li>funtion - Control flow, blocking or sending additional messages</li>
```
Yndira-FlowForge, line 18:
```suggestion
    <li>function - Control flow, blocking or sending additional messages</li>
```

ZJvandeWeg:
@Yndira-FlowForge Can you have a look at improving the styling of the generated pages? I think it does need some work.

Yndira-FlowForge:
> Can you have a look at improving the styling of the generated pages? I think it does need some work.

Hi @ZJvandeWeg you were right. There were some missing classes in our CSS. I've added them and made some adjustments to the template's margins.

Check it out and let me know what you think.

PR #912, opened 2023-06-29
Comms slack update

ZJvandeWeg:
## Description

Two changes around usage of slack. Firstly, be explicit that the handbook is a place to store decisions etc made in Slack.

Second; normalize the channel names for departments.

Engineering got `#dev`, marketing `#marketing`, bizz ops has none. There's an `art-requests` channel, which is a github stream but unlike all other channels like it doesn't feature the GH prefix. These little things I want to fix.

I propose we rename a few things:
1. art-requests -> gh-art-requests
2. dev -> dept-engineering
3. marketing -> dept-marketing
4. peopleops -> dept-peopleops

### DIFF
--- a/src/handbook/company/communication.md
+++ b/src/handbook/company/communication.md
@@ -75,7 +75,7 @@ below so you get to decide if you want more channels or not.
 
 #### Channel name structure
 
-To make it easier to find your way around in Slack, it's 
+To make it easier to find your way around in Slack, we have some rules for naming channels. This will help give everything some context and implicit sense of purpose.
 
 |Type | Slack prefix | Purpose |
 |:--- |:--- |:---|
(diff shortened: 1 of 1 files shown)


src/handbook/company/communication.md
```diff
@@ -75,12 +75,15 @@ below so you get to decide if you want more channels or not.
 
 #### Channel name structure
 
-When a channel is created for projects; work that's scoped to a certain topic
-that's going to be completed at some time in the future, it might be valuable to
-collect all internal communications into one Slack channel. Please prefix the
-channel name with `proj-` to make grouping easier.
+To make it easier to find your way around in Slack, it's
```
knolleary, line 78: Not sure where you planned to go with this sentence, but here's a proposed second half...

```suggestion
To make it easier to find your way around in Slack, we have some rules for naming channels. This will help give everything some context and implicit sense of purpose.
```

knolleary (approved):

PR #920, opened 2023-07-03
fix: Add date value to HTML element

ZJvandeWeg:
The element was empty, as item.date is undefined. The string interpolation created the empty string instead. With this page the value is filled.

### DIFF
--- a/src/_includes/layouts/post.njk
+++ b/src/_includes/layouts/post.njk
@@ -64,7 +64,7 @@ layout: layouts/base.njk
                             {% renderTeamMember people[author] %}
                         {% endif %}
                     {% endfor %}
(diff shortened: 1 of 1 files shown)


ZJvandeWeg:
Rebased on current main @joepavitt

joepavitt:
Confused why this is a "fix" though, the published on dates show fine on the site as they are: https://flowforge.com/blog/2023/06/dashboard-announcement/

ZJvandeWeg:
@joepavitt the value in the HTML property wasn't set. So humans indeed could read it, though machines found it hard to parse the output

joepavitt (approved):
Thanks for the clarification

PR #942, opened 2023-07-07
Fix "home" link on Left Nav

joepavitt:
## Description

Regression introduced in https://github.com/flowforge/website/pull/911/ meant that the "docs" and "handbook" links at the top of the left navigation were just going to our homepage.

Problem was a missing `else` in the ternary operator.

Can confirm that this link now works across all of handbook, docs and core-nodes documentation.

## Checklist

### DIFF
--- a/src/_includes/layouts/left-nav.njk
+++ b/src/_includes/layouts/left-nav.njk
@@ -1,8 +1,7 @@
 <!-- Navigation -->
 <div class="border-r" data-{{nav}}>
     <ul class="handbook-nav" data-el="navigation">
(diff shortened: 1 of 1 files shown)


ZJvandeWeg:
This change would break the core node Nav though?

joepavitt:
> This change would break the core node Nav though?

It doesn't, test it out: https://deploy-preview-942--flowforge-website.netlify.app/node-red/core-nodes/

knolleary (approved):
Have checked the netlify preview and it all appears to work.

PR #943, opened 2023-07-07
Redirect support for group index & navOrder to control left nav option

joepavitt:
## Description

- `navOrder` can now be set in the frontmatter for an `.md` file, and it will affect the order in which the page appears in the left-side navigation. The higher the number, the more priority it gets at being shown first.
- Added `redirect` as a frontmatter property too. This meant that we could redirect to a given child when a group (.e.g "Device Agent" is clicked, rather than having to have multiple `README.md` files with just a table of content. In this case we can redirect to a `Getting Started` or `Introduction` child.

## Related Issue(s)

Will be opening a FF PR shortly with the first batch of docs changes that utilise these and will backlink once I've got that opened.

## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -414,7 +414,7 @@ module.exports = function(eleventyConfig) {
             let groups = {
                 'Other': {
                     name: 'Other',
-                    order: -1,    // always render last
+                    order: Number.MAX_SAFE_INTEGER,    // always render last
                     children: []
                 }
             }
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -373,7 +373,7 @@ module.exports = function(eleventyConfig) {
                         accumulator[currentValue] = {
                             'name': currentValue,
                             'url': page.data.redirect || page.url,
-                            'order': page.data.navOrder || 0,
+                            'order': page.data.navOrder || Number.MAX_SAFE_INTEGER,
                             'children': {}
                         }
                         if (page.data.navTitle) {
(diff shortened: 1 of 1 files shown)


ZJvandeWeg:
> navOrder can now be set in the frontmatter for an .md file, and it will affect the order in which the page appears in the left-side navigation. The higher the number, the more priority it gets at being shown first.

I found this confusing when I read the code, when I type `1` as order, I'd expect it to go first?

.eleventy.js
```diff
@@ -372,7 +372,8 @@ module.exports = function(eleventyConfig) {
                     if (!accumulator[currentValue]) {
                         accumulator[currentValue] = {
                             'name': currentValue,
-                            'url': page.url,
+                            'url': page.data.redirect || page.url,
+                            'order': page.data.navOrder || 0,
```
ZJvandeWeg, line 376:
```suggestion
                            'order': page.data.navOrder || Number.MAX_SAFE_INTEGER,
```

.eleventy.js
```diff
@@ -439,7 +440,7 @@ module.exports = function(eleventyConfig) {
 
                 function sortChildren (a, b) {
                     // sort children by 'order', then alphabetical
-                    return b.order - a.order || a.name.localeCompare(b.name)
+                    return (b.order - a.order) || a.name.localeCompare(b.name)
```
ZJvandeWeg, line 443:
```suggestion
                    return (a.order - b.order) || a.name.localeCompare(b.name)
```

joepavitt:
> I found this confusing when I read the code, when I type 1 as order, I'd expect it to go first?

That's fair, in which case, in addition to your recommended changes, we'd also need to change the `order: -1` for the "Other" group catch all that's added.

ZJvandeWeg (approved):

PR #955, opened 2023-07-11
Better control over the section ordering in documentation

joepavitt:
## Description

As we've now split the documentation into new sections, we desire control over the ordering of said sections, that wasn't necessary with the Core Nodes or handbook documentation (which order alphabetically)

This change introduces a new map in the `createNav` function which stores an ordered list of the sections in an array. The `order` value for a section is then determined by the index in that array, or falls back to `MAX_INT`, and will show at the bottom (then by alpha if multiple here).

**Now:**

Screenshot 2023-07-11 at 18 28 33

## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -349,6 +349,14 @@ module.exports = function(eleventyConfig) {
         createNav('core-nodes')
 
         function createNav(tag) {
+            const groupOrder = {
+                docs: [
+                    'Overview',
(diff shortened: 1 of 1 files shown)


joepavitt:
Not the most elegant of solutions, but it works, and still provides the flexibility of adding new sections _without_ needing to define an order.

ZJvandeWeg (approved):

PR #956, opened 2023-07-11
July newsletter

iskerrett:
July newsletter


Hisma (approved):
looks good.  Have nothing additional to add.

PR #968, opened 2023-07-13
docs: Core nodes can include a use case text

ZJvandeWeg:
Some docs from upstream make sense to be short in their use case, but for reference documenation, it makes more sense to have more explicit details.

This change adds an example for the split node, as I know the node well to do so, and as-is this change adds an example for more people to be able to chip in.

## Checklist


iskerrett:
Is it possible to add new text interspersed into the auto generated text?

iskerrett:
I have been taking a closer look at this. In your previous article on Split, I like the example code that you gave and the IoT use case. Is it possible to add the example flows and the code here?  

I think an interesting format would be:
- Provide context on how the node can be used. Ideally in the context of IoT use case.
- Create some examples of the node being used in a flow, include a picture of the flow and the code.

ZJvandeWeg:
@iskerrett I've moved the split blog article and redirected the traffic to the new site on the website. Let me know if this is a good example for you.

iskerrett (approved):

PR #989, opened 2023-07-20
core nodes: Use case for inject

ZJvandeWeg:
With two examples

## Related Issue(s)

Part of: https://github.com/flowforge/content/issues/122 

## Checklist


src/_includes/core-nodes/inject-use-case.md
```diff
@@ -0,0 +1,31 @@
+The Inject node is the beginning of many flows that are triggered manually. The box to the left of the node sends a message to connected nodes. For that reason it's often used for debugging too, to inject values at a point of chosing. The message item can be empty too. The message to send defaults to the timestamp as payload and an empty topic.
```
ZJvandeWeg, line 1:
```suggestion
What's the Inject node in Node-RED used for?


The Inject node is the beginning of many flows that are triggered manually. The box to the left of the node sends a message to connected nodes. For that reason it's often used for debugging too, to inject values at a point of chosing. The message item can be empty too. The message to send defaults to the timestamp as payload and an empty topic.
```

PR #994, opened 2023-07-20
core node: example for debug node

iskerrett:
## Description

Add use case and example to Debug node documentation

## Checklist

 - [x ] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
 - [x ] I have considered the performance impact of these changes
 - [x ] Suitable unit/system level tests have been added and they pass
 - [x ] Documentation has been updated


iskerrett:
The images did not render correctly on my localhost but none of the images rendered correctly so I am hoping this is an issue with my machine.

PR #1007, opened 2023-07-28
Add use case and example for Link node

iskerrett:
Add use case description of Link node and an example.


robmarcer (changes requested):
If you install the SQLite custom node (https://flows.nodered.org/node/node-red-node-sqlite), the screenshots won't show the 'unknown alerts'.

iskerrett:
> If you install the SQLite custom node (https://flows.nodered.org/node/node-red-node-sqlite), the screenshots won't show the 'unknown alerts'.

Pushed a new version of the image with sqlite installed.

robmarcer (approved):

PR #1020, opened 2023-08-08
Create community-news-08.md

iskerrett:
August newsletter


Hisma:
looks good.  I was going to add some notes to buffer parser, but I think I can instead just cover it when I do a modbus video/article so it's easier to understand using an example

PR #1067, opened 2023-08-26
core nodes: Document the unknown node

ZJvandeWeg:
## Checklist


src/_includes/core-nodes/unknown-use-case.md
```diff
@@ -0,0 +1,18 @@
...
+
+!["List of nodes installed, including unused nodes"](./images/list-nodes-unused.png)
+
+Presently it is not possible to install older versions of a package into Node-RED so your flows will need to run on the latest version of each package, you may want to consider updating before you migrate.
+
+## Migrating your flows to FlowFuse
+
+When migrating your flows from any Node-RED to a FlowFuse managed Node-RED instance
+it's advices to use [the `nr-tools plugin`](/docs/user/migration) created by the FlowFuse team. Migrating this way will copy not only the flow, but also install
+the required packages, and copy over the credentials.
```
Steve-Mcl, line 18:
```suggestion
When migrating your flows from any Node-RED to a FlowFuse managed Node-RED instance
it is advised to use [the `nr-tools plugin`](/docs/user/migration) created by the FlowFuse team. Migrating this way will not only the copy your flows, it will include any credentials you have set and install
the required packages for you.
```

src/_includes/core-nodes/unknown-use-case.md
```diff
@@ -0,0 +1,18 @@
+The `unknown` node is not one users want to add to their flows, and there's no
+way to add them from the node palette. This node will only show up when importing
+a flow with nodes that have not yet been installed.
+
+To understand what nodes need installing when importing the flow, it's advised
+to create a list of nodes _before_ exporting your nodes. Such a list can be obtained
+by going to the manage palette. When a node is unused, it will be be labelled as
+such.
```
Steve-Mcl, line 8:
```suggestion
The `unknown` node is not one users can add add to their flows, and there's no
way to add them from the node palette. This node will only show up when when you
import a flow with nodes that have not yet been installed.

To understand which nodes need installing when importing a flow, it is advised
to create a list of installed node packages _before_ exporting your flows. Such a list 
can be obtained by looking in the palette manager.
```

Steve-Mcl:
You may want to tip a wink to the `System Info` action - that will also list out installed nodes in text form

chrome_mtrD4ykzwS

lib/image-handler.js
```diff
@@ -145,6 +145,7 @@ module.exports = function imageHandler(
         extraOpts = {
             sharpOptions: {
                 animated: true,
+                limitInputPixels: false
```
ZJvandeWeg, line 148: @Steve-Mcl This was the missing piece to get your Gif with that size to be compressed properly :)

Steve-Mcl (approved):

Steve-Mcl:
Not sure when you want this merged ZJ so I'll leave that to you.

PR #1080, opened 2023-08-30
sales: Document the outbound property

ZJvandeWeg:
## Description

Document when to set the outbound property to 'Yes'

### DIFF
--- a/src/handbook/sales/index.md
+++ b/src/handbook/sales/index.md
@@ -56,4 +56,10 @@ For each sales opportunity a clear answer for the customers needs and wants shou
 If, and only if, the answers are known and added as notes to HubSpot, an
 opportunity is added to the HubSpot deals board.
 
+## HubSpot Properties
+
+`Outbound` is a custom property that's set to `Yes` when the first meeting with the contact came through outbound
+drip campaigns or other outbound lead-gen actions. This property will be set to `Yes` when the contact was in HubSpot
+through other marketing activities too, but wasn't nurtured to the point of a meeting yet.
+
 See also the [stages each contact in HubSpot must go through](/handbook/marketing/hubspot.md#lifecycle-stage).
--- a/src/handbook/sales/index.md
+++ b/src/handbook/sales/index.md
@@ -58,7 +58,7 @@ opportunity is added to the HubSpot deals board.
 
 ## HubSpot Properties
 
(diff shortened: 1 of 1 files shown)


zackwasli (approved):

src/handbook/sales/index.md
```diff
@@ -56,4 +56,10 @@ For each sales opportunity a clear answer for the customers needs and wants shou
 If, and only if, the answers are known and added as notes to HubSpot, an
 opportunity is added to the HubSpot deals board.
 
+## HubSpot Properties
+
+`Outbound` is a custom property that's set to `Yes` when the first meeting with the contact came through outbound
+drip campaigns or other outbound lead-gen actions. This property will be set to `Yes` when the contact was in HubSpot
+through other marketing activities too, but wasn't nurtured to the point of a meeting yet.
```
iskerrett, line 63: What does this mean? If the contact was in Hubspot already isn't that the definition of Inbound?

ZJvandeWeg: I see what you mean. The property sources on creation are still set by HS. The intent is to understand which meetings are SDR generated. I bet we can rename the property. Any suggestions @iskerrett ?

iskerrett: Typically Inbound leads do not have a contact owner assigned. If an SDR starts working a contact, they should assign themselves as the owner. I don't think we need a new property for this?

ZJvandeWeg: @zackwasli Can we rename the field to "First Meeting Outbound".

ZJvandeWeg: Actually, let's go with "Activation Outbound"

src/handbook/sales/index.md
```diff
@@ -56,4 +56,10 @@ For each sales opportunity a clear answer for the customers needs and wants shou
 If, and only if, the answers are known and added as notes to HubSpot, an
 opportunity is added to the HubSpot deals board.
 
+## HubSpot Properties
+
+`Outbound` is a custom property that's set to `Yes` when the first meeting with the contact came through outbound
```
ZJvandeWeg, line 61:
```suggestion
`Activation Outbound` is a custom property that's set to `Yes` when the first meeting with the contact came through outbound
```

PR #1097, opened 2023-09-06
core nodes: Inject a static string

ZJvandeWeg:
## Description

Add another example where the inject node doesn't inject a timestamp to show it's more versatile.

## Checklist


src/_includes/core-nodes/inject-use-case.md
```diff
@@ -1,5 +1,3 @@
-## What's the Inject node in Node-RED used for?
```
ZJvandeWeg, line 1: If the page title is "Inject" already, this feels unnecessary

robmarcer (approved):

PR #1098, opened 2023-09-06
Make your builds faster with this simple trick!

ZJvandeWeg:
Before this change each save of a file would take 6.5 seconds on my laptop before the site was regenerated. In development scenario's that's slow.

By leveraging Eleventy's incremental builds feature this is slashed to 3.0s. While still slow, it's much better.

## Checklist


Pezmc (approved):
This seems sensible to me, I'm not clear on why it's not the default!

I'd like to give @joepavitt a moment to give his two cents, so not merging.

joepavitt (approved):

PR #1115, opened 2023-09-11
liquid: Render `{{` correctly

ZJvandeWeg:
As the markdown in our website is preprocessed by Liquid, the `{{` need escaping with the `{% raw %}`. That part was left out by me in 1dfd90aeeee174b871846b566d54893ba79e35c6

With this change it renders correctly.

## Checklist


Steve-Mcl:
Almost.  The JSON being presented to the user is also missing the mustache 

`"template": "<p>Hello !</p>",`
image


meaning if a visitor copies the flow snippet, it wont work.

I believe the JSON Flow needs to be also escaped.

Steve-Mcl (approved):
Success

PR #1119, opened 2023-09-14
handbook: Update expense section

ZJvandeWeg:
Incorporating changes from: https://github.com/flowforge/website/pull/879

Further, including a section about personal expenses.

## Checklist

### DIFF
--- a/src/handbook/peopleops/expenses.md
+++ b/src/handbook/peopleops/expenses.md
@@ -102,7 +102,7 @@ A screenshot of the route and the total distance travelled should be included wi
 
 ### Personal expenses
 
-At FlowFuse we value a culture that encourages team members to take ownership of their personal expenses, without the expectation of reimbursement. This approach is all about acknowledging personal preferences and recognizing the individual benefits that come with these expenses. Here are some examples, but are not limited to
+We expect team members to take ownership of their personal expenses without the expectation of reimbursement. This approach is all about acknowledging personal preferences and recognising the individual benefits that come with these expenses. Here are some examples, but are not limited to:
 
 1. Coffee at your home office
 1. Passport for travel
--- a/src/handbook/peopleops/expenses.md
+++ b/src/handbook/peopleops/expenses.md
@@ -71,8 +71,15 @@ or CEO, who will provision it for you.
 
 ### Workspace & Location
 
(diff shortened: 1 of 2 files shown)


Tracy-Anthony (approved):

src/handbook/peopleops/expenses.md
```diff
@@ -96,3 +99,11 @@ For example:
 A screenshot of the route and the total distance travelled should be included with any expense claim.
 
 [gh-admin]: https://github.com/flowforge/admin
+
+### Personal expenses
+
+At FlowFuse we value a culture that encourages team members to take ownership of their personal expenses, without the expectation of reimbursement. This approach is all about acknowledging personal preferences and recognizing the individual benefits that come with these expenses. Here are some examples, but are not limited to
```
ZJvandeWeg, line 105:
```suggestion
We expect team members to take ownership of their personal expenses without the expectation of reimbursement. This approach is all about acknowledging personal preferences and recognising the individual benefits that come with these expenses. Here are some examples, but are not limited to:
```

src/handbook/peopleops/expenses.md
```diff
@@ -96,3 +99,11 @@ For example:
...
 
 [gh-admin]: https://github.com/flowforge/admin
+
+### Personal expenses
+
+We expect team members to take ownership of their personal expenses without the expectation of reimbursement. This approach is all about acknowledging personal preferences and recognising the individual benefits that come with these expenses. Here are some examples, but are not limited to:
+
+1. Coffee at your home office
+1. Passport for travel
+1. Daycare for your children
```
joepavitt, line 109: @ZJvandeWeg This last example doesn't feel fair given that this change is very clearly linked to the fact that I requested childcare cover to look after my kids for the Monday of the BCN Summit.

This was required because the summit means I have to work on a day when I'm not contracted to and would not be paid for, meaning that we have nobody to look after the girls. We had two options:

- Incur a £120 charge to put them in childcare for the day
- I travel on the Tuesday and miss the first half day of the summit, which was made clear to me that wasn't an option.

If FF are enforcing that I work on a day I'm not contracted to, I would expect expenses consequently incurred to be covered. This does not fall under the same bracket as a passport, or coffee.

PR #1157, opened 2023-09-23
SEO: canonical in link tag

ZJvandeWeg:
## Description

We used the right data, just in the wrong tag. And it was never picked up by Google or other search engines.

## Checklist


MarianRaphael (approved):
Waiting for tests

PR #1183, opened 2023-09-29
rebrand: Update the GH org name

ZJvandeWeg:
org on GH was renamed too, this change updates this repo


knolleary (changes requested):
You've replaced `flowforge` with `FlowForge` ... right case, wrong spelling ;)

knolleary (approved):

PR #1188, opened 2023-09-29
Rename repo references

knolleary:
https://github.com/FlowFuse/admin/issues/213

## Description

Renames references to repositories, excluding `flowforge` and `flowforge-dev-env` which will follow in separate PR


ZJvandeWeg (approved):
Went through all files, LGTM

PR #969, opened 2023-07-13
Only clean docs in development

Pezmc:
## Description

Replacement PR for https://github.com/flowforge/website/pull/966, which only cleans out the docs directly in development mode. 

`npm run start` (dev mode) will now:
 - Clean `site` (but not `img`)
 - Clear out `docs/**md`
 - Builds the javascript via terser 
 - Run `dev:docs` which copies across `../flowforge/docs`, if it exists
 - Run `dev:postcss`
 - Builds the website via netlify

In production, it'll remain as always, which Joe detailed in https://github.com/flowforge/website/pull/965#issuecomment-1633870796

## Related Issue(s)

#966 and #965

~Fixes 967~

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/flowforge/flowforge/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [x] Suitable unit/system level tests have been added and they pass
 - [-] Documentation has been updated

PR #1194, opened 2023-10-02
Handle repo rename in build scripts and add GH action caching

knolleary:
## Description

The build scripts copy the FF docs from `../flowforge/docs`. This PR updates the scripts to work with `../flowfuse/docs` and fall back to `../flowforge/docs` otherwise. This eases the transition and doesn't break anyone who hasn't rebuild their local environment.

It also enables caching of the `_site/img` directory in the GH action. This *should* speed up the time it takes to run the `test` action (after the initial run completes to populate the cache).


knolleary:
Once the initial run of the [`Test` action](https://github.com/FlowFuse/website/actions/runs/6379117161) completes for this PR (in about 30 minutes), I think we should be able to rerun it to observe the caching in action...

package.json
```diff
@@ -12,8 +12,8 @@
         "build:js": "terser -c -m -o _site/js/cc.min.js node_modules/vanilla-cookieconsent/dist/cookieconsent.js src/js/cookies.js",
         "build": "cross-env NODE_ENV=production npm-run-all clean build:js --parallel prod:*",
         "clean:dev": "cross-env npx del-cli '_site/!(img)' 'src/docs/**/*.md' && cross-env npx mkdirp '_site/js'",
```
Pezmc, line 14: This is now redundant, it was originally added for the exclusion of the img dir, they can both just use `clean` now

Pezmc (commented):
Assuming the build re-runs faster and the redundant package script is removed, 👍

Pezmc:
Second run took only 1m 46s (down from 30 mins); I've just enabled npm caching too, which brought the build down to 1m 30s, a notable improvement.

package.json
Pezmc: Not quite true, we need to clean the docs directly locally: https://github.com/FlowFuse/website/pull/969

PR #1212, opened 2023-10-05
Ensure dir created before copying files

hardillb:
## Description

Fixes race condition in copy_docs.js


knolleary (approved):
Good spot 👍🏻

PR #1219, opened 2023-10-06
Slack theme

ZJvandeWeg:
## Description

While screensharing yesterday I saw an ugly theme. That's why I record a
better way(tm) now.

## Checklist

### DIFF
--- a/src/handbook/company/communication.md
+++ b/src/handbook/company/communication.md
@@ -66,6 +66,9 @@ sharing. As such knowledge and decision made in Slack should always be recorded
 in the handbook, GitHub issue, or Google Doc on that topic to keep a log that also works
 for asynchronous communication.
 
+Our recommended Slack theme is:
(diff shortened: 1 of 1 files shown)


knolleary (approved):
This matches the theme I've been using since the early days, so can't remember the origin of it. Appreciate it may not by to the taste of everyone (particular dark theme users), but it's some harmless fun.

All that said, as I understand it, the new Slack UI will end up adapting any custom themes to use a palette of 'approved' colours... we'll have to wait and see what that brings when our workspace is blessed with the update.

src/handbook/company/communication.md
```diff
@@ -66,6 +66,9 @@ sharing. As such knowledge and decision made in Slack should always be recorded
 in the handbook, GitHub issue, or Google Doc on that topic to keep a log that also works
 for asynchronous communication.
 
+Our recommended Slack theme is:
```
knolleary, line 69:
```suggestion
Our recommended Slack (light) theme is:
```

PR #1313, opened 2023-11-10
Add blueprint library page

### LINKED ISSUE #1256: FlowFuse Blueprint Library - Page
The idea is to have a page on the website to present  our "blueprints," allowing users to search for specific solutions before signing up and to improve our SEO

## Examples

Node-RED Library: https://flows.nodered.org/
Tulip Library: https://tulip.co/library/

Yndira-FlowForge:
## Description

First iteration of the blueprint library page

## Related Issue(s)

closes #1256

## Checklist


Yndira-FlowForge:
@MarianRaphael, this is the first iteration for the page.

It still needs some work, and the idea is to have it synced with https://github.com/FlowFuse/blueprint-library/tree/main.

For now, I've simply copied the files from that repository, but those markdown files don't match the proposed structure in the issue. I've also copied the flow.json files, although I'm not sure if you'll need them.

How many blueprints and categories do we expect to have? This will determine the structure of the library.

Yndira-FlowForge:
BTW add /blueprints/ to the preview URL to view the page. Where should we place it in the menu?

ZJvandeWeg:
>>>
It still needs some work, and the idea is to have it synced with https://github.com/FlowFuse/blueprint-library/tree/main.

For now, I've simply copied the files from that repository, but those markdown files don't match the proposed structure in the issue. I've also copied the flow.json files, although I'm not sure if you'll need them.

How many blueprints and categories do we expect to have? This will determine the structure of the library.
>>>

@Yndira-FlowForge Can you look at how we copy the docs and achieve this PR like this? That's easier to manage for everyone.

Yndira-FlowForge:
Done! I've opened a [PR](https://github.com/FlowFuse/blueprint-library/pull/20) in the blueprints repository too. Once merged, it should enable the desired rendering.

MarianRaphael:
First iteration looks good! I need to improve the Readme representation for the website, but I like the concept and its sync with the blueprint repository.
Screenshot 2023-11-18 at 16 28 15

.gitignore
```diff
@@ -4,6 +4,7 @@ node_modules
 src/handbook/media
 src/docs/*
 !src/docs/docs.json
+src/blueprints/*
```
ZJvandeWeg, line 7: This only works now because you've already put blueprints.njk and templates.njk are in Git already. Can we add:

```
!src/blueprints/*.njk
```

Yndira-FlowForge:
Tests are failing, and I believe it's because the blueprints repository is private.

One solution involves creating a personal token (which I already did) and adding it as a secret in the website repository; but I don't have access to its settings. Additionally, if I'm understanding correctly, this would tie my gh user to the blueprints repository access, so that content would only be updated if I open the PR.

The other solution, could be [this one](https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#:~:text=Instead%20of%20using%20a%20personal%20access%20token%2C%20consider%20using%20a%20GitHub%20App). I consider this is a better option, because it would update regardless of who opens the PR, but I would need help with this one, and I think it has to be an admin who creates/adds the app.

.github/workflows/build.yml
```diff
@@ -21,6 +21,12 @@ jobs:
           repository: 'FlowFuse/flowfuse'
           ref: maintenance
           path: 'flowfuse'
+      - name: Check out FlowFuse/blueprint-library repository (to access the blueprints)
+        uses: actions/checkout@v3
+        with:
+          repository: 'FlowFuse/blueprint-library'
+          ref: main
+          path: 'blueprints'
```
ZJvandeWeg, line 29:
```suggestion
          path: 'blueprints'
          token: ${{ secrets.ACCESS_TOKEN }}
```

Yndira-FlowForge:
It's passing the tests now! 🎊 The issue was with the path in the image key in the front matter; now, just the filename is enough, as long as it is in the same folder as the markdown file.

Since we don't have many blueprints per category, I removed the filter. We can add it as part of a future iteration when we have more blueprints.

Yndira-FlowForge:
It's passing the tests now! 🎊 The issue was with the path in the image key in the front matter; now, just the filename is enough, as long as it is in the same folder as the markdown file.

Since we don't have many blueprints per category, I removed the filter. We can add it as part of a future iteration when we have more blueprints.

ZJvandeWeg (approved):

MarianRaphael:
PR looks good. I just have to rework all the blueprint READMEs to have better representation on our website. Additionally, not all blueprints in the repo are yet available in Flowfuse due to UI limitations that we will fix next week.

MarianRaphael (approved):
@Yndira-FlowForge good to go from my side. In the main blueprint branch, there are now just the three available blueprints.

PR #1360, opened 2023-12-01
comms: Information storing systems

ZJvandeWeg:
## Checklist

### DIFF
--- a/src/handbook/company/communication.md
+++ b/src/handbook/company/communication.md
@@ -56,6 +56,20 @@ Slack to be matched automatically.
 
 ## Asynchronous communication
 
+Communication through different timezones or due to differences in availility
+requires knowledge to be persistent in sources that are approachable to others
+and non-emphermeral.
+
+Information should thus be stored in:
+1. Handbook
+1. GitHub issues
+1. Git Repositories pushed to GitHub
+1. Google Drive (Docs / Sheets / etc.)
+
+Don't store information in:
+1. Slack
+2. Figma
+
 ### Email
 
 Email for internal communication is only used for discussions that need
@@ -68,12 +82,7 @@ When using Slack, prefer to keep discussions in a thread. This reduces scroll
 back, and focusses channels.
 
 To keep everyone on the same page and to allow others to collaborate you should favour
-public channels over DMs. 
-
-Note that knowledge in Slack is ephemeral, it's not storage for information or knowledge
-sharing. As such knowledge and decision made in Slack should always be recorded
-in the handbook, GitHub issue, or Google Doc on that topic to keep a log that also works
-for asynchronous communication.
+public channels over DMs.
 
 Our recommended Slack (light) theme is:
 `#F2F2F2,#FFC806,#779FA1,#FFFFFF,#F0D36C,#111111,#2AB052,#FD0B00,#AA4444,#FFFFFF`
--- a/src/handbook/company/communication.md
+++ b/src/handbook/company/communication.md
@@ -56,9 +56,9 @@ Slack to be matched automatically.
 
 ## Asynchronous communication
 
-Communication through different timezones or due to differences in availility
+Communication through different timezones or due to differences in availability
 requires knowledge to be persistent in sources that are approachable to others
-and non-emphermeral.
+and non-ephemeral.
 
 Information should thus be stored in:
 1. Handbook


src/handbook/company/communication.md
```diff
@@ -56,6 +56,20 @@ Slack to be matched automatically.
 
 ## Asynchronous communication
 
+Communication through different timezones or due to differences in availility
+requires knowledge to be persistent in sources that are approachable to others
+and non-emphermeral.
```
knolleary, line 61:
```suggestion
Communication through different timezones or due to differences in availability
requires knowledge to be persistent in sources that are approachable to others
and non-ephemeral.
```

src/handbook/company/communication.md
```diff
@@ -56,6 +56,20 @@ Slack to be matched automatically.
...
+
+Information should thus be stored in:
+1. Handbook
+1. GitHub issues
+1. Git Repositories pushed to GitHub
+1. Google Drive (Docs / Sheets / etc.)
+
+Don't store information in:
+1. Slack
+2. Figma
```
knolleary, line 71: I think there is a difference between slack and figma here. We store designs in figma as it provides interactivity you don't get with static screenshots to better communicate a design as part of an issue. None of the 'approved' list provide that functionality.

Capturing screenshots as point-in-time records to include in a GH issue is a useful practice to encourage ease of reference, but I don't think we should say we don't store things in figma.

ZJvandeWeg: @knolleary I consider Figma like a paper notebook, it's great for ideation, but not an indexed and approachable store of information (for community members or some FF team members). Thus for any asset that needs delivery it should be stored in another system than Figma.

I think the difference between what you mean with what's OK in Figma is _intermediate_ deliveries. That's not knowledge or information, but on par with an unpublished branch for developers.

joepavitt: Figma should be used for storing Designs - that's literally what it's designed for. We can share elsewhere, but to say that Figma can't be used to store Designs seems a very odd angle to take.

ZJvandeWeg: > Thus for any asset that needs delivery it should be stored in another system than Figma.

That's not at all what it reads @joepavitt, it's for ideas. Delivery is always elsewhere, the actual product that users can use.

joepavitt: All of our designs for Device Groups were prototyped in Figma for @Steve-Mcl to use. He and I have been going back and forth actively in comments in Figma, iterating the design, because it's the best place for it. Should we not be doing that?

knolleary: There is a distinction between short-term and long-term information storing as well as the different uses we have for the tool.

I think the key point of the proposed change is about where we store information in the long-term for future reference.

Figma is the right tool for doing the design work in, and being able to refer back to it as we iterate is important. That doesn't preclude including snapshots of 'agreed' designs in issues as a point-in-time reference.

However there are other uses of figma that wouldn't be appropriate to treat as the long-term store for reference. For example, where we use figjams to ideate on a topic - the output will typically result in something being updated in the docs/handbook elsewhere. The figjam isn't kept as the single source of truth.

There are parallels with Google Docs; some uses are entirely appropriate (as per our Info Security policies) and some will be more short-lived 'working' documents as part of the overall process

ZJvandeWeg: @joepavitt AFAIK device groups were designed and the design is not being pursued, which makes it a great example why design and Figma are part of the ideation phase and hold little end-user benefit.

> Also not sure how Figma is different to Google Drive in this case?

I'll interpret it as a sarcastic comment for your benefit.

joepavitt: > AFAIK device groups were designed and the design is not being pursued, which makes it a great example why design and Figma are part of the ideation phase and hold little end-user benefit.

They were designed, and Steve is building it now. https://github.com/FlowFuse/flowfuse/issues/2997

ZJvandeWeg:
After another incident today we're I was told something already excited but it wasn't known to anyone except the "design team", I'm going to use my authority to merge this.

Going forward no information should be stored in Figma @joepavitt and @Yndira-FlowForge

joepavitt:
> After another incident today we're I was told something already excited but it wasn't known to anyone except the "design team", I'm going to use my authority to merge this.

Out of interest, what is this referring to?

> Going forward no information should be stored in Figma

Then what _can_ we use Figma for?

joepavitt:
can I also please get clarity on how you would like us to share prototypes that demonstrate UX flows without using Figma?

ZJvandeWeg:
@joepavitt Screenshots or plain PRs. Nothing in figma is to be considered information. Please read the diff

joepavitt:
> Screenshots or plain PRs. Nothing in figma is to be considered information. Please read the diff

I appreciate that is what the PR said, but you're now blocking us from using Figma to create prototypes for UX flow which makes front-end dev a lot harder. There is no alternative tool that we use for this?

Similarly, for retrieving padding, font sizes, styles, colors, etc. all of that is contained within the Figma file and lost in a screenshot - how do you propose we work around not being able to use Figma for these?

Also still interested in understanding _"it wasn't known to anyone except the "design team"_ was referring to please.

Steve-Mcl:
Not wanting to inflame the matter or step on toes but...

The work on device groups was helped hugely by Figma. I would go as far as to say invaluable & I hope the next large task is as well laid out. I was able to browse and discuss the design interactively and get responses and clarifications in a fast and highly useful manor - in a way that PRs and screenshots can not convey (far to noisy & cluttered etc).

It is my hope we can keep Figma for this kind of iteration (even if that means final screenshots are to be included in the PR)?

For reference - you can see our interactions and how it helped with implementing (and where necessary tweaking) the design here:  https://www.figma.com/file/rMB2w5ZZUtTC0nxbwLYKJY/FlowForge-Platform?type=design&node-id=3636%3A12502&mode=design&t=ti6Xpx7QyoRLfh3O-1

Yndira-FlowForge:
@joepavitt 

> Out of interest, what is this referring to?

It was actually a FigJam file, [this one](https://www.figma.com/file/dR9jvt0NVXSxl7RfHZJcM2/UX-Inspirations-%26-Competitors?type=whiteboard&node-id=0%3A1&t=EeKEJhnxgJtiYKKh-1). It's the UX/UI references that you put together, there's no design in there.

I think we could keep working prototypes as we've being doing it so far, like Steve said, as long as the final design is in the issue or PR as an image. We actually refer to this in the [handbook](https://flowfuse.com/handbook/design/process/#design-review) too, but maybe we could make it clearer and include this "condition".

joepavitt:
> It was actually a FigJam file, this one. It's the UX/UI references that you put together, there's no design in there.

Thanks Yndira. I put that together originally as my own reference point and notes. My point still stands though, if this instead lives in a Google Doc, it's no different.

joepavitt:
> I think we could keep working prototypes as we've being doing it so far

I would hope so too, but from what ZJ has put here, we are not permitted to do so, as no information (I.e. how the pages/buttons are linked together to create the UX flow) is allowed to be stored in Figma.

I agree entirely that everything should be referenced/linked in GH, but Figma is far superior for storing this information, so being able to link to a Figma UX Prototype in GH should still be permitted imo.

PR #1363, opened 2023-12-01
policy: CEO underperformance

ZJvandeWeg:
The performance of a CEO is reviewed by the full company, including employees. This change makes it clear, I hope, that there's a process for that too and the CEO doesn't have benefits in their role others don't enjoy with regards to underperformance.

## Checklist

### DIFF
--- a/src/handbook/peopleops/organization.md
+++ b/src/handbook/peopleops/organization.md
@@ -132,3 +132,6 @@ Directors.
 
 All employees are free to provide feedback or input to the board of directors
 through email. It's adviced not to include the CEO in such an email.
+
+If the board agrees with the CEO underperforming a seperate process will be followed
+tailored to the sitation and the bylaws of FlowFuse.
--- a/src/handbook/peopleops/organization.md
+++ b/src/handbook/peopleops/organization.md
@@ -123,3 +123,12 @@ a PIP. This ensures that a PIP is a genuine change.
 If an employee's performance does not improve, or if they violate the terms of
 their PIP, they may be terminated. Termination will be based on a documented
 pattern of underperformance or unacceptable behavior.
+
+## CEO underperformance
+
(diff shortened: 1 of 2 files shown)


src/handbook/peopleops/organization.md
```diff
@@ -123,3 +123,15 @@ a PIP. This ensures that a PIP is a genuine change.
...
 their PIP, they may be terminated. Termination will be based on a documented
 pattern of underperformance or unacceptable behavior.
+
+## CEO underperformance
+
+As with all employees in the company the CEO can underperform. Generally speaking
+the CEO does get performance reviews, though these are provided by the Board of
+Directors.
+
+All employees are free to provide feedback or input to the board of directors
```
joepavitt, line 133: Going to the board as an employee about CEO performance feels very last port of call and serious though?

Even for what might be general performance concerns where there could be areas of improvement for example?

joepavitt: For example, a single employees grievances about CEO performance may just be that single employee, but going to the board feels very serious, particularly in a case like this?

As someone that's never operated on a board, I don't really know _how_ these kind of things are received by a board.

ZJvandeWeg: It should be as serious as every other time discussing performance to someone's manager and not them.

I'll update this PR

src/handbook/peopleops/organization.md
```diff
@@ -123,3 +123,15 @@ a PIP. This ensures that a PIP is a genuine change.
...
+through email. It's adviced not to include the CEO in such an email.
```
ZJvandeWeg, line 134:
```suggestion
through email. This venue is to be used in all cases where you'd usually also go to someone's manager
to discuss a reports underperformance. It's advised not to include the CEO in such an email.
```

Issue #1346, opened 2023-11-23
Iterate on /node-red content


Yndira-FlowForge:
### Description

- [x] [Update hero section description](https://github.com/FlowFuse/website/pull/1353)
- [x] [Update "What is Node-RED" section and add 3 most recent customer stories](https://github.com/FlowFuse/website/pull/1355)
- [x] [Add 3 cards for: Cloud, Self Hosted and Fleet Management](https://github.com/FlowFuse/website/pull/1390)
- [x] [Update FAQs with JSON ld](https://github.com/FlowFuse/website/pull/1365)

### Epic/Story

#1314

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Yndira-FlowForge:
@ZJvandeWeg About the 3 cards section, would this content make sense?

## Flexible Deployment Options with FlowFuse

### Cloud
The easiest way to start building Node-RED applications—no installation required. Simply sign up and gain immediate access to the editor.

### Self Hosted
If you prefer running it on your own, that's possible too! Explore our documentation for available options and follow step-by-step instructions for installation.

### Fleet Management
Manage thousands of Node-RED instances, whether in the cloud or on edge devices. Explore our FlowFuse Device Agent to connect and utilize Node-RED on edge devices.

--

**I have doubts about where to link the first two:**
Cloud: another link to sign up? Pricing page?
Self hosted: docs? this would only take them to the free version, should we link to the pricing page instead?

ZJvandeWeg:
@Yndira-FlowForge 

Content wise I'd go with:

## Flexible Deployment Options with FlowFuse

Run and manage Node-RED where it's most helpful for your business.

### FlowFuse Cloud
The easiest way to start building Node-RED applications—no installation required. Simply sign up and gain immediate access to the editor.

### Self managed (on-premise)
Run FlowFuse where you prefer; that's possible too! In your cloud of preference, or even on your own hardware. Explore our documentation for available options and follow step-by-step instructions for installation.

### Edge device

Run Node-RED where the data is generated. Manage thousands of edge devices through the FlowFuse Device Agent.


-----

> Cloud: another link to sign up? Pricing page?

Pricing page

> Self hosted: docs? this would only take them to the free version, should we link to the pricing page instead?

"Talk to Sales" linked to the book a demo page

PR #1365, opened 2023-12-01
Add FAQs a structured data

Yndira-FlowForge:
## Description

Added the FAQs to the front matter, so the're recognized by Google as rich content

## Related Issue(s)

#1346 

## Checklist


src/node-red/index.njk
```diff
@@ -3,9 +3,56 @@ layout: page
 nohero: true
 meta:
     title: Start using Node-RED online for free
-description:
-    Node-RED's low-code brilliance hosted on FlowFuse cloud. Answer your questions about getting started with Node-RED including tips, best practices, and FAQs to using Node-RED for 
-    different use cases.
+    faq:
```
ZJvandeWeg, line 6: Can eleventyComputed data return data structures like an array of hashes?

ZJvandeWeg: As a first iteration I believe this is good though. Just wondering if you considered it

ZJvandeWeg (approved):

src/node-red/index.njk
Yndira-FlowForge: I'll look into it, I did investigate if I could call the data from the json I previously had and couldn't find a way. I'm gonna merge as is, but if you can point an article or some source of info to improve it, it would be very helpful 😊

PR #1371, opened 2023-12-05
No Credit Card required for billing

MarianRaphael:
## Description

- Add changelog entry about the new billing option.

## Checklist

### DIFF
--- a/src/changelog/2023/12/billing.md
+++ b/src/changelog/2023/12/billing.md
@@ -1,6 +1,6 @@
 ---
 title: No Credit Card required for billing
-date: 2023-12-01 13:00:00.0
+date: 2023-12-06 13:00:00.0
 authors: ["marian-demme"]
 tags:
     - changelog


src/changelog/2023/12/billing.md
```diff
@@ -0,0 +1,10 @@
+---
+title: No Credit Card required for billing
+date: 2023-12-01 13:00:00.0
+authors: ["marian-demme"]
+tags:
+    - changelog
+---
+Previously, it was necessary to provide credit card details when transitioning from a trial to a subscription. Recognizing the diverse needs of businesses, we have now removed this requirement. You can now seamlessly switch to a subscription without the need to enter credit card details. We're committed to accommodating your specific billing preferences and can set up a process that aligns with your organizational needs.
```
knolleary, line 8: I don't think it is right to say we've removed the need need to enter a CC - for most users that is still the case. It is also not something a user can do for themselves - it requires us to action for them.

```suggestion
We now have more flexible billing options in FlowFuse Cloud for customers that wish to purchase an annual subscription or where entering credit card details is not appropriate. We're committed to accommodating your specific billing preferences and can set up a process that aligns with your organizational needs.
```

knolleary (changes requested):

knolleary (approved):

PR #1376, opened 2023-12-05
A/B contact us form from /pricing

Yndira-FlowForge:
## Description
`control` would be the pricing page as is, linking to this improved Contact Us page:
image

And `test` would be the pricing page with the form at the end of the page, like this:
image

## Checklist

### DIFF
--- a/src/_includes/feature_lists/self-hosted.njk
+++ b/src/_includes/feature_lists/self-hosted.njk
@@ -134,8 +134,8 @@
             onclick: "capture('cta-contact', {'position': 'secondary'}, {'plan': 'sh-team'})"
         }, {
             cta: "CONTACT SALES",
-            url: "/contact-us/",
-            onclick: "capture('cta-contact', {'position': 'secondary'}, {'plan': 'sh-enterprise'})"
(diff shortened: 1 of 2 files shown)


Yndira-FlowForge:
@zackwasli I checked, and for this week, it only shows your hours. For next week it includes ZJs time slots too, so I think is good to go. Can you please review the PR?

zackwasli (approved):

Yndira-FlowForge:
Hi @joepavitt, I could use some assistance with the A/B testing.

- I've set up the experiment in PostHog, [here](https://eu.posthog.com/experiments/3343) in the "test" project. It's a straightforward test to check if it's recording anything, and seems like it's capturing both feature flags at the same time.
- I'm having trouble capturing form submissions. I attempted to add it using the PostHog toolbar, but it doesn't seem to recognize the button...
- I've also added the feature flags. It's not entirely ready, but I wanted to test it online.
- On both local and online environments, nothing is showing inside the edge tags. It's hiding everything, and upon inspection, I noticed all the edge blocks are commented out.
- I recall you mentioned that Edge functions would be deprecated, but as far as I understand, this hasn't happened yet.

Am I overlooking something?

I'll be out of the office this week and back on Monday. If you could provide some insights before I return, that would be greatly appreciated.

Thanks in advance!

Yndira-FlowForge:
@joepavitt I'm adding the form to the pricing page without the A/B test. So, no need to check this one out anymore 😉

PR #1359, opened 2023-11-30
Add flexible deployment section to /node-red

Yndira-FlowForge:
## Description

Added the flexible deployment section

## Related Issue(s)

#1346

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowforge/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated

PR #1390, opened 2023-12-12
Add deployment section to nr page

Yndira-FlowForge:
## Description

Added the flexible deployment section

## Related Issue(s)

This is a duplicate of https://github.com/FlowFuse/website/pull/1359 which I merged by mistake before it was approved

## Checklist


src/node-red/index.njk
```diff
@@ -182,6 +182,60 @@ sitemapPriority: 0.8
...
+                        <div class="title">
+                            <h4>
+                                <span class="text-gray-600">FlowFuse Cloud</span></br>
+                            </h4>
+                        </div>
+                        <p class="mb-0 text-left">
+                            The easiest way to start building Node-RED applications. No installation required! Simply sign up and gain immediate access to the editor.
+                        </p>
+                    </div>
+                    <a class="md:self-end ff-btn ff-btn--primary-outlined uppercase align-baseline w-full mt-3" href="/pricing/">MORE INFO</a>
```
Yndira-FlowForge, line 205: @ZJvandeWeg you asked in the PR that I accidentally merged ([here](https://github.com/FlowFuse/website/pull/1359/files#r1424066398)):

> Is this the correct CTA?

And I replied:

> It's linking to the pricing page, not the sign-up page. That's why I thought of "More info" instead of sign-up.
> 
> My thinking here is that we already have the sign up button in the hero section and the main nav, and this section includes the word "business", maybe someone could be interested in a different plan? Team or enterprise...
> 
> Do you think it should link to the sign-up page too?

It was the only thing pending from your review.

ZJvandeWeg: I still don't _know_ what's the right way, "Sign up" feels better. But this is fine, and we can iterate later.

ZJvandeWeg (approved):

Issue #1385, opened 2023-12-11
Iterate on /book-demo


Yndira-FlowForge:
### Description

- [x] [Add ZJ hours to the calendar](https://github.com/FlowFuse/website/issues/1369)
- [x] [Iterate on the content](https://github.com/FlowFuse/website/pull/1392)
    - Considering replicating the /contact-us structure

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #1392, opened 2023-12-12
/book-demo iteration

Yndira-FlowForge:
## Description

Added social proof and links to other contact channels

## Related Issue(s)

#1385

## Checklist


src/book-demo.njk
```diff
@@ -8,12 +8,64 @@ meta:
     description: Book a demo to understand how FlowFuse helps you professionalize your Node-RED deployment.
 ---
 <div class="text-left max-w-full ff-bg-light">
-  <div class="flex justify-center items-center md:block">
-    <div class="ff-bg-light w-full product px-4 pt-4 md:pt-12">
+  <div class="flex flex-col justify-center items-center md:block">
+    <div class="ff-bg-light w-full product px-4 pt-4 md:pt-12 -mb-24 sm:-mb-6">
```
ZJvandeWeg, line 12: Does tailwind have a way to collect classed together into one with a semantical naming scheme?

src/book-demo.njk
```diff
@@ -8,12 +8,64 @@ meta:
     description: Book a demo to understand how FlowFuse helps you professionalize your Node-RED deployment.
 ---
```
ZJvandeWeg, line 9: Now the pages are so alike, shouldn't it be a one layout and two pages that render them?

ZJvandeWeg: Or alternatively, more partials?

src/book-demo.njk
Yndira-FlowForge: I created new classes, it's cleaner now

src/book-demo.njk
Yndira-FlowForge: Done, one layout and two pages

Yndira-FlowForge:
@ZJvandeWeg 

> I think we can DRY the code out a bit

You were right, it's better now 😉

PR #1438, opened 2023-12-22
Pricing page update - Device Group Management

MarianRaphael:
## Description

Pricing page update - Device Group Management

## Checklist


ZJvandeWeg:
While I'll merge this now, I don't think this is a great feature name. It doesn't let the customer understand what the value is of the feature and reduces the value too much. Consider rebranding the feature set @MarianRaphael

Issue #1433, opened 2023-12-21
Add capability to schedule blog posts


Hasmin-AC:
### Description

See: [https://remysharp.com/2019/06/26/scheduled-and-draft-11ty-posts ](https://remysharp.com/2019/06/26/scheduled-and-draft-11ty-posts) 
Needs the blog post implemented
Needs a trigger at noon daily to trigger a deploy

This will enable us to publish blogs on a schedule. 

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #1439, opened 2023-12-23
Add capability to schedule blog posts

Yndira-FlowForge:
## Description

Created a new posts collection. 
Added filter to exclude posts with future dates (only in production, so the preview is available in dev_mode).
Added a warning message to scheduled posts.
Changed the cron function to build the site everyday at noon.

## Related Issue(s)

closes [#1433](https://github.com/FlowFuse/website/issues/1433)

## Checklist


Yndira-FlowForge:
I added the src/blog/2023/01/test.md to test how a post with a scheduled date behaves, I didn't delete it so you can see how it looks in the preview.

Please, do the review, but do not merge the PR, I'll delete that file once it's approved.

src/_includes/layouts/post.njk
```diff
@@ -1,8 +1,14 @@
 ---
 layout: layouts/base.njk
 type: post
+eleventyComputed:
```
ZJvandeWeg, line 4: Not quite sure why this direction was chosen over the linked blog post, with a new collection? I feel this approach is a bit too much magic.

src/_includes/layouts/post.njk
```diff
@@ -1,8 +1,14 @@
...
+    permalink: "{% if date | isFuturePost %}false{% else %}{{permalink}}{% endif %}"
 ---
-
+{% if date | shortDate > now | shortDate %}
```
ZJvandeWeg, line 7: What's the last `shortDate` for? It's a bit confusing for me. This whole section implies the post is rendered and served by the server, which shouldn't be the case IMO.

ZJvandeWeg (changes requested):

.eleventy.js
```diff
@@ -30,6 +30,29 @@ const DEV_MODE = process.env.ELEVENTY_RUN_MODE !== "build" // i.e. serve/watch
 module.exports = function(eleventyConfig) {
     eleventyConfig.setUseGitIgnore(false); // Otherwise docs are ignored
 
+    // Set DEV_MODE_POSTS to true if the deploy URL starts with 'https://deploy-preview-'
+    const DEV_MODE_POSTS = process.env.ELEVENTY_RUN_MODE !== "build" ||
```
ZJvandeWeg, line 34: You should use the env var from Netlify: https://docs.netlify.com/configure-builds/environment-variables/#build-metadata

src/_includes/layouts/post.njk
Yndira-FlowForge: I created the new collection, this is just setting the permalink to false so it doesn't generate the pages of future posts in production, but keep them visible in development

.eleventy.js
Yndira-FlowForge: Thanks for this @ZJvandeWeg! It's visible on the Netlify preview now. I tested it with `CONTEXT=production npm run build`, and it doesn't generate the page.

Should we keep the test file with the future date to check if it actually skips the file in production? I'll remove it afterward.

ZJvandeWeg (changes requested):
I believe you should update the handbook: https://flowfuse.com/handbook/marketing/blog/ and make sure people understand they can write _and merge!_ content in advance.

src/_includes/layouts/post.njk
```diff
@@ -1,8 +1,14 @@
...
+    <div class="w-full page post bg-teal-400 text-center text-gray-800 py-4 font-bold">
+        This post is scheduled to be published on {{ date | shortDate }}
```
ZJvandeWeg, line 9: This doesn't show up for me locally?

src/_includes/layouts/post.njk
```diff
@@ -1,8 +1,14 @@
 ---
 layout: layouts/base.njk
 type: post
+eleventyComputed:
+    permalink: "{% if date | isFuturePost %}false{% else %}{{permalink}}{% endif %}"
```
ZJvandeWeg, line 5: By using a collection I don't think you need this anymore?

Yndira-FlowForge: It's necessary, if I don't set the permalink to false it will certainly be excluded from the collection, but it will still generate the post page in production.

.eleventy.js
```diff
@@ -39,6 +39,12 @@ module.exports = function(eleventyConfig) {
         return isFuturePost && !DEV_MODE_POSTS;
     });
 
+    // Define a filter named 'isFutureDate'
+    eleventyConfig.addFilter('isFutureDate', (date) => {
+        const isFutureDate = date && date > new Date();
+        return isFutureDate;
```
ZJvandeWeg, line 45:
```suggestion
        return date && date > new Date();
```

src/_includes/layouts/post.njk
Yndira-FlowForge: Is it showing up now?

ZJvandeWeg (approved):

PR #1441, opened 2023-12-27
pidd-case-study

robmarcer:
## Description

Initial article and images

We are waiting for some photos from PIDD which I think we should add into this article before we put it on the web site


src/images/stories/pidd-view.png
ZJvandeWeg, line 1: Love this image, it's very good

robmarcer: I agree, 90% of the credit for it goes to @Yndira-FlowForge. I'm keen to do more like this in the future, I think it really adds value to the case study.

PR #1517, opened 2024-01-15
Update bug bounty payment details

knolleary:
Updates how we will make payments under the bug bounty.


ZJvandeWeg (changes requested):

knolleary:
Made a minor tweak to your suggestions, but the result is the same.

ZJvandeWeg (approved):

PR #1523, opened 2024-01-15
Add CSP to prevent embedding in an iframe

knolleary:
This sets a Content-Security-Policy header that prevents the site from being embedded in an iframe - one measure to prevent click-jacking on the site.

There are further policies to add to CSP, but starting with this.

This is a work-in-progress - need to review the preview site generated by netlify before pushing for full review.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,9 @@
+[[headers]]
+  for = "/*"
+  [headers.values]
+    X-Frame-Options = "DENY"
+    Content-Security-Policy = "frame-ancestors 'none';"
+
 [[redirects]]
 from = "http://flowforge.com/*"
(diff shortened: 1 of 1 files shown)


knolleary:
It doesn't look like custom headers get applied to the preview site. So I think we need to merge this and see what we get on the live site. Will push this when more of us are online to revert if needed.

ZJvandeWeg (approved):

ZJvandeWeg:
@knolleary Feel free to merge when you're ready and please validate when the new site is deployed.

knolleary:
Found the issue - a limitation of the netlify custom headers - https://docs.netlify.com/routing/headers/#limitations

We are using an edge function to serve the site - so we'll need to add custom headers in there instead.

PR #1528, opened 2024-01-16
Add CSP headers to 11ty edge function response

knolleary:
## Description

This is a second attempt to get custom headers set on the website.

The first attempt #1523 didn't work because we have an edge-function serving the content. This is a documented limitation of setting headers via the `netlify.toml` file.

The workaround is to set the headers explicitly in the edge function - which this PR does.

### DIFF
--- a/netlify/edge-functions/eleventy-edge.js
+++ b/netlify/edge-functions/eleventy-edge.js
@@ -178,7 +178,10 @@ export default async (request, context) => {
         });
         const distinctId = getDistinctId(context)
         setCookie(context, "ff-distinctid", distinctId, 1);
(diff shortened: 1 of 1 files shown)


knolleary:
Have verified via the deploy preview that the custom headers are set and that it blocks the site from being embedded in an iframe.

joepavitt (approved):

joepavitt:
Approved and merged, although worth noting we're likely steering away from Edge Functions, as our only use to date has been for A/B Testing, and that's now proving very unreliable. Eleventy have also announced removal of Edge Functions in future realese.

PR #1541, opened 2024-01-18
Filter recommended posts by comom tags

Yndira-FlowForge:
## Description

Added a related filter to recommend articles based on common tags.

Based on https://github.com/11ty/eleventy/discussions/2534#discussioncomment-3415878

## Related Issue(s)

https://github.com/FlowFuse/website/pull/1522#issuecomment-1896299923

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated

PR #1544, opened 2024-01-18
Iteration on recommended posts section

Yndira-FlowForge:
## Description

Only posts with the same set of tags were considered related. With this change, those with all tags except one in common are also considered related.

I also added an 'else' condition so that if there are no related posts, it will show the latest.

## Related Issue(s)

https://github.com/FlowFuse/website/pull/1541

## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -292,9 +292,12 @@ module.exports = function(eleventyConfig) {
 
     eleventyConfig.addFilter("relatedPosts", function (collection = []) {
         const { tags: requiredTags, page } = this.ctx;
-        return collection.filter(post => {
-            return post.url !== page.url && requiredTags?.every(tag => post.data.tags?.includes(tag));
-        });
+        return collection
+            .map(post => {
+                const commonTags = requiredTags?.reduce((count, tag) => count + (post.data.tags?.includes(tag) ? 1 : 0), 0);
+                return { ...post, commonTags };
+            })
+            .filter(post => post.url !== page.url && post.commonTags >= requiredTags.length - 1);
     });
     
     // Custom async filters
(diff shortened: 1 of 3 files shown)


src/_includes/layouts/post.njk
```diff
@@ -77,6 +77,19 @@
                                 </li>
                             {%- endfor %}
                         </ul>
+                    {%- else %}
+                        <h3 class="mb-3 pt-6 border-t-2">Recommended Articles:</h3>
```
ZJvandeWeg, line 81: I believe a partial would be better, or a paired shortcode? This feels a bit duplicate.

.eleventy.js
```diff
@@ -292,9 +292,12 @@ module.exports = function(eleventyConfig) {
...
         const { tags: requiredTags, page } = this.ctx;
-        return collection.filter(post => {
-            return post.url !== page.url && requiredTags?.every(tag => post.data.tags?.includes(tag));
-        });
+        return collection
+            .map(post => {
+                const commonTags = requiredTags?.reduce((count, tag) => count + (post.data.tags?.includes(tag) ? 1 : 0), 0);
+                return { ...post, commonTags };
+            })
+            .filter(post => post.url !== page.url && post.commonTags >= requiredTags.length - 1);
```
Yndira-FlowForge, line 300: @ZJvandeWeg I'm having second thought about this change, because I saw you worked on the tags for the UNS articles, and right now it looks like this:
Screenshot 2024-01-18 at 16 34 40

But with this change, there are too many newer related articles, and those ones ⬆️ are excluded from the list:
Screenshot 2024-01-18 at 16 34 57

So, I'm thinking of keeping it as we have it now, and only adding the "else" in the post.njk so it's never an empty section. WDYT?

PR #1549, opened 2024-01-18
relatedPosts: Sort first by most tags in common

Yndira-FlowForge:
## Description

Updated the filter to display the list of related posts by most common tags in descending order

## Related Issue(s)

https://github.com/FlowFuse/website/pull/1544

## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -297,7 +297,9 @@ module.exports = function(eleventyConfig) {
                 const commonTags = requiredTags?.reduce((count, tag) => count + (post.data.tags?.includes(tag) ? 1 : 0), 0);
                 return { ...post, commonTags };
             })
-            .filter(post => post.url !== page.url && post.commonTags >= requiredTags.length - 1);
+            .filter(post => post.url !== page.url && post.commonTags >= requiredTags.length - 1)
+            .sort((a, b) => b.commonTags - a.commonTags || b.date - a.date)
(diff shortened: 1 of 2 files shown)


Yndira-FlowForge:
I clicked on every post in the first /blog page, and couldn't find the one I told you that was suggesting the welcoming posts, maybe I saw it in a different version while working on the previous iteration...

.eleventy.js
```diff
@@ -297,7 +297,9 @@ module.exports = function(eleventyConfig) {
                 const commonTags = requiredTags?.reduce((count, tag) => count + (post.data.tags?.includes(tag) ? 1 : 0), 0);
                 return { ...post, commonTags };
             })
-            .filter(post => post.url !== page.url && post.commonTags >= requiredTags.length - 1);
+            .filter(post => post.url !== page.url && post.commonTags >= requiredTags.length - 1)
+            .sort((a, b) => b.commonTags - a.commonTags || b.date - a.date)
```
ZJvandeWeg, line 301: This must be JS magic, but when is `b.commonTags - a.commonTags` falsy?

Yndira-FlowForge: This seems like magic to me as well. It's subtracting the amount of tags in common of two posts at a time to determine the order, if two posts have the same amount of tags, the result is 0 (falsy), and then it compares by date and puts the most recent first.

PR #1577, opened 2024-01-24
tracking: Don't track in DEV_MODE

ZJvandeWeg:
## Checklist


.eleventy.js
```diff
@@ -82,6 +82,7 @@ module.exports = function(eleventyConfig) {
 
     // make global accessible in src/_includes/layouts/base.njk for loading of PH scripts
     eleventyConfig.addGlobalData('POSTHOG_APIKEY', () => process.env.POSTHOG_APIKEY || '' )
+    eleventyConfig.addGlobalData('DEV_MODE', () => !DEV_MODE && !DEV_MODE_POSTS)
```
Yndira-FlowForge, line 85: wouldn't it make more sense
```suggestion
    eleventyConfig.addGlobalData('DEV_MODE', () => DEV_MODE && DEV_MODE_POSTS)
```

src/_includes/layouts/base.njk
```diff
@@ -290,14 +290,15 @@ document.getElementById('nav-toggle').onclick = function(){
 <script async src="https://cdn.jsdelivr.net/npm/lite-youtube-embed@0.2.0/src/lite-yt-embed.min.js"></script>
 <link rel="preload" href="https://cdn.jsdelivr.net/npm/lite-youtube-embed@0.2.0/src/lite-yt-embed.min.css" as="style" onload="this.onload=null;this.rel='stylesheet'" crossorigin="anonymous" referrerpolicy="no-referrer" />
 
+<!-- Cookie banner -->
+<script async type="text/javascript" src="/js/cc.min.js"></script>
+
+{%- if DEV_MODE -%}
```
Yndira-FlowForge, line 296: and then 
```suggestion
{%- if not DEV_MODE -%}
```

.eleventy.js
Yndira-FlowForge: Or, instead of calling it `dev_mode`, make it `prod_mode`? And update de if in `base.njk`

src/_includes/layouts/base.njk
Yndira-FlowForge: It just feels weird to read that the tracking will only happen when in dev_mode, but dev_mode means not dev mode

.eleventy.js
```diff
@@ -82,6 +82,7 @@ module.exports = function(eleventyConfig) {
 
     // make global accessible in src/_includes/layouts/base.njk for loading of PH scripts
     eleventyConfig.addGlobalData('POSTHOG_APIKEY', () => process.env.POSTHOG_APIKEY || '' )
+    eleventyConfig.addGlobalData('DEV_MODE', () => DEV_MODE && DEV_MODE_POSTS)
```
Yndira-FlowForge, line 85: It's not working in the netlify preview, only on local, I think it should be
```suggestion
    eleventyConfig.addGlobalData('DEV_MODE', () => DEV_MODE || DEV_MODE_POSTS)
```

Yndira-FlowForge (approved):

Issue #1613, opened 2024-02-01
Upgrade netlify to node 20.x


Pezmc:
### Description

On 30th of January Netlify moved to node LTS v20. Several of the websites current versions of Netlify plugins and dependencies need to be upgraded for support. Currently the website builds with node 16.20.2, which appears to be set in Netflify config.

The build was broken but has been temporarily patched with an upgrade of _some_ of the dependencies. Issue first seen in: https://github.com/FlowFuse/website/issues/1613

The following errors show in the netlify logs.

```9:07:21 AM: ❯ Warning: Starting January 30, 2024 plugin "@algolia/netlify-plugin-crawler" will be executed with Node.js version 20.
9:07:21 AM:   We"re upgrading our system node version on that day, which means the plugin cannot be executed with your defined Node.js version 16.20.2.
```
​
```
9:07:21 AM: ❯ Warning: Starting January 30, 2024 plugin "netlify-plugin-cache" will be executed with Node.js version 20.
9:07:21 AM:   We"re upgrading our system node version on that day, which means the plugin cannot be executed with your defined Node.js version 16.20.2.
```
​
```
9:07:21 AM:   Please make sure your plugin supports being run on Node.js 20.
9:07:21 AM: ​
9:07:21 AM:   Read more about our minimum required version in our forums announcement
9:07:21 AM: ​
9:07:21 AM: ❯ Loading plugins
9:07:21 AM:    - @netlify/plugin-lighthouse@4.0.7 from Netlify app
9:07:21 AM:    - @algolia/netlify-plugin-crawler@1.0.15 from netlify.toml and package.json
9:07:21 AM:    - netlify-plugin-cache@1.0.3 from netlify.toml and package.json
9:07:21 AM: ​
9:07:21 AM: ❯ Outdated plugins
9:07:21 AM:    - @netlify/plugin-lighthouse@4.0.7: latest version is 6.0.0
9:07:21 AM:      To upgrade this plugin, please uninstall and re-install it from the Netlify plugins directory
```

I do not appear to have access to FlowForge/flowfuse.com on Netlify so cannot change the config myself (for plugin lighthouse). I have upgraded the dependencies that are defined in our package.json.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Yndira-FlowForge:
Hi @joepavitt could you please check this out when you have a chance?

joepavitt:
Will deal with this tomorrow morning

[...]

PR #1614, opened 2024-02-01
Explicitly upgrade node to version 20

Pezmc:
For #1613, upgrades need to also be made in Netlify's config.


ZJvandeWeg:
Interesting it cannot be set on the `netlify.toml`, but docs are clear about it. This seems to be the fix

PR #1625, opened 2024-02-05
Make meta keywords dynamic

Yndira-FlowForge:
## Description

I added conditions for dynamic keywords, establishing a group of keywords that will always be present. Additionally, if the page has tags or meta.keywords, those will be included as well.

## Checklist

### DIFF
--- a/src/pricing.njk
+++ b/src/pricing.njk
@@ -4,6 +4,7 @@ layout: layouts/page.njk
 sitemapPriority: 0.90
 meta:
     description: Find out how much it costs to use FlowFuse.
+    keywords: FlowFuse pricing, Cost, Plans, Features, Cloud, Self-hosted
 hubspot:
     script: "hubspot/hs-form.njk"
     formId: d43ab79b-d075-4b63-ba5e-bf971ca0b86e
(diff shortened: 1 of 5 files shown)


src/pricing.njk
```diff
@@ -4,6 +4,7 @@ layout: layouts/page.njk
 sitemapPriority: 0.90
 meta:
     description: Find out how much it costs to use FlowFuse.
+    keywords: FlowFuse pricing, Cost, Plans, Features, Cloud, Self-hosted
```
ZJvandeWeg, line 7: This is great, and I think would be great to be added to a "Page CMS" section of: https://flowfuse.com/handbook/customer/marketing/blog/#blog-cms

ZJvandeWeg (approved):

src/pricing.njk
Yndira-FlowForge: It's just that for the blog, priority is currently given to tags over meta.keywords. Do you think it would make more sense if meta.keywords were given priority?

Hasmin-AC: Meta keywords are important for SEO and for our pages are found on search. So I would make sure it gets as much priority

ZJvandeWeg: If I create a page, I should add these I think, going by the pricing page example. As such I think we should create instructions in the handbook how to and why.

The priority is implemented correctly.

ZJvandeWeg: @Hasmin-AC The discussion is about what to do on blog posts: Either take the tags as keywords, the current approach, or allow and almost require authors to set the keywords themselves.

Either way, keywords are added. And if you want to, we can swap the priority around -- that is; favour manually entered keywords when provided for blog posts.

For pages there's no contention as there's only the manual keywords.

ZJvandeWeg (approved):

PR #1633, opened 2024-02-06
quick add to a couple of pages

ZJvandeWeg:
## Checklist


src/product/index.njk
```diff
@@ -1,7 +1,9 @@
 ---
-title: Team-Based Low-Code Development
 layout: layouts/page.njk
+title: Team-Based Low-Code Development
```
ZJvandeWeg, line 3: @Hasmin-AC @MarianRaphael Can we do better here?

Hasmin-AC: What is this page exactly? I don't see it linked from any where

ZJvandeWeg: Let's remove it?

Hasmin-AC: yes I think we should.

PR #1757, opened 2024-03-06
Upload teal pictograms

Yndira-FlowForge:
## Description

While changing the website colours I removed the teal pictograms because they're no longer needed in the web, but some are in the platform. This PR brings them back.

## Checklist


knolleary:
I don't think we strictly need all of them, but can review exactly which ones are used by the app separately.

knolleary (approved):
The 'create account' view of the app uses 'time.png' and 'no_credit_card.png' from this collection.

Yndira-FlowForge:
Ok, I thought it would take too long, if you can provide a list, I'll upload only the ones we need.

knolleary:
Lets merge all to get things fixed and then we can be more selective afterwards.

Issue #1791, opened 2024-03-14
Add Link to eBook in Node-RED page


Hasmin-AC:
### Description

This can be added as a link in the "Learn Node-RED" section or as its own call-out box with a graphic just under it. 

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Yndira-E:
@Hasmin-AC I'll add it to the key concepts column, I don't think that document qualifies as a quick tip 😄

Yndira-E:
Do you want it in any particular order? First? Last?

Hasmin-AC:
Yes agree its not a quick tip and neither is it a concept, so can we make it span across the two columns as a learning resource by itself?

Yndira-E:
It would be like this

![Image](https://github.com/FlowFuse/website/assets/129537638/d57c282e-0961-429c-af7b-8e3fd90a3057)

But I think the space to the right is too much, how about like this? And seems more obvious that it's not under quick tips or key concepts (at least on desktop)


![Image](https://github.com/FlowFuse/website/assets/129537638/4d1cfd7a-4002-4824-ac3f-120ef53cc3d7)

PR #1794, opened 2024-03-15
Update ebook title source

Yndira-E:
## Description

When the title is set, the meta.title isn't recognized as such; it's overwritten by the title property. I've changed the property to contentTitle to ensure that meta.title is recognized as both the Open Graph (og) and page title.

## Checklist

### DIFF
--- a/src/_includes/layouts/ebook.njk
+++ b/src/_includes/layouts/ebook.njk
@@ -8,7 +8,7 @@ hubspot:
         <label class="text-red-600">eBook</label>
         <div class="md:flex md:flex-row gap-6">
             <div class="text-left md:w-1/2">
-                <h1>{{ title | safe }}</h1>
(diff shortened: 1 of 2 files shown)


ZJvandeWeg (approved):

ZJvandeWeg:
@Yndira-E Should a resources menu item be added to navigate to the page?

Yndira-E:
We could add it as we did with the pillar pages, but if we add enough ebooks, I think we should have an index page listing them all, like a catalog. @Hasmin-AC did you have a plan for this?

Hasmin-AC:
i've created [an issue ](https://github.com/FlowFuse/website/issues/1791)to add this to the Node-RED page for now. Once we have a few more ebooks and whitepapers we can list them all in a /resources page.

PR #1795, opened 2024-03-15
Add example to Bearer Token changelog

hardillb:
## Description

Add example of setting a Bearer token with Curl

## Checklist


ZJvandeWeg (approved):
Just to check, you've removed this token? As in, it's not providing access to anything right now @hardillb ?

PR #1837, opened 2024-03-25
doc: Add instructions for utilizing the pre-staging environment

ppawlowski:
## Description

This Pull Request extends development documentation by adding information about the possibility of validating changes before emerging to the `main` branch.

## Checklist

### DIFF
--- a/src/handbook/development/contributing.md
+++ b/src/handbook/development/contributing.md
@@ -61,4 +61,9 @@ When merging a PR, you should choose the "Merge pull request" option. There is n
 
 When conducting a PR review, if you are the last (or only) reviewer and all reviews (including your own) are approvals, unless there is a comment from the author stating otherwise, you are free to conduct the merge. Otherwise, leave the merge to the author of the PR, or a future reviewer.
 
+For a comprehensive review of the Pull Request, utilize the designated FlowFuse pre-staging environment. As of the composition of this document, the pre-staging verification is only available for the primary [FlowFuse NPM package](https://github.com/FlowFuse/flowfuse).
+To validate changes within the pre-staging environment, apply the deploy:pr label to the PR. The PR will then undergo automatic deployment to the pre-staging environment, and the deployment link will be appended to the associated workflow summary. The label should remain until the PR is merged.
+A unique pre-staging environment is established for each Pull Request. This environment is terminated upon PR merging or closure. The FlowFuse application deployed from the Pull Request comes pre-configured. Access credentials for the pre-staging environment are located in the FlowFuse 1Password vault.
+
+
 [website-repo]: https://github.com/FlowFuse/website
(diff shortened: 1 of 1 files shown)


joepavitt:
Why can't we do this for _every_ PR - why does it need the manual step of adding a label?

knolleary:
@joepavitt because so far we choose not to.

I believe that for many PRs the test env would go untouched because the PR has been tested locally and the nature of the change doesn't require a dedicated env for further manual testing. We would be throwing money at AWS for no value in return.

If we decided it was worthwhile spinning up on every PR we could do so. I don't think that's necessary today.

hardillb:
@joepavitt because not every change needs a K8s environment to test it

ppawlowski:
@joepavitt  AFAIR this was the main expectation - to have a possibility to verify changes which may break the staging environment.  Not all of them will tend to do this and can be merged and deployed to staging without validation on the pre-staging environment. 
@ZJvandeWeg as it is mentioned in the docs, `This environment is terminated upon PR merging or closure.`

ppawlowski:
To be more clear - from the infrastructure perspective, an environment is in fact FlowFuse Cloud instance running on an existing EKS cluster we use for the staging environment. No new AWS resources are created.

joepavitt:
Can we just sanity check one thing, fairly sure it's covered by the `synchronise` event, but if the base `main` branch is updated, would we auto-deploy the relevant changes to any open PRs?

Thinking of a situation where we have _potentially_ clashing PRs. When one is merged (having been verified in staging) that would update `main`, an associated PR should have it's staging environment also updated to reflect the latest `main`.

joepavitt:
Let's get 1. and 2. ticked of please @ppawlowski

src/handbook/development/contributing.md
```diff
@@ -61,4 +61,9 @@ When merging a PR, you should choose the "Merge pull request" option. There is n
 
 When conducting a PR review, if you are the last (or only) reviewer and all reviews (including your own) are approvals, unless there is a comment from the author stating otherwise, you are free to conduct the merge. Otherwise, leave the merge to the author of the PR, or a future reviewer.
 
+For a comprehensive review of the Pull Request, utilize the designated FlowFuse pre-staging environment. As of the composition of this document, the pre-staging verification is only available for the primary [FlowFuse NPM package](https://github.com/FlowFuse/flowfuse).
+To validate changes within the pre-staging environment, apply the deploy:pr label to the PR. The PR will then undergo automatic deployment to the pre-staging environment, and the deployment link will be appended to the associated workflow summary. The label should remain until the PR is merged.
```
joepavitt, line 65:
```suggestion
To validate changes within the pre-staging environment, apply the `deploy:pr` label to the PR. The PR will then undergo automatic deployment to the pre-staging environment, and the deployment link will be appended to the associated workflow summary. The label should remain until the PR is merged.
```

joepavitt: Can we also get a bit more detail on where to find the link (maybe a screenshot) - given these instructions, and [this example PR](https://github.com/FlowFuse/flowfuse/pull/3652#discussion_r1542676749) - I cannot find the link

joepavitt (changes requested):

joepavitt (approved):

PR #1907, opened 2024-04-08
handbook: Add our legal agreement

ZJvandeWeg:
It's now linked from a Google Doc, but when there's no changes it's better to just capture this doc in the handbook. Also adds a form to request the docx through an email.

## Checklist

### DIFF
--- a/src/handbook/customer/sales/legal.md
+++ b/src/handbook/customer/sales/legal.md
@@ -7,7 +7,7 @@ navGroup: Customer Department
 
 ## Subscription agreement
 
-For self-managed licenses FlowFuse has a standard [subscription agreement template](https://docs.google.com/document/d/1szqCRC7w-f82I4HL-cOTTCiSMKDhVWju84Md4M07RGk){rel="nofollow"}.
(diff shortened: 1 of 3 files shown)


ZJvandeWeg:
@Yndira-E I think the form is just broken right now on the page, and I don't quite understand why. Can you poke at this for a couple of minutes?

Yndira-E:
On it @ZJvandeWeg

Yndira-E:
It works on .njk files, if you change the extension of this md to njk it'll work. You could just copy/paste the script into the md or we could add a configuration to eleventy.js to make all md process nunjucks syntax, but I'm not sure that's what we want here.

ZJvandeWeg:
Done, thanks @Yndira-E. Appreciate your time and debugging

Yndira-E:
What I meant was that the form would appear if you change the extension, but you lose the md formatting if you do. The quickest fix would be to add the script directly into the md, [like in our docs](https://github.com/FlowFuse/flowfuse/blob/main/docs/install/introduction.md?plain=1).

PR #1944, opened 2024-04-18
fix: Internal links don't use a prefix

ZJvandeWeg:
## Description

To have internal links be checked by the hyperlink checker, we cannot use the `https://flowfuse.com` prefix as the tool doesn't know what domain we're serving from.

This change allows us to check for that better.
 
## Checklist


ZJvandeWeg:
cc @sumitshinde-84

PR #1977, opened 2024-04-27
sales: Renewal dates for deals

ZJvandeWeg:
## Description

To make sure our automation for renewals work we should ensure that for all deals next years renewal date is set correctly.

## Checklist


ZJvandeWeg:
@zackwasli Can you please take a look?

zackwasli (approved):
All opps should be updated with renewal date as of now, will continue marking this moving forward

PR #1985, opened 2024-04-30
Node-RED: Allow flows to be rendered as diagrams

ZJvandeWeg:
Utilizes the upstream package remove the need for
images + plain JSON.

For this change I've only shown the split node as an example page. The rest of the nodes will be added
later, though there's some upstream changes that
would be better to add before we roll these changes out wholesale.

## Related Issue(s)

Would be ideal to have: https://github.com/FlowFuse/flow-renderer/issues/28 soonish to fix the height requiring manual checks and updates.

## Checklist


src/_includes/core-nodes/split-use-case.md
```diff
@@ -97,8 +89,6 @@ the catch-all category. Node-RED uses hashes, key-value pairs of property names
 and values. JSON is example of this. In the example below, we'll split the mapping
 of the words "one" and "two" versus the number values.
 
-![Split an object](./images/split-node-hash.png "Split arbitrary objects")
-
-```json
+{% renderFlow 500 %}
```
ZJvandeWeg, line 92: As the flow was exported from a flow that had many other nodes etc present, the coordinates start at Y=340. That means that the view needs to be at least 400 px tall to show the flow. It makes the top part awkward.

cc @joepavitt @Steve-Mcl

ZJvandeWeg: Already done in the description :)

Yndira-E (commented):
Is there a way to set the height of the flow viewer? It's not terrible, but some seem too big for a single line flow.

Screenshot 2024-04-30 at 15 19 52

src/_includes/core-nodes/split-use-case.md
```diff
@@ -69,26 +65,22 @@ Keep the `Split using` set to `\n`, which is the line ending character in Unix
...
 #### Splitting by word
 
 When you've split something by line, you might need to split by word. Simply
 putting a space in the `Split using` setting. That doesn't show a visible character
 in the form, as opposed to the line ending character.
 
-![Node-RED split by word](./images/split-node-by-word.png "Split text by space in Node-RED")
-
-```json
+{% renderFlow 400 %}
```
joepavitt, line 81: @Yndira-E here's an example where height is set

Yndira-E: Thanks Joe!!
I tried to make them all the same height, but sometimes the flows aren't centered or even visible... I'm guessing that's the reason some have an unusual height.
Screenshot 2024-04-30 at 16 47 04

Maybe some information regarding how to set the height should be added to handbook as well?

joepavitt: Yeah, that's caused by https://github.com/FlowFuse/flow-renderer/issues/28

joepavitt:
I'd say it's worth a quick fix of https://github.com/FlowFuse/flow-renderer/issues/28 before putting this live onto the website

ZJvandeWeg:
@Yndira-E @joepavitt the last commit, b31a9ca, bumps the upstream package and decreased the elements' height by default to 200px as it automatically puts the top-left node close to the screenspace top-left (sorry for the awkward description).

I believe this PR is now ready to be merged. The cascading changes for other pages I'll pick up later today to ensure that PR is highly focussed and can serve as an example for the writers.

ZJvandeWeg:
Validated the deploy preview, and going ahead with merging.

PR #2014, opened 2024-05-06
core-node-docs: trigger node

sumitshinde-84:
## Checklist

### DIFF
--- a/src/_includes/core-nodes/trigger-use-case.md
+++ b/src/_includes/core-nodes/trigger-use-case.md
@@ -4,7 +4,7 @@ The Trigger node in Node-RED allows for the initiation and repetition of message
 
 ## Inject Node Vs Trigger Node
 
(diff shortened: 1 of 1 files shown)


src/_includes/core-nodes/trigger-use-case.md
```diff
@@ -0,0 +1,52 @@
+## What is the Trigger Node in Node-RED?
+
+The Trigger node in Node-RED facilitates the initiation and repetition of messages at customizable intervals, offering precise control over when messages are sent, their recurrence frequency, and optional delays. This functionality is valuable for automating tasks and efficiently managing communication flow within Node-RED flows.
```
ZJvandeWeg, line 3: I really don't know, but how is this different than the delay node? Can you explain and link to the delay node?

ZJvandeWeg: Nevermind, next paragraph

PR #2045, opened 2024-05-10
Improve docs copy/watch handling

knolleary:
Recent changes around the netlify part of the website build meant running the build locally didn't always generate all of the product docs.

I believe this was due to a timing issue around the copy_docs script running in parallel to 11ty starting up - a timing issue that was exposed by removing the netlify part of the build.

This PR modifies how the copy_docs/watch_docs scripts operate so that:

1. copy_docs can be run once before 11ty is started
2. a new watch mode in `copy_docs` can be used in parallel to 11ty starting. This only copies things over when changes are made - it does not do an initial copy over.

To achieve this, I've scrapped the `watch_docs` script that used `nodemon` - it was a sledgehammer that would copy the full docs tree over on every change - leading 11ty to rebuild far more than it needed it.

The new watch mode handles individual file changes and only copies/creates/deletes files that have changed.

This is working well locally - but will appreciate some local testing by others before merging.

I haven't (yet) touched the blueprint copy/watch scripts, but if this approach works well for core docs, I'll refactor those scripts to reuse what I've done here.


ZJvandeWeg:
Naive first question: what happens if we just symlink the docs directory?

joepavitt:
> what happens if we just symlink the docs directory?

Probably fine locally, but not sure that's a solution for the GH automation?

ZJvandeWeg:
I assume that's fine @joepavitt, it's just a linux machine in an Azure DC. But the `docs.json` wouldn't be in the directory. Maybe we need to move that upstream?

knolleary:
The copy of files does more than just copy the raw file - it rewrites the yml header to include additional meta-data. So a symlink wouldn't be appropriate. It also wouldn't work with the netlify build as I understand it.

The GH Action doesn't actually build the website - it does all the file copies then checks in the result to the `live` branch. This is the branch that netlify watches and pulls from then runs the build without having to run the doc copying.

joepavitt (approved):

Issue #1968, opened 2024-04-23
Rotating banner for website hello bar


Hasmin-AC:
### Description

Would be great if we could have more than one item to promote in the events banner at the top of the website. 

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #2061, opened 2024-05-14
Rotating event banner

Yndira-E:
## Description

This PR adds a rotating banner feature to the website. Multiple events or announcements can now be displayed at the top of the website, rotating every few seconds. The events data is pulled from a YAML file and the rotation functionality is implemented in JavaScript.

## Related Issue(s)

[#1968](https://github.com/FlowFuse/website/issues/1968)

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,7 @@ on:
       - main
   schedule:
     - cron: "30 9 * * *"
+    - cron: "5 16 * * 3,4"
 jobs:
   build_deploy:
     if: ${{ github.repository == 'FlowFuse/website' }}
--- a/src/_data/events.yaml
+++ b/src/_data/events.yaml
@@ -2,3 +2,4 @@
   title: "Deploy FlowFuse on Industrial IoT with NCD.io"
   buttonText: "Learn more"
   link: "/webinars/2024/deploy-flowfuse-on-industrial-iot-with-ncd-io/"
+  expire: "2024-05-29T16:00:00Z"
--- a/src/_data/events.yaml
+++ b/src/_data/events.yaml
@@ -1,0 +1,4 @@
+- type: "Webinar"
+  title: "Deploy FlowFuse on Industrial IoT with NCD.io"
+  buttonText: "Learn more"
+  link: "/webinars/2024/deploy-flowfuse-on-industrial-iot-with-ncd-io/"
--- a/src/_includes/components/events-banner.njk
+++ b/src/_includes/components/events-banner.njk
@@ -1,15 +1,63 @@
-<script>
-</script>
+{% set colors = ["bg-indigo-700 hover:bg-indigo-600", "bg-indigo-800 hover:bg-indigo-700", "bg-indigo-900 hover:bg-indigo-800"] %}
 
-<a data-action="event-banner" href="/webinars/2024/deploy-flowfuse-on-industrial-iot-with-ncd-io/"
-    class="hidden md:flex text-center w-full py-1 pb-3 sm:py-3 bg-indigo-700 text-indigo-100 text-sm px-2 gap-1 sm:gap-2 flex-col sm:flex-row items-center justify-center hover:bg-indigo-600">
-    <span class="hidden sm:block font-semibold">Webinar</span>
-    <span class="hidden sm:block">-</span>
-    <span class="">     
-        Deploy FlowFuse on Industrial IoT with NCD.io
-    </span>
-    <span class="hidden sm:block">-</span>
-    <span class="flex items-center gap-1 font-medium underline" >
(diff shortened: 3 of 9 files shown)


src/_data/events.yaml
```diff
@@ -0,0 +1,4 @@
+- type: "Webinar"
+  title: "Deploy FlowFuse on Industrial IoT with NCD.io"
+  buttonText: "Learn more"
+  link: "/webinars/2024/deploy-flowfuse-on-industrial-iot-with-ncd-io/"
```
ZJvandeWeg, line 4: Maybe we should add two timestamps. Not before, that is from what point we can show the banner, and and expire. That saves a bunch of hassle each month

src/_includes/components/events-banner.njk
```diff
@@ -1,15 +1,63 @@
...
-    <span class="flex items-center gap-1 font-medium underline" >
-        Learn more {% include "components/icons/chevron-right-sm.svg" %}
-    </span>
-</a>
+        // Show the first banner
+        banners[0].style.visibility = 'visible';
+        changeContainerColor(banners[0]);
+
+        // If there is more than one banner, change the visible banner every 5 seconds
+        if (banners.length > 1) {
```
ZJvandeWeg, line 41: Do we need this clause? A transition from one to the same other should be fine?

src/_includes/components/events-banner.njk
```diff
@@ -1,15 +1,63 @@
...
+                banners[currentBanner].style.animation = 'none'; // reset animation
+                // force reflow, triggering a new animation
+                void banners[currentBanner].offsetWidth;
+                banners[currentBanner].style.animation = 'slideDown 0.5s ease';
+
+                // Change the container color to match the new banner
+                setTimeout(function() {
+                    changeContainerColor(banners[currentBanner]);
+                }, 500); // wait for the slideDown animation to finish
+            }, 7000);
```
ZJvandeWeg, line 60: 7 seconds isn't a lot imo, do we want to push this back?

ZJvandeWeg (changes requested):

src/_includes/components/events-banner.njk
Yndira-E: I've changed it to 10s

src/_includes/components/events-banner.njk
Yndira-E: I think the transition is meant to grab the user's attention, conveying something like _the event you saw in the banner when the page first loaded has changed somehow, take a look, there's something different_  and that's why I added this. I don't think we should be grabbing their attention again when there's nothing new to say.

src/_data/events.yaml
Yndira-E: Great idea ZJ, thanks! I've included an expiration date field. Since we won't be adding events or announcements before the URL is created, I believe it's sufficient for the banner to start showing when the info it's added or updated. The crucial aspect is setting when it will disappear. WDYT?

.github/workflows/build.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
   schedule:
     - cron: "30 9 * * *"
+    - cron: "5 16 * * 3,4"
```
ZJvandeWeg, line 9: Why the last 3,4?

Yndira-E: Wednesday and Thursdays, the possible days for webinars

src/handbook/customer/marketing/website.md
```diff
@@ -12,19 +12,39 @@ navTitle: Marketing - Website
...
+  title: "Deploy FlowFuse on Industrial IoT with NCD.io"
+  buttonText: "Learn more"
+  link: "/webinars/2024/deploy-flowfuse-on-industrial-iot-with-ncd-io/"
+  expire: "2024-05-29T16:00:00Z"
+```
+The `expire` field is used to set the date and time when the event should stop being displayed on the banner. The date and time are set in the ISO 8601 format: `YYYY-MM-DDTHH:MM:SSZ`, and the time is in Coordinated Universal Time (UTC).
+
+For example, `expire: "2024-05-29T15:00:00Z"` means that the event will stop being displayed on the banner at 16:00 UTC on May 29, 2024.
+
+Please note that the website is built once a day at 9:30 AM UTC, and also on Wednesdays and Thursdays at 16:05 UTC. This means that if an event expires at some point during the day, it will still be displayed until the next time the website is built.
```
ZJvandeWeg, line 30: I'd just rebuild it every day at 16:00. This contains a bit too much business logic now tied to webinars.

ZJvandeWeg: This is the only stuff i can complain about, so I'll merge as is @Yndira-E

ZJvandeWeg (approved):

PR #2064, opened 2024-05-15
Update tcp-in-use-case.md

sumitshinde-84:
## Checklist


src/_includes/core-nodes/tcp-in-use-case.md
```diff
@@ -24,6 +24,8 @@ TCP (Transmission Control Protocol) is one of the core protocols of the Internet
         - **delimited by:** Specify the delimiter for splitting incoming data streams. Specify the delimiter for splitting incoming data streams. Commonly, `,`, `\r`, `\n`.
         - **re-attach delimiter:** Enabling this option will reattach the delimiter to its original place.
 
+**Note: The default TCP nodes have been removed from the Node-RED palette in the FlowFuse Cloud due to limitations in routing connections to the container running Node-RED inside the FlowFuse platform**
```
ZJvandeWeg, line 27: Also TCP out? As the connection is initiated by the client, those might work?

sumitshinde-84: I will add a doc on TCP-out today! Yes, it might work, but since the TCP nodes are disabled, we will still need to add this note.

PR #2082, opened 2024-05-21
core-node-docs:write

sumitshinde-84:
## Checklist

### DIFF
--- a/src/_includes/core-nodes/write-file-use-case.md
+++ b/src/_includes/core-nodes/write-file-use-case.md
@@ -1,6 +1,6 @@
-## What is Write File Node in Node-RED?
+## What is the Write File Node in Node-RED?
 
-The "Write File" node in Node-RED is used to write data to a file on the filesystem. It's commonly employed in flows where you need to save data or logs for later analysis or storage. The content to be written is specified using `msg.payload`.
+The "Write File" node in Node-RED is used to write data to a file on the filesystem. It's commonly employed in flows where you need to save data or logs for later analysis or storage. However, In FlowFuse Cloud, the Write File node interacts with a cloud-based storage solution, leveraging AWS S3 for file storage. The content to be written is specified using `msg.payload`.
 
 # Configuring the Write File Node in Node-RED:
 
@@ -21,7 +21,7 @@ The "Write File" node in Node-RED is used to write data to a file on the filesys
 
 - **Encoding:** Specifies the character encoding to be used when writing data to the file, when selecting "set by `msg.encoding`" you can set it dynamically.
 
-*Note: This Write File node is designed for use with a file-storage server in FlowFuse.*
+*Note: The Edge devices running the FlowFuse device agent will use the local filesystem for file storage.*
 
 *Tip: Always use an absolute path for the filename to ensure Node-RED can accurately locate and manipulate the specified file.*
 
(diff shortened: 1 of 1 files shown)


src/_includes/core-nodes/write-file-use-case.md
```diff
@@ -0,0 +1,64 @@
+## What is Write File Node in Node-RED?
+
+The "Write File" node in Node-RED is used to write data to a file on the filesystem. It's commonly employed in flows where you need to save data or logs for later analysis or storage. The content to be written is specified using `msg.payload`.
```
ZJvandeWeg, line 3: How does this work on FlowFuse?

src/_includes/core-nodes/write-file-use-case.md
```diff
@@ -0,0 +1,64 @@
+## What is the Write File Node in Node-RED?
+
+The "Write File" node in Node-RED is used to write data to a file on the filesystem. It's commonly employed in flows where you need to save data or logs for later analysis or storage. However, In FlowFuse Cloud, the Write File node interacts with a cloud-based storage solution, leveraging AWS S3 for file storage. The content to be written is specified using `msg.payload`.
```
ZJvandeWeg, line 3: Same as the Read File node, please explain the full context. Now you'll get people off guard on devices.

sumitshinde-84: Yes, I have added the note at the end of the configuration section about how it will work on devices

Hasmin-AC:
Why am I not able to find this in preview? 

@Yndira-E How are we handling meta title and descriptions in this type of content? Should Sumit add the specific details similar to webpages or blogs?

src/_includes/core-nodes/write-file-use-case.md
```diff
@@ -0,0 +1,62 @@
+## What is the Write File Node in Node-RED?
+
+The "Write File" node in Node-RED is used to write data to a file on the filesystem. It's commonly employed in flows where you need to save data or logs for later analysis or storage. In FlowFuse Cloud, the Write File node interacts with a cloud-based storage solution, leveraging AWS S3 for file storage. However, in the Node-RED instance running on edge devices using the FlowFuse device agent, this node will interact with the local file system of that device. The content to be written is specified using `msg.payload`.
```
Hasmin-AC, line 3:
```suggestion
The "Write File" node in Node-RED is used to write data to a file on the filesystem. It's commonly employed in flows where you need to save data or logs for later analysis or storage. In FlowFuse Cloud, the Write File node interacts with a cloud-based storage solution, leveraging AWS S3 for file storage. However, in the Node-RED instance running on edge devices using the[ FlowFuse device agent](https://flowfuse.com/docs/device-agent/introduction/), this node will interact with the local file system of that device. The content to be written is specified using `msg.payload`.
```

sumitshinde-84: No need to add the domain when interlinking. If we do not add the domain, then the build will test which links are not valid for us.

Yndira-E:
@sumitshinde-84, @Hasmin-AC asked:

> Why am I not able to find this in preview?

I can't find it either, I'm guessing the text in this PR should be added to this file?
Screenshot 2024-06-03 at 17 07 29

If that's the case, it has to do with the name of your file, it should be `file-use-case.md` without the `write`

sumitshinde-84:
> @sumitshinde-84, @Hasmin-AC asked:
> 
> > Why am I not able to find this in preview?
> 
> I can't find it either, I'm guessing the text in this PR should be added to this file? Screenshot 2024-06-03 at 17 07 29
> 
> If that's the case, it has to do with the name of your file, it should be `file-use-case.md` without the `write`

Yes, I need to update the `coreNodes.json` file to include the documentation for the file-write and file-read and other nodes nodes.

Hasmin-AC (approved):

PR #2111, opened 2024-05-28
Big Strategy

gdziuba:
## Description

Update the Company Strategy to reflect a more inspiring objective that trancends industries.  

## Goal
To create messaging that is consistent from customer presentations to Social Media.

## Checklist


ZJvandeWeg:
@gdziuba My understanding was that only the mission statement would be updated, and I don't think this scope of change is warranted to be made right now.

PR #2113, opened 2024-05-29
Add eleventy navigation to /node-red/*

Yndira-E:
## Description

I've installed the [Eleventy Navigation plugin](https://www.11ty.dev/docs/plugins/navigation/) and applied it to the `node-red/*` directory, since the script we use for other nav-bars can't be used in this case due to conflicts with the blog `node-red` tags.

Currently, the only content is in the `/core-nodes/*` directory. To add new content and have it appear in the navigation, the front matter of the content file should include:
```
eleventyNavigation:
        key: UniquePageKey # This will be the name displayed in the navigation and the link text for group children.
```
 
Creating folders and adding content to them won't automatically create groups in the navigation. Instead, the front matter of the files that are part of the same group should include:  
```
eleventyNavigation:
        key: UniquePageKey # This will be the name displayed in the navigation.
        parent: ParentPageKey # This is the key value of the parent/group name. This must correspond to the 'key' of another file.
```

More details can be found in the documentation of the plugin.

## Checklist


src/_includes/layouts/learning-resources.njk
```diff
@@ -0,0 +1,62 @@
...
+<div class="eleventy md:grid md:grid-cols-[265px_1fr_auto] flex flex-col-reverse text-left pb-24 md:max-w-screen-xl m-auto">
+    {% include "./learning-resources-nav.njk" %}
+
+    <div class="max-w-screen-lg mx-auto px-8 pt-8">
+        <!-- Breadcrumbs - Desktop -->
+        <div class="hidden w-full mx-auto md:block">
+            <div class="font-medium border-b pb-1">
+                {% if version %}
+                v{{ version }}
+                {% endif %}
```
ZJvandeWeg, line 14: What's this?

src/_includes/layouts/learning-resources.njk
```diff
@@ -0,0 +1,62 @@
...
+            </div>
+        </div>
+    </div>
+    <div class="border-l">
+        <!-- Breadcrumbs - Mobile -->
+        <div class="block px-8 pt-4 w-full mx-auto md:hidden">
+            <div class="font-medium border-b pb-1">
+                {% if version %}
+                v{{ version }}
+                {% endif %}
```
ZJvandeWeg, line 38: What do you mean `version`?

src/_includes/layouts/learning-resources.njk
```diff
@@ -0,0 +1,62 @@
...
+        <!-- Right side bar -->
+        <div class="sticky top-6 w-full md:w-64 mt-4 md:mt-6 px-8">
+            <h3 class="font-medium border-b pb-1 mb-4">On this page</h3>
+            <ul id="toc" class="text-sm border-b mb-4"></ul>
+            {%- for maintainer in page | pageOwners | ghUsersToTeamMembers(team) -%}
+                {%- if loop.first -%}
+                    <h3 class="font-medium border-b pb-1 mb-4">Page maintainer</h3>
+                {%- endif -%}
+
+                {% renderTeamMember maintainer %}
```
ZJvandeWeg, line 53: ??

src/node-red/core-nodes/nav.njk
```diff
@@ -0,0 +1,15 @@
+---
+pagination:
+    data: coreNodesArray
+    size: 1
+    alias: category
+    addAllPagesToCollections: true
+permalink: "node-red/core-nodes/{{ category.key | lower | replace(\" \", \"\") }}-nav/"
```
ZJvandeWeg, line 7: I'm not sure what this page is, and why we're adding it?

src/_includes/layouts/learning-resources.njk
```diff
@@ -0,0 +1,62 @@
...
+            {%- endfor -%}
+            <div class="text-sm pb-1 text-right mb-4"><a href="{{ page | handbookEditLink }}">Edit this page</a></div>
+            <div class="text-xs pb-1 text-right mb-4 italic">Updated: {% filter shortDate %}{{ updated if updated else page.date }}{% endfilter %}</div>
```
ZJvandeWeg, line 56: Does this work for core nodes? I'm not sure this provides the correct data

ZJvandeWeg (changes requested):

src/_includes/layouts/handbook.njk
```diff
@@ -71,55 +71,5 @@ date: git Last Modified
     </div>
 </div>
 </div>
-<script>
```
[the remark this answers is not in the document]
Yndira-E: I've moved it to the `common-js.njk` partial, since both that script and the partial are only used in the `handbook` and `learning-resources` template

src/node-red/core-nodes/nav.njk
Yndira-E: This is the page that generates the `(coreNodeCategory)-nav.njk`, one per category, so we can assign the eleventyNavigation.key that will be the parent of all the nodes from the same category.

We can't assign a parent if it doesn't corresponds with some file that has the same value as a key.

And since we don't have an index file for each category, we're redirecting to the first child of the group and making sure the empty page isn't generated, only the nav item is added in the navigation, I took the idea from https://www.11ty.dev/docs/plugins/navigation/#overriding-the-url even though it's not an external link.

src/_includes/layouts/learning-resources.njk
Yndira-E: This was in the handbook template and I brought it in here when duplicated the file, I've removed the unused nunjucks blocks.

src/_includes/layouts/learning-resources.njk
Yndira-E: The `page.date` key was omitted, but it was still working. According to the documentation:

> If a date key is omitted from the file, we then look for a YYYY-MM-DD format anywhere in the file path (even folders). If there are multiple dates found, the first is used.

I've added it, but it still shows today's date because all the node templates were modified to implement the eleventy navigation plugin, and it's not picking up the content changes along, but the front matter changes as well.

I've also simplified that line of code to scope it to just the `page.date` since we're not using the `updated` key.

PR #2118, opened 2024-05-30
core-node-docs:mqtt-broker-config

sumitshinde-84:
## Checklist


src/_includes/core-nodes/mqtt-broker-config-use-case.md
```diff
@@ -0,0 +1,51 @@
+## What's the MQTT-broker config Node in Node-RED Used For?
```
Hasmin-AC, line 1:
```suggestion
## What is the MQTT-broker config Node in Node-RED Used For?
```

Hasmin-AC (commented):
Are we not adding an example to this similar to the other pages?

sumitshinde-84:
> Are we not adding an example to this similar to the other pages?

Since it is a config node, we do not need to add an example flow here.

PR #2124, opened 2024-05-31
Adds support details

robmarcer:
We should encourage all contracted customers to user support@flowfuse.com and our live chat to get support rather than intro-ing them directly to Rob. This PR adds that details to the Handbook.


ZJvandeWeg (approved):

zackwasli (commented):
👍

PR #2126, opened 2024-05-31
Update layout for /node-red/*

Yndira-E:
## Description

Changed the layout to `base` instead of `page` so if a title is added, it won't get rendered on the hero section.

## Checklist


ZJvandeWeg:
Missing an example, so the deploy preview isn't valuable. But I'll merge

PR #2130, opened 2024-05-31
Add meta properties to core-nodes template

Yndira-E:
## Description

I've added `meta` properties to the front matter of the templates that generate the core-nodes pages.

This required some updates to the base template, so it could recognise the `title` and `keywords` of each node instead of taking only the data of the last node of each category.

I've added the description and keywords field to the `inject` node in the `coreNodes.json` file to provide a guide, since this is the file where the data needs to be added.

## Checklist

### DIFF
--- a/src/_data/coreNodes.json
+++ b/src/_data/coreNodes.json
@@ -3,7 +3,9 @@
         {
             "xpath": "inject",
             "name": "Inject",
-            "file": "20-inject"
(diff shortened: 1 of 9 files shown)


ZJvandeWeg (approved):

Yndira-E:
Thanks @ZJvandeWeg that was quick!! 🚀

PR #2136, opened 2024-06-04
Updated help desk processes

robmarcer:
Added various new processes for using Hubspot help desk.


robmarcer:
@ZJvandeWeg I also noted that this section of the handbook seems to be outside of the Customer section. Feels like that's wrong?

ZJvandeWeg:
Feel free to move it @robmarcer :)

robmarcer:
That was in iteration for another day but agreed it should be done. :-)

PR #2139, opened 2024-06-04
Create white paper template and manufacturing page

Yndira-E:
## Description

Created a template for white papers and added the first one: manufacturing.

The template supports markdown files, so the future white papers can be added in the same manner we currently add blog posts.

## Checklist


.eleventy.js
```diff
@@ -89,6 +89,12 @@ module.exports = function(eleventyConfig) {
         return `<span class="ff-tooltip" data-tooltip="${text}">${content}</span><span></span>`
     });
 
+    eleventyConfig.addPairedShortcode("blueCard", function(content) {
+        const md = new markdownIt();
```
ZJvandeWeg, line 93: Please look at line 12

src/_includes/layouts/right-nav.njk
```diff
@@ -0,0 +1,50 @@
...
+    document.querySelectorAll("[data-{{nav}}] a").forEach((link) => {
+        try {
+            if (!link.target && window.location.host !== new URL(link.href).host) {
+                link.target = "_blank";
+            }
+        } catch {
+            // Swallow errors
+        }
+    })
+</script>
```
ZJvandeWeg, line 50: Can you ask your editor to format docs with a trailing CR?

.eleventy.js
```diff
@@ -89,6 +89,12 @@ module.exports = function(eleventyConfig) {
...
+        let markdownContent = md.render(content);
```
ZJvandeWeg, line 94:
```suggestion
        const markdownContent = markdownIt.render(content);
```

ZJvandeWeg: @Yndira-E Can you test this?

Yndira-E: I tried this approach before, but kept getting this error on the terminal
```
[11ty] Problem writing Eleventy templates: (more in DEBUG output)
[11ty] 1. Having trouble rendering liquid template ./src/whitepaper/accelerating-innovation-in-manufacturing-with-flowfuse.md (via TemplateContentRenderError)
[11ty] 2. markdownIt.render is not a function
```
Apparently this won't work because `render` is an instance method of markdownIt, not a static method. This means we need to create an instance of markdownIt before we can use the render method.

Here are the docs: https://markdown-it.github.io/markdown-it/

PR #2158, opened 2024-06-05
Add build task that skips image pipeline for CI to use

knolleary:
## Description

There's no need to run the full image pipeline when running the website tests. This PR adds the ability to run a 'full' build, but with the image pipeline disabled.

This should speed things along.


knolleary:
I think that's successful:

image

ZJvandeWeg (approved):

PR #2163, opened 2024-06-07
blog: how node-red works

sumitshinde-84:
## Checklist

### DIFF
--- a/src/blog/2024/06/node-red-architecture.md
+++ b/src/blog/2024/06/node-red-architecture.md
@@ -1,0 +1,102 @@
+---
+title: How Does Node-RED works
+subtitle: Detailed guide explaining how Node-RED works behind the scenes
+description: A comprehensive exploration of Node-RED's internal architecture and workings, from its runtime environment to flow management.
+date: 2024-06-11
+authors: ["sumit-shinde"]
+image:
+tags:
+   - posts
+   - node-red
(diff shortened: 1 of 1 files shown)


ZJvandeWeg:
@knolleary Can you pick up the review on this one?

knolleary:
I have started reviewing the post, but as I started adding comments, I felt the need to ask a higher level question about this post: How detailed do we want this post to be?

As it stands, I think it is too high level in some areas and then super specific in other areas that don't necessarily support the overall title of 'how node-red works'. There are some areas where it talks about the flow file, for example, which overlooks the fact the storage is pluggable and FlowFuse, for example, uses a database rather than file to manage the flows. There's an opportunity for this post to talk about how Node-RED works, whilst showing how FlowFuse makes use of it.

Given this is a post about how node-red works, I would be happy to do a redraft it and put my name on it as a piece coming from the creator of Node-RED itself - something we can market it with.

sumitshinde-84:
We aim for this post to be as detailed as possible. Despite my efforts to find external sources, I couldn't locate any. Therefore, I went through the 'red.js' file and explored the '.node-red' folder structure to gain a high-level understanding, but it didn't meet my expectations. Please feel free to revise it and publish it under your name. As the creator of Node-RED, you can write better.

PR #2164, opened 2024-06-07
marketing: Update CB tracking script

ZJvandeWeg:
## Description

As Clearbit updated their sign up flow now they've been acquired by Hubspot, we somehow are forced to update our JS snippet. This change is an attempt to make that happen.

## Checklist

### DIFF
--- a/src/_includes/layouts/base.njk
+++ b/src/_includes/layouts/base.njk
@@ -449,7 +449,7 @@ eleventyComputed:
 </script>
 
 <!-- Clearbit Tracking -->
-<script type="text/plain" data-category="analytics" async src="https://tag.clearbitscripts.com/v1/pk_a40db34c3e210290335fe7ad7f2a7197/tags.js" referrerpolicy="strict-origin-when-cross-origin"></script>
+<script src="https://tag.clearbitscripts.com/v1/pk_d62bf91706baf1c207f6010e8f08b8ce/tags.js" referrerpolicy="strict-origin-when-cross-origin"></script>
 
 <!-- HubSpot Tracking -->
 {% set hubspotJS %}{% include "hubspot.js" %}{% endset %}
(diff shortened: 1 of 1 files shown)


src/_includes/layouts/base.njk
```diff
@@ -449,7 +449,7 @@ eleventyComputed:
 </script>
 
 <!-- Clearbit Tracking -->
-<script type="text/plain" data-category="analytics" async src="https://tag.clearbitscripts.com/v1/pk_a40db34c3e210290335fe7ad7f2a7197/tags.js" referrerpolicy="strict-origin-when-cross-origin"></script>
+<script src="https://tag.clearbitscripts.com/v1/pk_d62bf91706baf1c207f6010e8f08b8ce/tags.js" referrerpolicy="strict-origin-when-cross-origin"></script>
```
sumitshinde-84, line 452: should we add this script inside the head tag as clearbit suggested :https://help.clearbit.com/hc/en-us/articles/360063559573-Install-Clearbit-JavaScript/#01H7ZCC768ZKF9JCG7Q5KKPCN8#h_01FSW7RBT5DWT91T917CF6EX8T, but adding it into the head tag will cause delay in  rendering the visible content of the page

sumitshinde-84, line 452:
```suggestion
<script async src="https://tag.clearbitscripts.com/v1/pk_d62bf91706baf1c207f6010e8f08b8ce/tags.js" referrerpolicy="strict-origin-when-cross-origin"></script>
```

sumitshinde-84: We should load it asynchronously as we were doing before.

sumitshinde-84: no need, rendering it after body tag is perfect

PR #2183, opened 2024-06-12
Update modbus article

sumitshinde-84:
## Description

I have updated the Modbus article to add a section on building a Modbus server and sending data to that server. Additionally, I have updated the article to use Node-RED Dashboard 2.0 and added the final flow at the end.

## Checklist


Hasmin-AC:
why not create a new article with the title such as using dashabord 2.0 for Modbus?

sumitshinde-84:
Thank you for the suggestion, but writing a blog on this topic might not be appropriate. We are not actually using Dashboard 2.0 with Modbus. Instead, we use protocols with Node-RED. Dashboards are just used for visualizing data in our setup.

Hasmin-AC (approved):

PR #2187, opened 2024-06-13
Merge pricing features data files

Yndira-E:
## Description

In an effort to streamline the process of adding, editing, or removing features from the pricing page, I have consolidated all features into a single YAML file.

Previously, we maintained separate files for cloud and self-hosted features. This meant that features available on both platforms had to be duplicated across both files. With this update, there's no longer a need for such duplication. If a feature is available on both platforms, it simply needs to be tagged appropriately and include both cloud and self-hosted values within the same item.

## Checklist


ZJvandeWeg:
Screenshot 2024-06-15 at 14 03 22
Nice!

ZJvandeWeg (approved):

PR #2243, opened 2024-06-24
Removing meeting calendar from contact us thank you page.

Hasmin-AC:
The meeting link will be sent by the respective AE the lead is assigned to.

## Checklist

### DIFF
--- a/src/thank-you/contact.njk
+++ b/src/thank-you/contact.njk
@@ -3,7 +3,5 @@ title: Thank you for contacting us
 subtitle: Schedule a call now
 description: 
 downloadFollowUp: false
-hubspot:
-    script: "hubspot/hs-book-meeting.njk"
-    dataSrc: https://meetings-eu1.hubspot.com/michael-davis?embed=true
+
 ---


src/thank-you/contact.njk
```diff
@@ -3,7 +3,5 @@ title: Thank you for contacting us
 subtitle: Schedule a call now
 description: 
 downloadFollowUp: false
-hubspot:
-    script: "hubspot/hs-book-meeting.njk"
-    dataSrc: https://meetings-eu1.hubspot.com/michael-davis?embed=true
```
Yndira-E, line 8: Hi @Hasmin-AC removing this won't do the trick since the template is expecting this info. We need to change the template.

I'm opening an issue for this one, we also need to decide how the new layout it's going to be.

Hasmin-AC: I see. So yes we'll need a new template. How about under the thank you messages, we have a block for the latest on the. blog and the Cloud sign-up CTA? Below that will be the logos...

Yndira-E:
Replaced by https://github.com/FlowFuse/website/pull/2247

Issue #2275, opened 2024-07-02
Change our product menu on main nav and create landing pages accordingly


Yndira-E:
### Description

Change our current product menu to be something like:

Product:
- “ecosystem” TBD
  - FlowFuse
  - Device Agent
  - Dashboard
  - Project Nodes

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Yndira-E:
@joepavitt This description is very simple. Do you think it accurately describes what you intended to convey during our call?

Feel free to add your reasoning behind this proposal or anything else you see fit.

While creating this, a question came up: where would the "FlowFuse" option take users? The [FlowFuse cloud landing page](https://flowfuse.com/product/cloud/) we recently created? Should we add two items instead: one for cloud and another for self-hosted?

joepavitt:
I think the "FlowFuse" option would be the "Features" page - maybe? Or we list "Dashboard", "Agent", "Project Nodes" as a subsection of that navigation instead?

Yndira-E:
Like this?

Product:
- Features
  - Dashboard
  - Device Agent
  - Project Nodes

joepavitt:
Was more thinking:

- Features
- Customer Stories
- Blueprints
- FlowFuse Offerings:
    - FlowFuse Cloud
    - Dashboard
    - Device Agent
    - Project Nodes

Yndira-E:
Ok, and isn't it FlowFuse self-hosted part of our offering?

For device Agent and Project nodes we don't have landing pages, but could link to docs lik in this PR: https://github.com/FlowFuse/website/pull/2302 WDYT?

joepavitt:
The Device Agent page is being created at the moment in https://github.com/FlowFuse/website/issues/2236

For Project Nodes - docs is fine. Even docs for Device Agent (until the marketing page is ready) is also reasonable.

> FlowFuse self-hosted part of our offering?

Yes, that generally falls under the "Features" page really as it's common across all of FlowFuse (Cloud & Self-Hosted). The "Cloud" offering gets a particular call out because it's easier to sign up, etc.

I'd be open to having a "FlowFuse Self-Hosted" there too, maybe even changing it to:

- FlowFuse
    - Features
    - FlowFuse Cloud
- Customer Stories
- Blueprints
- Extras:
    - Dashboard
    - Device Agent
    - Project Nodes

PR #2344, opened 2024-07-14
Main nav: Update product menu

Yndira-E:
## Description

Updated the product menu to include device agent, project nodes and group items accordingly

## Related Issue(s)

[#2275](https://github.com/FlowFuse/website/issues/2275)

## Checklist


Yndira-E:
I didn't remove the roadmap. I don't see why it would be beneficial, and it's the only link we have to the changelog. However, I would be open to changing it to link directly to the changelog if we decide the roadmap page adds no value.

joepavitt:
Changing it to Changelog would be a great idea imo.

The "Roadmap" is not maintained and makes us look a little poorly planned. The changeling shows activity, and is updated with shiny new things regularly

Yndira-E:
> The "Roadmap" is not maintained

Just to clarify, it pulls the issues from GH, do you mean it's not maintained on GH? Is it pulling from the wrong place? Or did you think it had to be updated manually?

src/_includes/layouts/base.njk
```diff
@@ -188,14 +188,22 @@ eleventyComputed:
...
+                        <ul class="narrow sm:grid sm:grid-flow-col sm:grid-rows-6 sm:pr-1 align-left sm:auto-rows-auto items-center">
+                            {% navoption "FlowFuse", null, 1, null, true, "pl-3 title border-l-2" %}{% endnavoption %}
+                            <ul class="sub-menu grid grid-rows-subgrid row-span-3 ml-7 auto-rows-auto border-l-2">
+                                {% navoption "Features", "/product/features/", 2, "star" %}{% endnavoption %}
+                                {% navoption "FlowFuse Cloud", "/product/cloud/", 2, "cloud" %}{% endnavoption %}
+                                {% navoption "Roadmap", "/product/roadmap/", 2, "map" %}{% endnavoption %}
+                            </ul>
+                            {% navoption "Customer Stories", "/customer-stories/", 1, "presentation" %}{% endnavoption %}
+                            {% navoption "Blueprint Library", "/blueprints/", 1, "building-library" %}{% endnavoption %}
+                            {% navoption "Extras", null, 1, null, true, "pl-3 title border-l-2" %}{% endnavoption %}
```
ZJvandeWeg, line 200: Extra's? That's a bit weird IMO, and devalues the items below

joepavitt:
> Do you mean it's not maintained on GH? Is it pulling from the wrong place? Or did you think it had to be updated manually?

It's running from an old project board that is no longer actively maintained: https://github.com/orgs/FlowFuse/projects/5 and in fact, uses a soon-to-be-deprecated GH project type too.

> Extra's? That's a bit weird IMO, and devalues the items below

Good point, not sure of an alternative phrasing? Or we don't put them into a sub-section, and surface it all at the same level?

ZJvandeWeg:
Not calling them extra's, but components? Does that work? These are all elements of what's needed to provide our end to end data service, but the word 'services' doesn't fit for me.

joepavitt:
> These are all elements of what's needed to provide our end to end data service, but the word 'services' doesn't fit for me.

The other term I'd used in a blog was "offerings"? But it feels odd as a heading in a navigation menu

Yndira-E:
How about:

**Product:**
- **Platform**
  - Features
  - FlowFuse Cloud
  - Changelog
- Customer Stories
- Blueprints
- **Components**
  - Dashboard
  - Device Agent
  - Project Nodes
  
I changed the title "FlowFuse" to "Platform" because it seemed to me that what's outside of it is not FlowFuse... Maybe Blueprints should also be under Platform? And I think Components could work.

joepavitt:
Yep, agree with that Yndira, also agree: blueprints

Yndira-E:
I implemented the changes yesterday, including moving Blueprints under Platform. Let me know if anything else needs adjustment. Thanks!

joepavitt:
Feels a little odd having "Customer Stories" on the right, I know it balances the menu, but it gets a little lost, how about having it underneath the "Product" like this instead?:

Screenshot 2024-07-17 at 11 56 48

Yndira-E:
Yeah, I had it like that at first, but it looks out of balance and still feels off, like the orphan one, hehe. So, I thought to place it to the right; it would still be the orphan but at least balanced... I even thought about removing it, we also have a link to customer stories under resources.

PR #2349, opened 2024-07-15
Pricing: Add contact-us form in cloud section

Yndira-E:
## Description

Brought back the contact-us form for the Cloud section and updated button link and CTA

## Related Issue(s)

https://github.com/FlowFuse/customer/issues/225

## Checklist

### DIFF
--- a/src/pricing/index.njk
+++ b/src/pricing/index.njk
@@ -416,13 +416,13 @@ hubspot:
         <h2 class="w-full text-center"><span class="text-indigo-600">Contact us</span> for pricing
         </h2>
         <div>
-            {% hubspotForm hubspot.dedicated.formId, hubspot.dedicated.cta, hubspot.reference, 'contactUsForPricingForm' %}
+            {% hubspotForm hubspot.dedicated.formId, hubspot.dedicated.cta, hubspot.reference, 'dedicatedPricingForm' %}
         </div>
     </div>
 </div>
 
 <!-- Frequently Asked Questions -->
-<div class="md:px-14 max-w-screen-lg mx-auto">
+<div class="px-6 md:px-14 max-w-screen-lg mx-auto">
     {% include "faq.njk" %}
 </div>
 <!-- Scroll to anchor -->
--- a/src/pricing/index.njk
+++ b/src/pricing/index.njk
@@ -27,7 +27,7 @@ hubspot:
     script: "hubspot/hs-form.njk"
     cloud:
         formId: d43ab79b-d075-4b63-ba5e-bf971ca0b86e
-        cta: "cta-any-questions"
+        cta: "cta-contact-us"
     selfHosted:
         formId: 41e858e1-6756-45be-9082-3980237fa229
         cta: "cta-request-trial-license"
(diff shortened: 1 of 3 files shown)


src/pricing/index.njk
```diff
@@ -375,6 +378,40 @@ hubspot:
...
+        <div class="container max-w-5xl m-auto grid grid-cols-1 md:grid-cols-2 px-6 pt-16 gap-x-16 gap-y-6">
+            <div>
+                <h3 class="w-full text-center md:text-left">
+                    Contact Us
+                </h3>
+                <p>
+                    To learn more about our plans, discuss your specific needs, or get advice on the best fit for your project, please fill out the form and we will get back to you as soon as possible.
+                </p>
+            </div>
+            {% if hubspot.cloud %}
```
ZJvandeWeg, line 393: Why an if statement for a variable that's always set?

src/pricing/index.njk
```diff
@@ -375,6 +378,40 @@ hubspot:
...
+            {% set formId = hubspot.cloud.formId %}
+            {% set cta = hubspot.cloud.cta %}
```
ZJvandeWeg, line 395: I don't understand why we alias these variables?

src/pricing/index.njk
```diff
@@ -375,6 +378,40 @@ hubspot:
...
+            <script>
+                function displayHubSpotForm2() {
```
ZJvandeWeg, line 397: What's the difference with line 447 and this form? They seem very much the same, and in that case we should DRY the code?

ZJvandeWeg (changes requested):

src/pricing/index.njk
Yndira-E: You're right, I've removed the if statements

src/pricing/index.njk
Yndira-E: This was inherited from a previous approach that I couldn't make work as I wanted a while ago. I have now managed to make it work.

src/pricing/index.njk
Yndira-E: I was initially struggling with the function name. If I reused the hs-form.njk in the same file, it only added the last form. However, I've since refactored the HubSpot script, and this is no longer necessary.

src/pricing/index.njk
```diff
@@ -25,13 +25,15 @@ meta:
             To allow business to standardize on FlowFuse and Node-RED for all their data acquisition, visualization, and interaction it's possible to purchase a flat-fee license to deploy an unlimited amount of Node-RED instances through FlowFuse.
 hubspot:
     script: "hubspot/hs-form.njk"
+    cloud:
+        formId: d43ab79b-d075-4b63-ba5e-bf971ca0b86e
+        cta: "cta-any-questions"
```
Yndira-E, line 30:
```suggestion
        cta: "cta-contact-us"
```

src/pricing/index.njk
ZJvandeWeg: At which point do we make it a pairedShortcode?

ZJvandeWeg (changes requested):

ZJvandeWeg (approved):

PR #2350, opened 2024-07-15
Features: Replace book a demo with contact us

Yndira-E:
## Description

At the bottom of `/features` we have this section:

Screenshot 2024-07-15 at 14 05 47

Those numbers represent the amount of clicks from the past 90 days (according to PostHog after the cookie banner was set up properly).

This changes the copy and the link from `book a demo` to the `contact us` page to see if we can improve that number since filling out a form can seem less intimidating than scheduling a call.

## Related Issue(s)

https://github.com/FlowFuse/customer/issues/225

## Checklist

### DIFF
--- a/src/_includes/common-cta.njk
+++ b/src/_includes/common-cta.njk
@@ -35,15 +35,15 @@
         </div>
         <div class="features-card">
             <div>
-                <label class="flex items-center text-center justify-center font-medium text-2xl text-gray-800 mt-5">Contact Sales</label>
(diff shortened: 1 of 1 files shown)


ZJvandeWeg:
@Yndira-E What about reducing the number of CTA's, and remove the "install now" and make the "On-Premise" card have the "contact us" CTA?

ZJvandeWeg:
@Yndira-E Merging this PR, but I think having just 2 CTA's is better and less confusing. Please create an issue to reduce the CTA's

PR #2394, opened 2024-07-26
ensure to add trailing slash at the end of relative paths and the ff …

sumitshinde-84:
…signup page

## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -670,25 +670,30 @@ module.exports = function(eleventyConfig) {
     }
 
     markdownLib.renderer.rules.link_open = (tokens, idx, options, env, self) => {
-    const hrefIndex = tokens[idx].attrIndex('href');
-    if (hrefIndex >= 0) {
-        let href = tokens[idx].attrs[hrefIndex][1];
-        let classIndex = tokens[idx].attrIndex('class');
-        
-        // Exclude the link if it has the class 'header-anchor'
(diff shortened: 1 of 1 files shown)


joepavitt (approved):

.eleventy.js
```diff
@@ -670,25 +670,30 @@ module.exports = function(eleventyConfig) {
...
+            }
+    
+            // Ensure the URL has a trailing slash, but do not update if it contains a '#' or ends with '.md' or https
+            if (!href.endsWith('/') && !href.includes('#') && !href.endsWith('.md') && !href.includes('https')) {
+                href += '/';
+            }
+    
+            // make sure to update the flowfuse signup URL
+            if (href.includes('https://app.flowfuse.com/')) {
+                href += '/';
```
knolleary, line 690: @sumitshinde-84 isn't this going to be adding an extra `/` regardless of whether it already ends with `/`? You've added this outside of any check to guard against that.

joepavitt: good point - I miss that

sumitshinde-84: Yes, I should first check if it already has the trailing slash or not, but I also want to make sure that the host, like app.flowfuse.com, will not have the slash added at the end.

sumitshinde-84: i will fix it now !

PR #2395, opened 2024-07-26
update the trailing slash logic

sumitshinde-84:
## Checklist

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -686,9 +686,10 @@ module.exports = function(eleventyConfig) {
             }
     
             // make sure to update the flowfuse signup URL
-            if (href.includes('https://app.flowfuse.com/')) {
+            if (!href.endsWith('/') && href.includes('https://app.flowfuse.com/account/create')) {
                 href += '/';
             }
+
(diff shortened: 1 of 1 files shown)


knolleary:
Why do we need to rewrite the sign up url? The canonical version is `https://app.flowfuse.com/account/create` without the trailing `/`.

sumitshinde-84:
I've just checked that our sign-up page is not indexed—neither `https://app.flowfuse.com/account/create` nor `https://app.flowfuse.com/account/create/`.

sumitshinde-84:
@joepavitt, that's why I think our sign-up page doesn't have any views ( you mentioned that in the channel).

joepavitt:
> that's why I think our sign-up page doesn't have any clicks( you mentioned that in the channel).

That's a PostHog problem - we're getting the traffic/views, PostHog just isn't recording it accurately.

sumitshinde-84:
Okay! For now, I am removing the logic to include the / for the sign-up page.

knolleary (approved):

PR #2398, opened 2024-07-29
Add nr-assistant to list of unmanaged releases

Steve-Mcl:
## Description

Adds nr-assistant to list of unmanaged releases

## Checklist


knolleary (changes requested):
This page is need of more updates to drop the flowforge naming. We shouldn't add even more wrong instances

Steve-Mcl:
Oh my, copy paste at its worst - sorry I missed that.

I'll update the full Page.

knolleary (approved):

PR #2399, opened 2024-07-29
Handbook and docs: Adjust max-width for pre elements

Yndira-E:
## Description

This PR addresses an issue with the max-width of the parent element of `<pre>` elements in our handbook and docs pages. Previously, when the content inside the `<pre>` tag was too long, it would stretch the parent element beyond the max-width of the grandparent element, pushing the nav bars off the screen if they were too small.

This change adjusts the max-width property of the parent of `<pre>` elements to prevent this issue and removes extra space at the bottom of the code blocks.

## Checklist

### DIFF
--- a/src/css/style.nohero.css
+++ b/src/css/style.nohero.css
@@ -42,10 +42,6 @@
     line-height: 2.9rem;
 }
 
-.prose code::after {
-    content: '' !important;
-}
-
 .prose code::before {
     content: '' !important;
 }
--- a/src/css/style.handbook.css
+++ b/src/css/style.handbook.css
@@ -9,7 +9,7 @@
     }
 
     .handbook .prose {
-        max-width: none;
+        @apply max-w-full;
     }
 
     .handbook-nav {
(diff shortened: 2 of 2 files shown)


knolleary:
Do you have an example page I can easily compare a before/after of this change?

Yndira-E:
Sure, you can compare the changes on this page: https://flowfuse.com/docs/device-agent/install/

Since the preview doesn't show docs pages, I'm providing screenshots for comparison:

Here is how it currently looks live (you can see the right nav is off the screen):
Current Live Version

And here is how it looks with the changes from this PR, at the same screen size:
PR Version

knolleary:
Tnanks for the screenshots - saves me building :)

Looks good

knolleary (approved):

src/css/style.nohero.css
```diff
@@ -42,10 +42,6 @@
     line-height: 2.9rem;
 }
 
-.prose code::after {
```
knolleary, line 45: Just spotted this has caused all inline code blocks to be rendered with a ` at the end - because this rule was previously forcing it to be removed.

image

Yndira-E: Thanks for the heads up, Nick. I appreciate you pointing this out. Here's the PR that addresses the issue: https://github.com/FlowFuse/website/pull/2402

PR #2407, opened 2024-07-30
Replace the svg with d2 gif

sumitshinde-84:
## Checklist


knolleary:
Does our image pipeline know how to optimise animated GIFs? The files attached to this PR are huge but not seen them get optimised. We may want to manually optimise them.

sumitshinde-84:
I thought it would compress all the images, but I think it is only compressing the images in the blog posts. I will compress the this images manually and update the pr.

src/product/dashboard.njk
```diff
@@ -14,8 +14,10 @@ meta:
             <h4 class="font-light mt-2 text-gray-500">
                 Build visually appealing and useful UIs in minutes!
             </h4>
-            <div class="justify-center items-center m-auto w-full flex mt-10">
-                {% image "../images/dashboard/dashboard.svg", "FlowFuse Dashboard", [1920] %}
+            <div class="justify-center items-center m-auto w-full flex mt-10 gap-3 dashboard-image">
+                {% image "../images/dashboard/bar-chart.gif", "bar chart",[1920]%}
```
Yndira-E, line 18: `[1920]%}`

That's meant to set the final size of the images. The entire image with the 3 charts was 1920px, but 1024px would have been enough. You should add the size of each chart individually now that each one is a separate file.

src/product/dashboard.njk
```diff
@@ -14,8 +14,10 @@ meta:
             <h4 class="font-light mt-2 text-gray-500">
                 Build visually appealing and useful UIs in minutes!
             </h4>
-            <div class="justify-center items-center m-auto w-full flex mt-10">
-                {% image "../images/dashboard/dashboard.svg", "FlowFuse Dashboard", [1920] %}
+            <div class="justify-center items-center m-auto w-full flex mt-10 gap-3 dashboard-image">
```
Yndira-E, line 17: I think on mobile, we should place the two smaller images next to each other, with the third one below them. What do you think?

sumitshinde-84: image

same like this?

Yndira-E: Yes!

Yndira-E (approved):

PR #2422, opened 2024-08-02
Using CoAP with Node-RED for Low-Power Devices

### LINKED ISSUE #2420: Blog: CoAp with Node-RED
### Description

The blog will include what CoAP is, when to use it, and why it matters. Then, it will show how we can utilize it with a device agent.

### Epic/Story

_No response_

sumitshinde-84:
## Checklist


Steve-Mcl (commented):
Overall, the article reads will and is nicely written however I have 2 concerns with releasing this material.

1. Unrealistic "Why does it matter" argument.  A Rasperry Pi can only run on batteries for a matter of hours (obviously depending on battery size, RPi type and workload) - however we are talking 10s of hours not days or weeks - just feels a little unrealistic and I dont think we should be publishing that kind of documentation if I am honest.

2. The table of data rates reads great but there is no reference to how we came to these numbers or proofs.  The numbers "feel" a bit "perfect" and without any reference to WireShark or our methods, I am hesitant to include such detail.   I would either hope to see hard facts to accompany this OR a slimmed down "potential benefits" kind of data

sumitshinde-84:
Yes, I agree. There is one more reason why I am not considering publishing this article: the node is not working properly right now, as some of its features do not function correctly.

sumitshinde-84:
That's why I haven't followed up for its review. I think we should close this PR.

Steve-Mcl:
Happy to close it if the node package is not working well.

Out of interest, what issues did you find? Did you note if the node package had an issue tracker and what the authors responsiveness was?

sumitshinde-84:
> Happy to close it if the node package is not working well.
> 
> Out of interest, what issues did you find? Did you note if the node package had an issue tracker and what the authors responsiveness was?

The author is active in merging dependency bot PRs but has not been adding new features or fixing existing issues. The CoAP node has several problems: it doesn't support IPv6 (even though there is an option for it), the agent functionality isn't working, and it lacks support for multicast and conformability( even this option is provided. Additionally, there's a field to set the timeout, but it doesn't work properly—when you try to input a value, it gets wiped out.

PR #2473, opened 2024-08-15
Add feedback form to Node-RED docs

Yndira-E:
## Description

This PR adds [this HS form](https://app-eu1.hubspot.com/submissions/26586079/form/e1f3e0ff-6e19-484d-88dc-9e4bb6e54c03/performance?redirectUrl=https%3A%2F%2Fapp-eu1.hubspot.com%2Fforms%2F26586079%2F) to the docs on `/node-red/*` to collect feedback.

@ZJvandeWeg, please check the form setup as well. Currently, it isn't part of any workflow; it just notifies us if we receive a submission. Maybe Sumit should be notified as well?

## Checklist

### DIFF
--- a/src/_includes/layouts/learning-resources.njk
+++ b/src/_includes/layouts/learning-resources.njk
@@ -1,6 +1,8 @@
 ---
 layout: layouts/base.njk
 date: git Last Modified
+hubspot:
+    formId: e1f3e0ff-6e19-484d-88dc-9e4bb6e54c03
 ---
 <div class="w-full bg-white/50 nohero">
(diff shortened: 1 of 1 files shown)


ZJvandeWeg:
@Yndira-E What do you think about adding this one instead: https://app-eu1.hubspot.com/feedback/26586079/satisfaction/2/submissions?published=true&dateRange=this-year&startDate=2024-01-01&endDate=2024-08-15&page=1&count=10&sortDir=descending&sortBy=submittedAt

Yndira-E:
> @Yndira-E What do you think about adding this one instead:

I like it, even if we don't get their emails every time. I've already set it up, so there's no need for this PR. I'm closing it now.

PR #2501, opened 2024-08-27
Update customer-success.md

zackwasli:
adding churned details

## Description

adding churned details

## Related Issue(s)

n/a

## Checklist

### DIFF
--- a/src/handbook/customer/customer-success.md
+++ b/src/handbook/customer/customer-success.md
@@ -174,3 +174,8 @@ and getting more value from the product
 1. Data collection -- Provide space and time for the customer to provide feedback
 on the product and experience. Allow them to request feature ideas, or ask for bug
 fixes.
+
+## Churned customers
+
+When an annually-billed customer chooses not to renew their subscription, the assigned Account Executive should login into Stripe and cancel the subscription. This will effectively close the account on the platform and prevent invoices to continue being sent.
+


zackwasli:
Still verifying what happens to platform account when Stripe subscription is cancelled

src/handbook/customer/customer-success.md
```diff
@@ -174,3 +174,8 @@ and getting more value from the product
 1. Data collection -- Provide space and time for the customer to provide feedback
 on the product and experience. Allow them to request feature ideas, or ask for bug
 fixes.
+
+## Churned customers
+
+When an annually-billed customer chooses not to renew their subscription, the assigned Account Executive should login into Stripe and cancel the subscription. This will effectively close the account on the platform and prevent invoices to continue being sent.
```
ZJvandeWeg, line 180: @joepavitt @knolleary Can we verify this would work for FFC too? Self-hosted this process is fine.

knolleary: Annually billed customers on FFC are in manually billing mode. This means there is no tie between stripe and the platform - so cancelling the Stripe subscription will have no effect on the platform - the team will remain active.

ZJvandeWeg: @knolleary Can you propose a way to ensure the team is no longer active?

knolleary: Have raised #2527 to track and update the docs

PR #2510, opened 2024-08-28
Update customer-success.md

robmarcer:
We need to close tickets once a customer no longer needs help. Currently this is mostly being done by Rob but that does not scale.

### DIFF
--- a/src/handbook/customer/customer-success.md
+++ b/src/handbook/customer/customer-success.md
@@ -104,9 +104,12 @@ You can also add comments to a ticket which will only be visible to our team. Th
 
 When replying to the ticket one should have the intent to always reply with links to FlowFuse articles or documentation. If these resources are missing, please start or update where applicable.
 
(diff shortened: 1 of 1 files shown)


src/handbook/customer/customer-success.md
```diff
@@ -104,10 +104,13 @@ You can also add comments to a ticket which will only be visible to our team. Th
 
 When replying to the ticket one should have the intent to always reply with links to FlowFuse articles or documentation. If these resources are missing, please start or update where applicable.
 
-Where a conversation needs to happen with regards to a ticket between FlowFuse team members we should use the comments feature rather than having a parallel conversation is Slack so that anyone who wants to help with the ticket can see all relevant information in one place.
+Where a conversation needs to happen with regards to a ticket between FlowFuse team members we should use the comments feature int eh ticket rather than having a parallel conversation in Slack, so that anyone who wants to help with the ticket can see all relevant information in one place.
```
ZJvandeWeg, line 107:
```suggestion
Where a conversation needs to happen with regards to a ticket between FlowFuse team members, we should use the comments feature in the ticket rather than having a parallel conversation in Slack. This allows anyone who wants to help with the ticket can see all relevant information in one place.
```

PR #2631, opened 2024-09-30
Blog : Using FlowFuse Project Nodes for Faster and More Efficient Com…

sumitshinde-84:
…munication

## Checklist


Steve-Mcl (approved):
This is a good clean well written article.

If I had to say one thing it would be not calling out when/why to use broadcast vs direct.  That could be a follow up article (a part 2, deeper dive) that refers back to this article as reference.

I will approve, you can decide to update or follow up.

Steve-Mcl:
PS, @sumitshinde-84 If something is ready for review, please mark it as read for review.

Steve-Mcl:
Sorry, forgot to add, the date probably needs updating (it is currently marked as Sept 13th)

Steve-Mcl (changes requested):
You seem to have an image added to the `09/images` directory

sumitshinde-84:
> You seem to have an image added to the `09/images` directory

Removed Now !

Steve-Mcl (approved):
all good.

Will probably need the date updating one last time before publish.

Issue #2620, opened 2024-09-26
Iterate on handbook.njk layout


Yndira-E:
### Description

Move the search bar to take less space
Unify the layout so `/docs` `/handbook` and `/node-red/*` use the same one

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

PR #2633, opened 2024-09-30
Iterate on handbook.njk layout

Yndira-E:
## Description

- Moved the search bar to take less space
- Unified the layout so `/docs` `/handbook` and `/node-red/*` use the same one
- Made `Edit this page` and `Updated:` visible only on desktop

## Related Issue(s)

https://github.com/FlowFuse/website/issues/2620

## Checklist


src/node-red/node-red.json
```diff
@@ -1,5 +1,6 @@
 {
-    "layout": "layouts/learning-resources.njk",
+    "layout": "layouts/handbook.njk",
```
ZJvandeWeg, line 2: Feels dirty if the filename is `node-red.json`?

Yndira-E: Fixed! https://github.com/FlowFuse/website/pull/2635

PR #2678, opened 2024-10-17
LinkedIn Converstion Tracking

Hasmin-AC:
## Checklist


ZJvandeWeg:
@Hasmin-AC Did you test this on the deployment preview?

Yndira-E:
@Hasmin-AC I was under the impression this would be implemented through Google Tag Manager, so there wouldn’t be a need to add more code.

Would you mind sharing the documentation with the instructions? Just to confirm it’s in the right place.

Hasmin-AC:
I was able to connect HS to LinkedIn so we can track conversion from HS now. Looks like we don't need this code.

Issue #2638, opened 2024-10-02
Blueprints Revamp


joepavitt:
### Description

After we complete https://github.com/FlowFuse/website/issues/2637 we also want to revamp the "Blueprints" page.

### New features

- Ability to submit your own blueprints, which should include a file uploader, details about you as an author/FF customer. Emphasis here should be on the fact that Blueprints are authored by _companies_, not individuals.
- Highlight on each of the blueprints who the author is, e.g. "FlowFuse" or "Company A"

### Considerations

- Aesthetics should be similar to the ["Integrations"](https://github.com/FlowFuse/website/issues/2637) page.
- Filtering by category/use-case should also be possible, e.g "Machine Learning", "Industrial", "Getting Started".
- Each Blueprint should have a "Deploy to FlowFuse" button, which will enable them to quickly get started on FF Cloud.
- Each Blueprint will have a standalone page too, which more detail and description. On this page emphasis should be made on the flow itself, maybe with a preview (flow viewer?), as well as a profile of the Blueprint on the right-hand side, containing information like the Author, published date, categories etc.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

joepavitt:
@gstout52 what's the appropriate language here flor a CTA around submitting your own blueprints?

joepavitt:
@gstout52 could you also share some loose copy that you'd like to see on the Upload/Submit page. Should be stressing something around professional standards, will be reviewed by FF team?

Also need to know what we want in the upload form?

- Name
- E-Mail for correspondance
- Company
- `.flow.json`
- Integrations (third-party node packages) needed to run it
- Screenshots?

gstout52:
> @gstout52 could you also share some loose copy that you'd like to see on the Upload/Submit page. Should be stressing something around professional standards, will be reviewed by FF team?
> 
> Also need to know what we want in the upload form?
> 
> * Name
> * E-Mail for correspondance
> * Company
> * `.flow.json`
> * Integrations (third-party node packages) needed to run it
> * Screenshots?

Let's go with these:

*CTA button: Submit Blueprint

*Sample Copy, feel free to rework:

The best Blueprints come from Node-RED flows our community has put to use. Share your Blueprints to help the FlowFuse community build a best-in-class library of commercially useful content and build recognition of your company as top Node-RED experts. Join the ranks of professionals who elevate Node-RED with FlowFuse!

[...]

PR #2684, opened 2024-10-18
Blueprints: Add page for customers to submit their own flows

joepavitt:
## Description

Add a CTA on the Blueprint page to "Submit Your Own" (CTA to be checked over with @gstout52) 

Screenshot 2024-10-18 at 17 04 56

Also adds a placeholder page for "Submit Your Blueprint". 

- Need to configure the correct form in Hubspot
- Check over the language/content with @gstout52, in particular, what we want this text to say:
Screenshot 2024-10-18 at 17 03 52

## Related Issue(s)

Merge after #2683 
Part of https://github.com/FlowFuse/website/issues/2638


joepavitt:
Screenshot 2024-10-22 at 16 16 26

@Yndira-E I'm thinking we may want a graphic of some kind on that left-side to jazz it up a little? 

The form is long, and we need the informative text, so I went split view.

joepavitt:
> Were you aiming for it to resemble how the blueprints page will eventually look?

I was, and I agree with your assessment

Yndira-E:
> Don't think it needs to block anything here though as a first iteration

I agree

joepavitt:
@gstout52 you can check out deployment preview here: https://deploy-preview-2684--flowforge-website.netlify.app/

src/blueprints/submit.njk
```diff
@@ -0,0 +1,67 @@
...
+  <div class="w-full py-10 text-center ">
+    <div class="container m-auto max-w-lg md:max-w-4xl">
+        <h2>Existing Blueprints</h2>
+        <p>Here are a few examples of Blueprints from our collection.You can take a look at the full collection in our <a href="/blueprints">Blueprint Library</a></p>
+        <ul class="flex flex-wrap">
+            {%- asyncEach item in blueprints -%}
+            <li class="w-full md:w-1/3 my-2 px-3 pb-6 border-b mt-6 flex flex-col justify-between group">
+                <a href="{{ item.url }}" class="relative w-full flex flex-col hover:no-underline">
+                    <div class="ff-blog-tile overflow-hidden">
+                        <div class="w-full h-auto transition-transform group-hover:scale-105">
```
Yndira-E, line 56: On hover, the corners become sharp again, I've created some classes that I've been using to address this
```suggestion
                        <div class="w-full h-auto transition-transform ff-image-cover ff-image-rounded scale">
```

Screenshot 2024-10-23 at 12 05 55

src/blueprints/submit.njk
```diff
@@ -0,0 +1,67 @@
...
+                            {% tileImage item, "./images/og-blog.jpg", "Image with logo and the slogan: Elevate Node-RED with Flowfuse", 285 %}
+                        </div>
+                        <label class="absolute w-full left-0 bottom-0 bg-indigo-700 text-white font-medium text-left p-2">{{ item.data.title }}</label>
```
Yndira-E, line 59: And this one adds round corners to the title background
```suggestion
                        <label class="absolute w-full left-0 bottom-0 bg-indigo-700 text-white font-medium text-left p-2 rounded-b-lg">{{ item.data.title }}</label>
```

Yndira-E:
And this:
Screenshot 2024-10-23 at 11 55 06

It should look like this:
Screenshot 2024-10-23 at 11 55 19

I know this class is causing the issue:
Screenshot 2024-10-23 at 11 55 27

But it seems like you didn't introduce it in this PR, so I'm not sure which of your changes is causing this to happen.

gstout52 (approved):
Looks great! Let's make the Job Title field mandatory. This will be useful for users browsing the page and our own outreach efforts.

sumitshinde-84:
@joepavitt, I think we need to add one more field to upload the `package.json` file that will include the nodes the blueprint is using.

joepavitt:
@sumitshinde-84 we can form one ourselves here, they're going to want to export from Node-RED, and having to generate a `package.json` is friction we can get them to avoid. We have their e-mail address to be able to check which nodes are used, etc. if they do not detail it in the existing form submission.

PR #2698, opened 2024-10-23
Update Redirect statergy

sumitshinde-84:
## Description

Currently, we are using a meta refresh for redirects, which is not considered SEO-friendly and negatively impact SEO. Also meta-refersh method does not clearly indicate a redirect to search engine crawlers bot, as they typically recognize only 301 and 302 status codes. Meta refresh redirects return a 200 status code, which does not efficiently transfer SEO equity from the old URL to the new one.

In the current pull request, we have implemented server-side redirects that correctly respond with a 301 status code. However, the hyperlink test on github actions is failing, maybe because it expects a 200 status code.

To check the redirects, use the following tool: 

tool : https://www.redirect-checker.org/
test with meta refresh : `https://flowfuse.com/blog/first-deploy/`
test with server side redirect: `https://deploy-preview-2698--flowforge-website.netlify.app`

For further reading on the impact of meta refresh redirects on SEO, refer to these resources:  

https://risify.app/meta-refresh-redirects/#negative-impact-of-meta-refresh-redirects-on-seo
https://seodity.com/blog/redirects-guidline/#html-redirects
https://help.ahrefs.com/en/articles/2433739-what-is-meta-refresh-redirect-and-why-is-it-considered-a-critical-issue
https://firozahmad.com/what-is-meta-refresh-redirection/

## Checklist


Yndira-E:
I suggest that in the next iteration, we automate the generation of the redirects list instead of writing them directly in the file. This can be based on the approach described in this article: https://www.aleksandrhovhannisyan.com/blog/eleventy-netlify-redirects/

sumitshinde-84:
> @sumitshinde-84 I think it’s working now. The errors don’t seem to be related to the `redirects` file but to the .md files that still have the old links.

Yes, it was working, but the href link tests are failing. It's related to the package because the current package isn't working well with server-side redirects.

src/redirects.njk
```diff
@@ -0,0 +1,75 @@
+---
+permalink: /_redirects
```
ZJvandeWeg, line 2: I'm confused why this file exists, and the file without this extension?

sumitshinde-84: It’s not ready yet; the review was requested automatically.

.eleventy.js
```diff
@@ -85,6 +85,9 @@ module.exports = function(eleventyConfig) {
         "src/public/": "/",
     });
 
+    // Copy the contents of the `_redirects` to the output folder
+    eleventyConfig.addPassthroughCopy("src/_redirects");
```
ZJvandeWeg, line 89: Don't think this is needed if we keep the redirect.njk?

src/redirects.njk
ZJvandeWeg: I think it already works on my machine?

ZJvandeWeg: Keep this file, remove the changes in .eleventy.js and it's ready?

.eleventy.js
sumitshinde-84: yes, i have removed it

Yndira-E (approved):

PR #2719, opened 2024-10-28
Linking to updated overview PDF

Hasmin-AC:
## Checklist

### DIFF
--- a/src/thank-you/download-platform-overview.njk
+++ b/src/thank-you/download-platform-overview.njk
@@ -1,6 +1,6 @@
 ---
 title: Thank you for your interest! 
-subtitle: "<a href='https://26586079.fs1.hubspotusercontent-eu1.net/hubfs/26586079/FlowFuse%20Platform%20Overview.pdf'>Click here</a> to access the The FlowFuse Paltform Overview"
+subtitle: "<a href='https://26586079.fs1.hubspotusercontent-eu1.net/hubfs/26586079/FlowFuse%20Platform%20Overview-October%202024.pdf'>Click here</a> to access the The FlowFuse Platform Overview"
 description: "We've also sent a copy to your inbox" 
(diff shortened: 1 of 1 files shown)


src/thank-you/download-platform-overview.njk
```diff
@@ -1,6 +1,6 @@
 ---
 title: Thank you for your interest! 
-subtitle: "<a href='https://26586079.fs1.hubspotusercontent-eu1.net/hubfs/26586079/FlowFuse%20Platform%20Overview.pdf'>Click here</a> to access the The FlowFuse Paltform Overview"
+subtitle: "<a href='https://26586079.fs1.hubspotusercontent-eu1.net/hubfs/26586079/FlowFuse%20Platform%20Overview-October%202024.pdf'>Click here</a> to access the The FlowFuse Platform Overview"
```
Yndira-E, line 3: Hi Hasmin, I noticed you used one of the PDF files I had shared earlier, but it wasn’t the final version—I was just checking quality with that one. Would you mind downloading the latest PDF from Google Docs to ensure it’s the most up-to-date? Thank you!

Hasmin-AC: Did you have a PDF? when I converted [this](https://docs.google.com/document/d/1VtLQMZvvGDqTjiplUoUtdLwEeS9W5F7Y-42N2zGmRFc/edit?tab=t.0), it looks pretty much the same as the one I uploaded.

Yndira-E: I know there were tweaks after that, the only one that I can recall that would be obvious is the order of text/image in the section where two points share a screenshot

Hasmin-AC: replaced the file on the same location.

PR #2747, opened 2024-11-06
November Webinar : FlowFuse MQTT Broker

Hasmin-AC:
@Yndira-E can you please create the image banner for this webinar and add to the page? thanks! 

## Checklist


src/webinars/2024/FlowFuse-MQTT-Broker-for-Industrial-Transformation
Yndira-E, line 1: @Hasmin-AC The reason it isn't showing is that the file doesn't have an extension; it needs to have .md at the end of the name. Also, the names should all be in lowercase.

Yndira-E: I've just fixed it and added the image

Yndira-E (approved):

PR #2778, opened 2024-11-14
Update engagements.md

zackwasli:
## Description

Noticed that the handbook doesn't match the HS templates we're currently using. The templates say:

"For any additional capacity required beyond the included capacity of Node-RED instances and edge
devices, add-on blocks of 10 instances or 10 edge devices are available for purchase. Pricing for these
add-ons is detailed in the Products section of this quote."

So we should include the add ons but for 0 quantity as shown here: https://26586079.hs-sites-eu1.com/GvD59WWUq7Ci

Or change the templates to include all product pricing for reference.

I prefer this proposed change (the example I shared) since per product pricing can be deal dependent and is easier to change from the product page when quoting. Also, template terms can not be changed for one offs, the changes need to be made globally and then switched back per deal (not ideal).

## Checklist

### DIFF
--- a/src/handbook/customer/sales/engagements.md
+++ b/src/handbook/customer/sales/engagements.md
@@ -12,23 +12,7 @@ Clearly written, easy to understand quotes are crucial for customers to understa
 
 **What to Quote**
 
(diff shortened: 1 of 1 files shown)


gstout52:
I'm going to defer to @ZJvandeWeg and @MichaelBDavis on this one - want to make sure everyone who writes quotes is aligned on the approach.

ZJvandeWeg:
I like it, thanks.

PR #2779, opened 2024-11-14
feature table reorg

zackwasli:
## Description

We should re-organize our feature table to match our current messaging framework, which is:

Manage
Scale
Secure
Collaborate

Additionally, we need to add a box for "BAA for HIPAA compliance" under Secure, but only checked for Enterprise.

Please note, current commit does not contain necessary changes yet. 

## Checklist

### DIFF
--- a/src/_data/pricingFeatures.yaml
+++ b/src/_data/pricingFeatures.yaml
@@ -165,6 +165,12 @@ sections:
         selfHostedValues: [ null, 'check' ]
         tags: [ 'cloud', 'self-hosted' ]
         info: "<p>FlowFuse can be configured to run with your own Single Sign-On (SSO) provider. Allow users to access FlowFuse with a single set of login credentials, improving security, user experience, and save them time.</p>"
+      - id: hipaa
+        label: BAA for Hipaa
+        cloudValues: [ null, null, 'check' ]
+        selfHostedValues: [ null, 'check' ]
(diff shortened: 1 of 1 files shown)


ZJvandeWeg:
@Yndira-E LMK WDYT.

Yndira-E:
Looks good! but that only covers

> we need to add a box for "BAA for HIPAA compliance" under Secure, but only checked for Enterprise.

For the re-organizing part, would this make sense?

### Manage
- Node-RED Hosting
- Hosted Instances
- Edge Devices
- Instance Monitoring
- Persistent files
- Persistent Context
- Custom Hostnames

### Scale
- Edge Development
- Device Management
- Device Group Management
- MQTT Broker
- High Availability
- DevOps Pipelines

### Secure
- Audit Log
- Role-Based Access Control
- Endpoint Security
- Two-factor authentication
- Certified Nodes
- Single Sign-On (SSO)

### Collaborate
- Team members
- Access to Blueprints
- Node-RED Dashboard
- Personalised Multi-user Dashboards
- Seamless Data Stream

zackwasli:
That looks good, maybe Seamless data stream should go into scale? It's one of the most sought after features, so I want to be careful not to bury it. But I understand that we don't want to take too much away from the Collaborate column, so I'm not sure

Yndira-E:
Done! I’ve already reorganized the features table, including the feature added by ZJ.

PR #2795, opened 2024-11-21
Home page: Add IT/OT section

Yndira-E:
## Description

This PR adds a new section with a diagram showcasing FlowFuse as IT/OT middleware. The section links to the `manufacturing solution`, which has been renamed to `IT/OT middleware` to align better with the other solutions.

## Related Issue(s)

#2796

## Checklist


src/images/home/ff_ot_it.png
joepavitt, line 1: not a fan of the middle gradients clashing here. Looks a little unfinished. Maybe gradient the stokes rather than the fills?

Yndira-E: Have you had a chance to see it in context? The idea is that each space/color represents a different area, and they merge together in the middle rather than clash. Both fade to white at the center, and there’s no stroke. Instead, I’ve used an inner shadow to create an engraved effect, which ties in with the wheel style in the section above.

joepavitt: I appreciate that, just visually, I felt it uncomfortable to look at.

Yndira-E: Ok, I get that the style isn't exactly your cup of tea 😉 I think conceptually it’s alright, though, right?

Yndira-E: Thanks for the feedback Joe, I've just pushed the updated image

Hasmin-AC (approved):

joepavitt (approved):

PR #2806, opened 2024-11-25
Improve font-size for new lead h1

joepavitt:
## Description

- Adds line break to the new tagline to improve readability
- Improves the font-sizing on smaller screens to make the `<h1/>` stand out more

Think we could follow this up with a clean up of that leading hero section too, as it doesn't scream "home page" to me at the moment.

### DIFF
--- a/src/_data/site.json
+++ b/src/_data/site.json
@@ -8,7 +8,7 @@
         "d2fe": "https://boards.greenhouse.io/flowfuse/jobs/5185319004"
     },
     "messaging": {
-        "tagLine": "The Power of Integration at Industrial Scale",
+        "tagLine": "The Power of Integration</br>at Industrial Scale",
         "subtitle": "Supercharge your industrial operations with tools built for seamless collaboration, effortless deployment, and scalable solutions. FlowFuse simplifies development, accelerates delivery, and ensures reliability—all in one platform",
         "title": "Enterprise-ready Node-RED for Rapid Industrial Process Improvement",
         "keywords": "Node-RED, Application Development, IoT, IIoT, Low-Code, Open-Source, open source"
--- a/src/index.njk
+++ b/src/index.njk
@@ -12,8 +12,8 @@ hubspot:
     <div class="max-w-md sm:max-w-screen-lg mt-12 mx-auto">
         <div class="container m-auto text-left max-w-screen-lg lg:flex lg:flex-row lg:items-start lg:justify-between lg:gap-6">
             <div class="max-w-xl mx-auto lg:mx-0 lg:my-auto lg:py-10 lg:max-w-[600px]">
-                <h1 class="font-medium leading-tight max-w-xl m-auto lg:m-0 text-center lg:text-left text-indigo-800 text-4xl">
-                    {{ site.messaging.tagLine }}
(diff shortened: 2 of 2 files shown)


src/_data/site.json
```diff
@@ -8,7 +8,7 @@
         "d2fe": "https://boards.greenhouse.io/flowfuse/jobs/5185319004"
     },
     "messaging": {
-        "tagLine": "The Power of Integration at Industrial Scale",
+        "tagLine": "The Power of Integration</br>at Industrial Scale",
```
Yndira-E, line 11: The `</br>` element makes it lose the space between words for the `<title>`
Screenshot 2024-11-25 at 11 25 44

Maybe it would be better to change the `max-width` on the `index` file

src/index.njk
```diff
@@ -12,8 +12,8 @@ hubspot:
     <div class="max-w-md sm:max-w-screen-lg mt-12 mx-auto">
         <div class="container m-auto text-left max-w-screen-lg lg:flex lg:flex-row lg:items-start lg:justify-between lg:gap-6">
             <div class="max-w-xl mx-auto lg:mx-0 lg:my-auto lg:py-10 lg:max-w-[600px]">
-                <h1 class="font-medium leading-tight max-w-xl m-auto lg:m-0 text-center lg:text-left text-indigo-800 text-4xl">
-                    {{ site.messaging.tagLine }}
+                <h1 class="font-medium leading-tight max-w-xl m-auto lg:m-0 text-center lg:text-left text-indigo-800 text-4xl sm:text-5xl lg:text-4xl">
```
Yndira-E, line 15: From `4xl` to `5xl`, I didn’t expect the change to be so significant. It looks enormous—I don’t think this is necessary...

joepavitt: It is large, yes, I wanted it somewhere between, but this was still better than `4xl` imo at that screen size, makes it read as an actual hero-line, rather than just a smaller header. Happy to be overruled though

Yndira-E:
> Think we could follow this up with a clean up of that leading hero section too, as it doesn't scream "home page" to me at the moment.

What elements would make it feel more like a 'home page' to you?

It's definitely open to revision. The current layout follows a standard hero structure with a headline, brief description, CTAs, supporting visuals (the video), and social proof, but happy to explore any suggestions!

src/index.njk
Yndira-E: Great! I did try it locally, and it didn’t seem to work for me. I’m glad it was as simple as that.

src/_data/site.json
Yndira-E: Great! I did try it locally, and it didn’t seem to work for me. I’m glad it was as simple as that.

src/index.njk
Yndira-E: Let's leave this for a separate PR if we decide to revisit the `h1` in future iterations.

joepavitt:
> What elements would make it feel more like a 'home page' to you?

Think we can emphasise the H1 better, it's currently styled the same as our other headers on that page, and reads very formal/enterprisey/boring.

If you consider:

- https://buildship.com/
- https://n8n.io/
- https://posthog.com/
- https://about.gitlab.com/

All of these have more emphasis on the single sales-pitch tag line

PR #2846, opened 2024-11-27
/integrations Page

### LINKED ISSUE #2637: "Integrations" Page
### Description

We want to render an equivalent of [Node-RED Flows Directory](https://flows.nodered.org/search?type=node) at `/integrations`.

- Each rendered node should also highlight the author, and whether or not it is "FlowFuse Certified".
- The page should offer search functionality
- The page should offer pre-built filters for categories, and to filter down to "FlowFuse Certified" nodes only.

The contents of this new page will replace https://flowfuse.com/certified-nodes/, and so below the tile searching/filtering we should also have a section about the certification program, and how you can submit requests for particular nodes to be certified.

### Inspirations

- https://buildship.com/integrations
- https://n8n.io/integrations/
- https://zapier.com/apps
- https://cloud.ibm.com/catalog

### Technical Details

- Will be hosted at flowfuse.com/integrations
- Will be building on the work of adding categorisation to the contents on flows.nodered.org in [this issue](https://github.com/FlowFuse/node-red/issues/24)

### Epic/Story

_No response_

joepavitt:
## Description

Screenshot 2024-11-29 at 13 45 31

- Adds a new `/integrations` page
- Includes client-side filtering and searching
- Utilises an API we are hosting in an FF NR instance that wraps the catalogue and includes "Certified" details
- Ended up venturing away from the planned design a little due to long package names which didn't work nicely with the inline certified tick, so took inspiration from https://nuxt.com/modules?orderBy=asc&q=&category=CMS instead

### Todo:

- Add Loading placeholder for the list on first load

#### Nice to haves

- Where possible, in place of say `@flowfuse` as the owner, we should be rendering company logos, similar for other authors too.
- "New" tag, similar to "Certified" which helps justify lower download numbers, but also flags new content and the active community

#### Next Steps

- Add standalone pages for each integration, have already added the `integrations` collection in that can be used here. See https://www.11ty.dev/docs/pages-from-data/ for documentation
- Add CTA to submit requests for "Certified Nodes" and more details about the Certified Nodes program - possibly replace the `/certified-nodes` content?

## Related Issue(s)

Closes #2637


joepavitt:
Load + Hover State Demo (FYI @Yndira-E):
 
https://github.com/user-attachments/assets/162de0ad-748d-4c80-a87a-ab6ebf1dec58

Yndira-E:
> Load + Hover State Demo

Love it Joe! It looks great!

Yndira-E:
I think this was supposed to replace the `/certified-nodes/` page, right? If that's the case, there are a couple of files to delete, URLs to update, and maybe a redirect to add as well. Were you planning to address that in a different PR?

joepavitt:
> I think this was supposed to replace the /certified-nodes/ page, right? 

See:

> Add CTA to submit requests for "Certified Nodes" and more details about the Certified Nodes program - possibly replace the /certified-nodes content?

In the "Up Next" section, this will be `/integrations`, as it covers _more_ than just certified nodes, next we'll replace the certified nodes page with more about the program and offer user the chance to request certifications

Yndira-E:
I think it's great, Joe. The only thing missing (as you already noticed) is the mobile version. I checked the examples you provided, and a dropdown seems to be the common solution for the filtering section.

One thing I'm hesitant about is the behavior when I filter one category, get a result, and then click on a second category—my result disappears if there's no match. I can't think of a better alternative, though. I considered keeping the previous result, but that wouldn't provide obvious feedback and might give the impression that it's not working. Just sharing my thoughts—not a blocker or anything like that.

joepavitt:
> The only thing missing (as you already noticed) is the mobile version. I checked the examples you provided, and a dropdown seems to be the common solution for the filtering section.

For quick iteration, I'll be hiding the filters all together and just showing the search bar on mobile

joepavitt:
@Yndira-E responsivness added

src/integrations/index.njk
```diff
@@ -0,0 +1,367 @@
...
+            <placeholder-bar class="block h-5 w-3/4"></placeholder-bar>
+        </ul>
+    </div>
+    <div class="grow">
+        <input id="search-catalogue" class="catalogue-search" type="text" placeholder="Search Integrations" onkeyup="filterCatalogue()" onchange="filterCatalogue()"/>
+        <div class="catalogue-meta">
+            <div id="count-container" style="display: none;"><span id="integrations-count">X</span> Integrations</div>
+            <div id="count-placeholder"><span id="integrations-count">Loading...</div>
+        </div>
+        <ul id="integration-tiles-placeholder" class="grid grid-cols-2 lg:grid-cols-3 gap-3">
```
Yndira-E, line 346:
```suggestion
        <ul id="integration-tiles-placeholder" class="grid grid-cols-1 sm:grid-cols-2 lg:grid-cols-3 gap-3">
```

src/integrations/index.njk
```diff
@@ -0,0 +1,367 @@
...
+            <integration-tile-placeholder></integration-tile-placeholder>
+            <integration-tile-placeholder></integration-tile-placeholder>
+            <integration-tile-placeholder></integration-tile-placeholder>
+        </ul>
+        <ul id="integration-tiles" class="grid grid-cols-2 lg:grid-cols-3 gap-3"></ul>
```
Yndira-E, line 351:
```suggestion
        <ul id="integration-tiles" class="grid grid-cols-1 sm:grid-cols-2 lg:grid-cols-3 gap-3"></ul>
```

src/integrations/index.njk
```diff
@@ -0,0 +1,367 @@
...
+                <label class="inline-flex gap-1 items-center" for="catalogue-filter-certified">FlowFuse Certified <certified-icon /></label>
+            </li>
+        </ul>
+        <label>Categories</label>
+        <ul id="catalogue-filter--categories"></ul>
+        <ul id="catalogue-filter--placeholder">
+            <placeholder-bar class="block h-5 w-3/4"></placeholder-bar>
+        </ul>
+    </div>
+    <div class="grow">
```
Yndira-E, line 340:
```suggestion
    <div class="grow  max-md:max-w-lg mx-auto">
```

Yndira-E (commented):
I've added a few suggestions to make it a single column in small screens since there's not enough room for the description.
Screenshot 2024-11-29 at 16 12 34

joepavitt:
All suggestions resolved/committed

gstout52:
Looks great! Excellent work, all!

PR #2851, opened 2024-11-28
Contact us Thank you page update

Hasmin-AC:
## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated

PR #2853, opened 2024-11-29
/thank-you/contact/: Add stories block

Yndira-E:
## Description

This PR introduces a partial with the most recent customer stories.
The block was added to the `/thank-you/contact/` page.
The subtitles are now grey instead of indigo for increased contrast with `<a>` elements.

## Related Issue(s)

This is a follow-up to https://github.com/FlowFuse/website/pull/2851

## Checklist

### DIFF
--- a/src/_includes/explore-more-content.njk
+++ b/src/_includes/explore-more-content.njk
@@ -1,6 +1,7 @@
 {% if readingResources and readingResources == "stories" %}
  {% include "stories-block.njk" %}
-{% else %}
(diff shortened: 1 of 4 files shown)


Hasmin-AC:
@Yndira-E can we also have the blogs/webinar/newsletter block as well?

Yndira-E:
Sure, if we just add it, it would look like this:
image

Seems like the first row and the second are the same thing, but with different styling... Maybe the last section could use a title?

Yndira-E:
@Hasmin-AC title added!

Hasmin-AC (approved):

PR #2929, opened 2025-01-16
Update security.txt canonical

Yndira-E:
## Description

After merging https://github.com/FlowFuse/website/pull/2928/files I realised the canonical was still pointing to `flowforge.com` it has a redirect, so it was still working, but it's better if we avoid the unnecessary redirect.

## Checklist


hardillb (approved):

hardillb:
Good catch, I was on auto pilot this morning

PR #2944, opened 2025-01-24
docs(handbook): Introduce maintenance description

ppawlowski:
## Description

This pull request introduces `Infrastructure Maintenance` page to the handbook. The goal of this page is to provide high level documentation on steps required to peroform successfull maintenance of various components of FlowFuse Cloud infrastructure.

## Related Issue(s)

https://github.com/FlowFuse/CloudProject/issues/624

## Checklist


src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,76 @@
...
+* at least two weeks' notice should be provided to customers
+* maintenance should not overlap with [scheduled release](../releases/process.md) week
+
+## Customers communication
+
+Once the maintenance date is set, customers should be informed about the upcoming maintenance. The communication should include:
+* the scope of the maintenance
+* the date and time of the maintenance
+* expected duration of the maintenance
+* self-care actions which customers can/should take before the maintenance date
```
ZJvandeWeg, line 34: How can they prevent this? The way **our** infra is set up influences all this, and we never get this from Heroku et al. Why do we get this from FlowFuse?

hardillb: It's fundamental to how kubernetes works.

You can only have a version mismatch of +- 2 (e.g. 1.24 to 1.26) between the core kubernetes api and the version of the kubelet managing the node the pods are running on.

So to upgrade kubernetes version you upgrade the core api, then you need to create new nodes, and then migrate the running pods to those new nodes. The migration requires stopping the instance on the old node and starting it on the new version.

In theory we could start the instance, but that does end up with 2 instances running for a short while, which requires planning in the flow design (around caching or state management)

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,76 @@
...
+navTitle: Infrastructure Maintenance
+---
+
+Periodically, FlowFuse performs maintenance on FlowFuse Cloud resources. This page describes how we manage and execute maintenance tasks around the infrastructure.
+
+# Maintenance scope
+
+The maintenance most often involves updates to the following resources:
+* underlying cluster hosts
+* Kubernetes versions
```
ZJvandeWeg, line 11: These seem highly predictable, when do we schedule these?

src/handbook/development/ops/maintenance.md
ZJvandeWeg: @hardillb My point is that other compute providers just restart, even cloud providers. Somehow we don't and send emails regularly to customers asking them to restart instances, and we don't even tell them which ones.

src/handbook/development/ops/maintenance.md
ppawlowski: That is true. Right now try to go out from the extended support version. Whole AWS EKS release calendar is available [here](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html#kubernetes-release-calendar) . Once we catch up with versions, we can plan migrations in advance.

ZJvandeWeg: If I read that correctly, that means that one restart every year would make this hassle go away?

ppawlowski: That is most likely true, once we reach `1.31` version. The upcoming maintenance will bump the version to `1.30` - we will have to run another maintenance before the 25th of July 2025 if we want to stay within standard support (and don't pay extra for the extended one).

ZJvandeWeg: @ppawlowski I would in that case split this document into "general maintainance", and "Kubernetes Nodes Upgrades". The latter is predictable and I suspect we want to do this twice a year?

Also, I suspect all new hosts will use 1.32 like a week after the release on EKS? And we effectively start draining as soon as these nodes go live?

ppawlowski: I have to disagree - maintenance work (any kind), by definition, is something that should be planned ahead. This is in contrast to an incident, which should be [handled differently](https://flowfuse.com/handbook/development/ops/incident-response/). To be fair I do not see any benefit of splitting the document. Moreover, I tried to structure it in a way, where we could easily extend it with other maintenance tasks. 

> Also, I suspect all new hosts will use 1.32 like a week after the release on EKS? And we effectively start draining as soon as these nodes go live?

No, that is not possible. Workers cannot use kubernetes version newer than the one which is currently running on control planes. We have to upgrade control planes first, before we can create node groups in exact same kubernetes version.

hardillb (approved):

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Planning
+
+The maintenance process is initiated by the Operations Team. The process is planned in advance and communicated to the customers. The maintenance window is scheduled to minimize the impact on the team and customers.
+
+Within a team, a maintenance date should be agreed upon, and the team should be informed about the maintenance window.
+
+While picking the correct date and time for maintenance, the following should be considered:
+* at least two weeks' notice should be provided to customers
+* maintenance should not overlap with [scheduled release](../releases/process.md) week
```
knolleary, line 26:
```suggestion
* maintenance should not overlap with [scheduled release](../releases/process.md) week
* at a time to minimise disruption to customers; outside of core business hours or the weekend
```

We should scheduled disruptive maintenance to be outside core business hours by default.

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+* at a time to minimise disruption to customers; outside of core business hours or the weekend
+
```
knolleary, line 28:
```suggestion
The maintenance process is initiated by the Operations Team. The process is planned in advance and communicated to the customers. The maintenance window is scheduled to minimize the impact on the team and customers.

Where the maintenance action will cause disruption to customers, it must be scheduled with the following considerations:
* at least 45 days notice should be provided to customers
* not during the working week
* maintenance should not overlap with [scheduled release](../releases/process.md) week
```

Based on contract negotiations with customers, we have a new base line set of requirements to document

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+## Customers communication
+
+Once the maintenance date is set, customers should be informed about the upcoming maintenance. The communication should include:
```
knolleary, line 31:
```suggestion

Once the maintenance date is set, customers should be informed about the upcoming maintenance. This must be done in co-ordination with Customer team. The communication should include:

```

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Maintenance day
+
+### Kubernetes worker nodes upgrade
+
+On the maintenance day, the Operartions Team should perform the following steps to migrate workloads to the new node group(s):
+* selectively restart workloads still running on old node group(s).
+* monitor the workloads and ensure that:
+  * they were schedulled on correct node group
+  * NodeRED flows started correctly - take corrective action (e.g., redeployment or log analysis) if issues occur
```
knolleary, line 74:
```suggestion
  * Node-RED flows started correctly - take corrective action (e.g., redeployment or log analysis) if issues occur
```

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+See documentation for more details.
+
+## Maintenance day
+
+### Kubernetes worker nodes upgrade
+
+On the maintenance day, the Operartions Team should perform the following steps to migrate workloads to the new node group(s):
+* selectively restart workloads still running on old node group(s).
+* monitor the workloads and ensure that:
+  * they were schedulled on correct node group
```
knolleary, line 73:
```suggestion
  * they were scheduled on correct node group
```

knolleary (changes requested):
Some updates based on negotiated terms with a customer (and a couple typos I'll self-merge)

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+
+The maintenance process is initiated by the Operations Team. The process is planned in advance and communicated to the customers. The maintenance window is scheduled to minimize the impact on the team and customers.
+
+Where the maintenance action will cause disruption to customers, it must be scheduled with the following considerations:
+* at least 45 days notice should be provided to customers
+* not during the working week
+* maintenance should not overlap with [scheduled release](../releases/process.md) week
+## Customers communication
+
+Once the maintenance date is set, customers should be informed about the upcoming maintenance. This must be done in co-ordination with Customer team. The communication should include:
```
ZJvandeWeg, line 28: Let's describe how this process of commucation is done. GitHub issue seems best with a mention to @Esmewieringa, @MichaelBDavis, @gstout52  and myself.

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+
+* the scope of the maintenance
+* the date and time of the maintenance
+* expected duration of the maintenance
+* self-care actions which customers can/should take before the maintenance date
+* other relevant details to help customers to prepare for potential restart of their NodeRED instances
```
ZJvandeWeg, line 34:
```suggestion
* other relevant details to help customers to prepare for potential restart of their Node-RED instances
```

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+
+E-mail template for maintenance communication can be found in the [FlowFuse Cloud Maintenance Communication document](https://docs.google.com/document/d/1Pkd0qifjgs7xv96hfQ5OHnvX5n_WPHyA0AvPd9UUPL0). Adjust the template accordingly.
+
+The notification message should be sent to all FlowFuse Cloud Team Owners who have active instances in the FlowFuse Cloud. To get the contact list, please refere to the [internal documentation](https://docs.google.com/document/d/1s0mXpuuPKl-1U-YVzoTEGBhllUd-UPJjFXTqa0O481c/#heading=h.hing87m0bsmu).
+
+We should sent notification at least twice - two weeks before the maintenance and a reminder one week before the maintenance day.
```
ZJvandeWeg, line 40: Only applies to FFC self-service customers.

ZJvandeWeg (changes requested):

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+
+* the scope of the maintenance
+* the date and time of the maintenance
+* expected duration of the maintenance
+* self-care actions which customers can/should take before the maintenance date
+* other relevant details to help customers to prepare for potential restart of their Node-RED instances
+
+E-mail template for maintenance communication can be found in the [FlowFuse Cloud Maintenance Communication document](https://docs.google.com/document/d/1Pkd0qifjgs7xv96hfQ5OHnvX5n_WPHyA0AvPd9UUPL0). Adjust the template accordingly.
+
+The notification message should be sent to all FlowFuse Cloud Team Owners who have active instances in the FlowFuse Cloud and who are affected by the maintenance. To get the contact list, please refere to the [internal documentation](https://docs.google.com/document/d/1s0mXpuuPKl-1U-YVzoTEGBhllUd-UPJjFXTqa0O481c/#heading=h.hing87m0bsmu).
```
knolleary, line 38:
```suggestion
The notification message should be sent to all FlowFuse Cloud Team Owners who have active instances in the FlowFuse Cloud and who are affected by the maintenance. To get the contact list, please refer to the [internal documentation](https://docs.google.com/document/d/1s0mXpuuPKl-1U-YVzoTEGBhllUd-UPJjFXTqa0O481c/#heading=h.hing87m0bsmu).
```

src/handbook/development/ops/maintenance.md
gstout52: @ppawlowski there isn't a doc describing this process. Here are pieces that should be included in the handbook regarding how communications are sent to customers as well as the internal work necessary to accomplish this:

- Engineering prepares a list of customers who need to receive the email. This email should include team owner email addresses, team names, and the names of instances that need to be restarted. The intent is that it should be very easy for the customer to quickly identify what is needed from them.
- Email copy should be drafted and approved by both Engineering and the Customer Department. Email copy should aim to be informative while being accessible.
- Email copy should include the timeframe of the migration/restart, and this timeframe should be a brief window (for example, a specific two-hour time period).
- If possible, the email should include the recipient's first name.

If at all possible, avoid emailing customers about behind-the-scenes work. When there is a way to provide excellent service within our stated SLA, just take the needed action.

ZJvandeWeg: @ppawlowski Handbook in combination with a GH issue template seems to be the way to go. Note there there's more and more requirements by customers for scheduled downtime and the company needs to properly manage this -- best way to make sure we do things the right way is to make a process and follow it.

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+* Kubernetes versions
+* databases
+* networking components
+
+ 
+# Maintenance process
+
+## Planning
+
+The maintenance process is initiated by the Operations Team. The process is planned in advance and communicated to the customers. The maintenance window is scheduled to minimize the impact on the team and customers.
```
knolleary, line 20:
```suggestion
The maintenance process is initiated by the Operations Team.

A new [Infrastructure task](https://github.com/FlowFuse/CloudProject/issues/new?template=infra-task.yml) is created to co-ordinate all activities regarding the task - including the technical tasks and communications with customers.

The maintenance window is scheduled to minimize the impact on the team and customers.
```

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,77 @@
...
+* self-care actions which customers can/should take before the maintenance date
+* other relevant details to help customers to prepare for potential restart of their Node-RED instances
+
+E-mail template for maintenance communication can be found in the [FlowFuse Cloud Maintenance Communication document](https://docs.google.com/document/d/1Pkd0qifjgs7xv96hfQ5OHnvX5n_WPHyA0AvPd9UUPL0). Adjust the template accordingly.
+
+The notification message should be sent to all FlowFuse Cloud Team Owners who have active instances in the FlowFuse Cloud and who are affected by the maintenance. To get the contact list, please refer to the [internal documentation](https://docs.google.com/document/d/1s0mXpuuPKl-1U-YVzoTEGBhllUd-UPJjFXTqa0O481c/#heading=h.hing87m0bsmu).
+
+We should sent notification at least twice - 45 days before the maintenance and a reminder one week before the maintenance day.
+
+Before sending the initial message, consult with the Operations team to ensure that the information is up-to-date and infrastructure is ready to handle self-care maintenance by customers.
```
knolleary, line 42:
```suggestion
The e-mail template for all maintenance communication can be found in the [FlowFuse Cloud Maintenance Communication document](https://docs.google.com/document/d/1Pkd0qifjgs7xv96hfQ5OHnvX5n_WPHyA0AvPd9UUPL0). Create a copy of the template and link to from the CloudProject issue.

The notification message should be sent to all FlowFuse Cloud Team Owners who have active instances in the FlowFuse Cloud and who are affected by the maintenance. To get the contact list, please refer to the [internal documentation](https://docs.google.com/document/d/1s0mXpuuPKl-1U-YVzoTEGBhllUd-UPJjFXTqa0O481c/#heading=h.hing87m0bsmu).

The Customer team will handle sending the email once the content is agreed with engineering and the target list of customers has been provided.

The first notification should sent at least 45 days before the event. A reminder should be sent one week prior to the event.
```

knolleary (commented):
I have proposed some changes to try to get this first iteration merged. We can continue to refine, but lets get something merged as a base line for improvement.

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,81 @@
...
+The maintenance process is initiated by the Operations Team.
+
+A new [Infrastructure task](https://github.com/FlowFuse/CloudProject/issues/new?template=infra-task.yml) is created to co-ordinate all activities regarding the task - including the technical tasks and communications with customers.
+
+The maintenance window is scheduled to minimize the impact on the team and customers.
+
+Where the maintenance action will cause disruption to customers, it must be scheduled with the following considerations:
+* at least 45 days notice should be provided to customers
+* not during the working week
+* maintenance should not overlap with [scheduled release](../releases/process.md) week
```
ZJvandeWeg, line 29:
```suggestion
* maintenance should not overlap with [scheduled release](../releases/process.md) week

```

src/handbook/development/ops/maintenance.md
```diff
@@ -0,0 +1,81 @@
...
+* networking components
+
+ 
+# Maintenance process
+
+## Planning
+
+The maintenance process is initiated by the Operations Team.
+
+A new [Infrastructure task](https://github.com/FlowFuse/CloudProject/issues/new?template=infra-task.yml) is created to co-ordinate all activities regarding the task - including the technical tasks and communications with customers.
```
ZJvandeWeg, line 22: @ppawlowski @knolleary Can the linked template require the updating of customers to be coordinated?

ZJvandeWeg:
@allthedoll Can you asses if we should polish this PR and merge it?

allthedoll:
I've pushed one addition: a short "When does this process apply?" section at the top that acts as a decision gate. The point being: when we can act within SLA without customer-visible disruption, we shouldn't generate comms for behind-the-scenes work. That scopes the comms-heavy process to cases where it's genuinely needed (e.g. worker node migration) rather than defaulting to noisy customer emails for everything. I think this addresses the open thread on line 38.

On ZJ's question about the issue template (line 22): yes, the Infrastructure task template should make customer coordination explicit. We could open a follow-up against the CloudProject template (@ppawlowski can you do that?) - not a blocker for this one.

LFG and get this merged. 🔥

PR #3011, opened 2025-03-05
HM25 event page

Yndira-E:
## Description

This PR introduces the Hannover Messe 2025 event page.

## Related Issue(s)

https://github.com/FlowFuse/customer/issues/323

## Checklist


src/events/hannover-messe-2025.njk
```diff
@@ -0,0 +1,83 @@
+---
+layout: layouts/base.njk
```
ZJvandeWeg, line 2: Why not create `src/_data/events/hannover-messe-2025.json` and make a page based on the data in there? Or is that the second iteration?

Yndira-E: The page isn’t finalised yet; we still need to add some information, and the layout will change. But your suggestion makes sense, and I’ll implement it in a future iteration.

Issue #2796, opened 2024-11-21
Suggestions to Better Align Home Page Messaging with Manufacturing Personas


Yndira-E:
### Description

After conducting [benchmarking research](https://www.figma.com/board/XpQ37KXmcLOfAeUVekuZkV/FF---Home-page-revision---Nov-2024?node-id=1-203&t=9Q95NTTiZNFBXIo4-1) and considering our current ICP, I identified areas where our home page messaging could be improved to resonate more effectively with manufacturing personas.

Our current messaging does not directly address the language of manufacturing, and I believe the following elements are missing from our home page:

1. **Business value in the hero section**  
2. **Manufacturing-specific keywords** like PLC, SCADA, MES, ERP  
3. **Graphic showing FlowFuse’s role in a manufacturing environment**  
4. **Logos of recognised protocols/integrations** that appeal to our personas  
   - Previously included but moved to the Features page at Hasmin's request  

Additionally (not part of the benchmarking comparison): 
- **Short introductions to our solutions pages** with links to each  
- **Testimonials from ICPs** or replace with **manufacturing customer stories**  
- **Manufacturing pain points addressed by FlowFuse**  

Authority-building content:
- A dedicated **“Why FlowFuse”** section or curated content like UNS or white papers  
- Highlight **FlowFuse leadership**, specifically being led by one of Node-RED's creators (currently mentioned on the Features page)  

These changes aim to better align our website with manufacturing personas and clearly communicate our value in this space.

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

Yndira-E:
Hi @Hasmin-AC,

I’ve already taken care of the graphic updates. I’ve added you as assignee since most of the suggestions involve copy updates. As mentioned in the description, not all of these suggestions need to be applied. If you have any better ideas, feel free to add them to the list as well so we can track all the changes we make.

gstout52:
I created PR #2802 to change the tagline and subtitle in the spirit of this issue.

gstout52:
Pending @Yndira-E 's investigation in HubSpot, I will create PRs to make the main webpage CTA a free trial, and change the highlighted action in the main heading 'Free Trial'.

Yndira-E:
Now, thinking about the structure of the page and following this suggestion:  
![image](https://github.com/user-attachments/assets/4ec61466-f6c1-4901-882c-468599366742)

@joepavitt and I have gathered the content we already have, along with some new suggestions.

[...]

PR #3014, opened 2025-03-06
Homepage: Add blueprint section

Yndira-E:
## Description

This is part of the changes we'll be introducing to the homepage as described [here](https://github.com/FlowFuse/website/issues/2796#issuecomment-2684667303).

## Related Issue(s)

https://github.com/FlowFuse/website/issues/2796

## Checklist


src/index.njk
```diff
@@ -131,7 +131,62 @@ hubspot:
             </div>
         </div>
     </div>
-
+    
+    <!-- Blueprints -->
+    {% set blueprints = [
```
ZJvandeWeg, line 136: I would make this array a data collection instead, not to pollute markup with data. What do yo think?

src/blog/2023/07/influxdb-historical-data.md
```diff
@@ -4,7 +4,7 @@ subtitle: Detailed instructions on how to create a Node-RED dashboard that shows
 description: Discover how to build a Historical Data Dashboard with InfluxDB and Node-RED. Capture, store, and visualize data for insightful analysis.
 date: 2023-07-18
 authors: ["andrew-lynch"]
-image: blog/2023/07/images/historical-data-dashboard.png
+image: blog/2023/07/images/historical-data_dashboard.png
```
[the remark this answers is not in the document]
Yndira-E: Because the image in the blueprint repo has the same name, so this one was overwriting the blueprint one.

src/index.njk
Yndira-E: done!

src/_data/homepage_blueprints.yaml
```diff
@@ -0,0 +1,15 @@
+- title: "CRUD MongoDB API"
```
ZJvandeWeg, line 1: I would assume that this data file only need 3 keys, which are pointers to blueprints in the repo we already import. Why have you chosen to duplicate these values?

Issue #3037, opened 2025-03-12
Update pricing page for IIOT Scale use case


MichaelBDavis:
### Description

Our pricing page isn’t conveying the flexibility we offer for IIOT Scale customers, which is causing friction.

- [ ] Add a Prominent Volume Discount Banner - "Scaling to hundreds or thousands of devices? Contact us for customized volume discounts!"
- [x] Add a FAQ entry
Q: Do you offer discounts for large-scale deployments?
A: Yes, we provide customized volume discounts for use cases requiring hundreds or thousands of Node-RED instances. Please [contact us] to discuss your specific requirements.​

### Have you provided an initial effort estimate for this issue?

I have provided an initial effort estimate

ZJvandeWeg:
@MichaelBDavis Wait, you mean, the sliders can come back to the pricing page? 😸

ZJvandeWeg:
@MichaelBDavis I did intent it as a tongue in cheek comment, but still believe volume brackets can be communicated with customers on the pricing page. As most customers are adopting without a lot of different parameters that apply to their pricing currently, being open might not be bad.

Yndira-E:
Since the price is only visible for the Team tier, how about adding a highlighted note to that card only? For Enterprise, the CTA is already 'Contact us'.

<img width="1441" alt="Image" src="https://github.com/user-attachments/assets/5ce43b74-fd81-4de4-b1b4-0450f2fb1cbe" />

gstout52:
After conversation with @MichaelBDavis , we'll go with adding a new line in the Feature Comparison table, in the Scale section, that says "Volume Pricing", with checkmarks under Team and Enterprise. When the visitor hovers over "Volume Pricing", the following language appears:

"Scaling to hundreds or thousands of Node-RED instances? Contact us for customized volume discounts!"

PR #3057, opened 2025-03-24
Pricing page: Add volume pricing to features table

Yndira-E:
## Description

#3037

The info dialog is not available in the preview environment but will be available in production. It will look like this:

Screenshot 2025-03-24 at 10 02 38

## Checklist


ZJvandeWeg:
@Yndira-E @MichaelBDavis Shouldn't this also be visible on the top of the page per tier?

Yndira-E:
From what I saw in the discussion, the decision was to include volume pricing in the Feature Comparison table rather than a banner on the card (see [this comment](https://github.com/FlowFuse/website/issues/3037#issuecomment-2744000750)). Let me know if we should revisit this! [I had previously suggested](https://github.com/FlowFuse/website/issues/3037#issuecomment-2743836858) adding it only to the Team card since it's the only one with visible pricing.

PR #3100, opened 2025-04-01
Call out Metrics review as part of strategy meeting + list metrics we're reviewing

emilieschario:
## Description

* Introduces Metrics review at recurring strategy meeting
* Lists out metrics we'll be reviewing
* This PR does not introduce details on the strategy meeting

## Checklist


ZJvandeWeg:
I don't think we need an additional approver, merging for the sake of speed.

PR #3150, opened 2025-04-15
Update achieving-success.md

gstout52:
## Description

Added section on calculating MRR for self-service

## Related Issue(s)

n/a

## Checklist

 - [] Suitable unit/system level tests have been added and they pass

### DIFF
--- a/src/handbook/company/achieving-success.md
+++ b/src/handbook/company/achieving-success.md
@@ -44,7 +44,7 @@ Target growth for this metric is 10% week over week.
 
 Customers on self-service plans, which presently includes only FlowFuse Cloud (FFC) Starter and Team tier, pay on a monthly basis. For these customers, we track Monthly Recurring Revenue (MRR). 
 
-The MRR of FFC Starter + Team is equal to the sum of all monthly subscriptions for those two tiers, including add-ons for additional services like remote instances and larger instances. As fees are paid on a monthly basis, such revenue is considered to be recurring.
+The MRR of both the Starter and Team tier is equal to the sum of all monthly subscriptions for those two tiers, including metered add-ons for additional services like remote instances and larger instances. As fees are paid on a monthly basis, such revenue is considered to be recurring.
 
 At any time, the current MRR of FFC Starter and Team can be calculated by looking up the corresponding products in Stripe, and adding the sum of each individual product's MRR. 
 
--- a/src/handbook/company/achieving-success.md
+++ b/src/handbook/company/achieving-success.md
@@ -40,6 +40,26 @@ it's usually harder to scale.
 
 Target growth for this metric is 10% week over week.
 
+### Monthly Recurring Revenue (MRR) for Self-Service
+
+Customers on self-service plans, which presently includes only FlowFuse Cloud Starter and Team tier, pay on a monthly basis. For these customers, we track Monthly Recurring Revenue (MRR). 
+
+The MRR of FFC Starter + Team is equal to the sum of all monthly subscriptions for those two tiers, including add-ons for additional services like remote instances and larger instances. As fees are paid on a monthly basis, such revenue is considered to be recurring.
+
+At any time, the current MRR of FFC Starter and Team can be calculated by looking up the corresponding products in Stripe, and adding the sum of each individual product's MRR. 
+
+Those products are:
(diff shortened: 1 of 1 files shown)


src/handbook/company/achieving-success.md
```diff
@@ -40,6 +40,26 @@ it's usually harder to scale.
 
 Target growth for this metric is 10% week over week.
 
+### Monthly Recurring Revenue (MRR) for Self-Service
+
+Customers on self-service plans, which presently includes only FlowFuse Cloud Starter and Team tier, pay on a monthly basis. For these customers, we track Monthly Recurring Revenue (MRR).
```
ZJvandeWeg, line 45:
```suggestion
Customers on self-service plans, which presently includes only FlowFuse Cloud (FFC) Starter and Team tier, pay on a monthly basis. For these customers, we track Monthly Recurring Revenue (MRR). 
```

Allows you to use `FFC` as term later.

src/handbook/company/achieving-success.md
```diff
@@ -40,6 +40,26 @@ it's usually harder to scale.
 
 Target growth for this metric is 10% week over week.
 
+### Monthly Recurring Revenue (MRR) for Self-Service
+
+Customers on self-service plans, which presently includes only FlowFuse Cloud Starter and Team tier, pay on a monthly basis. For these customers, we track Monthly Recurring Revenue (MRR). 
+
+The MRR of FFC Starter + Team is equal to the sum of all monthly subscriptions for those two tiers, including add-ons for additional services like remote instances and larger instances. As fees are paid on a monthly basis, such revenue is considered to be recurring.
```
ZJvandeWeg, line 47:
```suggestion
The MRR of both the Starter and Team tier is equal to the sum of all monthly subscriptions for those two tiers, including metered add-ons for additional services like remote instances and larger instances. As fees are paid on a monthly basis, such revenue is considered to be recurring.
```
The fact that add-ons are metered is a key detail here.

src/handbook/company/achieving-success.md
```diff
@@ -40,6 +40,26 @@ it's usually harder to scale.
...
+- FlowFuse Cloud - Starter
+- Starter - Small Instance Add-On
+- Starter - Remote Instance Add-On
+- FlowFuse Team Platform - Cloud
+- Team - Small Instance Add-On
+- Team - Medium Instance Add-On
+- Team - Large Instance Add-On
+- Team - Remote Instance Add-On
+
+For Starter, we calculate MRR by multiplying the total number of subscriptions by the cost of the subscription. This helps keep the focus on growing the number of Starter subscriptions rather than being distracted by noise introduced by sort-term discounts, like promotional coupons.
```
ZJvandeWeg, line 61: Now FFC Starter has add ons too, I'm not sure this is a good shortcut. The MRR per team should increase slightly Month over Month?

ZJvandeWeg (changes requested):

src/handbook/company/achieving-success.md
gstout52: It's not immediately clear to me why Starter add ons should affect this way of calculating. Can you say more?

ZJvandeWeg: When I can add a third instance at $20/mo, we cannot stated we have X number of starter tier teams and multiply it by the package price of $20 and use that as MRR.

gstout52: Add-ons are definitely recurring. Discounts are ignored in subscribers*12, as they tend to be short term. The purer revenue is the value per product that Stripe presents. However, this also takes the focus off of the number of subscriptions.

We can go either way here. We had been doing (subscriptions*12 + add-ons*12), and that's what I wrote up here. I think that's the more meaningful number, but I will revise the page if you'd rather go with (actual product current value)*12 + (add-ons*12).

src/handbook/company/achieving-success.md
```diff
@@ -40,6 +40,26 @@ it's usually harder to scale.
...
```
gstout52, line 61:
```suggestion
For Starter, we calculate MRR by multiplying the total number of subscriptions by the cost of the subscription. This helps keep the focus on growing the number of Starter subscriptions rather than being distracted by noise introduced by short-term discounts, like promotional coupons.
```

PR #3290, opened 2025-05-30
Whitepaper landing pages update

Yndira-E:
## Description

To increase conversions for the OSS and Innovation in Manufacturing whitepapers, we're now displaying only an introduction to the PDF content on the page instead of the full text.

Additional changes, such as new thank-you pages and improvements to the template to move all data into the front matter, will be handled in separate PRs.

## Checklist


src/whitepaper/accelerating-innovation-in-manufacturing-with-flowfuse.njk
```diff
@@ -0,0 +1,21 @@
+---
+layout: layouts/base.njk
```
ZJvandeWeg, line 2: The extends is nice on line 15, but why not use this line instead?

src/whitepaper/whitepaper.json
```diff
@@ -1,5 +1,4 @@
 {
-    "layout": "layouts/whitepaper.njk",
```
ZJvandeWeg, line 2: Isn't this a better solution?

ZJvandeWeg (changes requested):

src/whitepaper/accelerating-innovation-in-manufacturing-with-flowfuse.njk
ZJvandeWeg: I think if the layout is `whitepaper`, all the `blocks` below can be in the frontmatter?

src/whitepaper/whitepaper.json
Yndira-E: You're right, this is a better solution. I moved the current template to the JSON to avoid repetition in this PR, but it's just temporary — it will change once I update the template.

ZJvandeWeg (approved):

PR #3298, opened 2025-06-02
Changelog: Instance Performance Tab

joepavitt:
## Description

Adds changelog entry for https://github.com/FlowFuse/flowfuse/pull/5528


src/changelog/2025/06/instance-performance-view.md
```diff
@@ -0,0 +1,17 @@
...
+tags:
+  - changelog
+---
+
+We've added a new view for every Hosted Instance in FlowFuse, Performance.
+
+![Screenshot of the Performance view for a Hosted Instance](./images/instance-performance.png){data-zoomable}
+_Screenshot of the Performance view for a Hosted Instance_
+
+This view provides a comprehensive insight into the performance of your hosted Node-RED instances, focussing on CPU usage over time, better enabling you to detect performance issues and memory leaks within your Node-RED Flows.
```
knolleary, line 15:
```suggestion
This view provides a comprehensive insight into the performance of your hosted Node-RED instances, focussing on CPU usage over time, better enabling you to detect performance issues within your Node-RED Flows.
```

Maybe not mention memory leaks as its CPU only at the moment?

knolleary (changes requested):
Suggest we shouldn't talk about memory leaks as its CPU only information at the moment.

knolleary (approved):

PR #3490, opened 2025-07-14
Rebrand Device Management to Edge Connectivity

gstout52:
## Summary
• Rebrand "Device Management" to "Edge Connectivity" in the solutions page
• Update page title, headings, and content to reflect the new branding
• Maintain technical accuracy while emphasizing connectivity focus

## Test plan
- [ ] Verify the solutions page displays correctly
- [ ] Check that all updated copy reads naturally
- [ ] Ensure no broken links or references

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/solutions/device-management.njk
+++ b/src/solutions/device-management.njk
@@ -1,11 +1,11 @@
 ---
 layout: solution
-title: FlowFuse for Edge Device Management
+title: FlowFuse for Edge Connectivity
 image: /images/solutions/og-solutions-device-management.jpg
 subtitle:
 heroImg: ./images/pictograms/edge_blue.png
 description:
-    <p>FlowFuse simplifies the management of deployments to edge devices in manufacturing, from equipment to PLCs. Our platform ensures secure and reliable handling of large-scale deployments, overcoming the challenges of extensive device management.</p>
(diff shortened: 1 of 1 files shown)


gstout52:
> @gstout52 Why keep the URL the same and not update that too? And just add a redirect?

Done

Yndira-E:
Sure! When we change the name of a file, it also changes the URL. That means any existing links, from social posts or other pages, would lead to a 404 unless we set up a redirect.

To prevent that, we add an entry in our [redirects file](https://github.com/FlowFuse/website/blob/main/src/redirects.njk), which handles those cases.

In my case, I also do a search for the old URL across the repo and replace it with the new one, it’s better to keep our internal links pointing to the current URLs instead of relying on redirects.

src/solutions/edge-connectivity.njk
```diff
@@ -50,7 +50,7 @@ description:
                 </div>
             </div>
             <div class="text-left self-start">
-                <h3 class="flex align-center">Remote <span class="text-indigo-600">Troubleshooting</span></h3>
+                <h3>Remote <span class="text-indigo-600">Troubleshooting</span></h3>
```
ZJvandeWeg, line 53: @gstout52 Why change the styling here?

gstout52: In the preview "RemoteTroubleshooting" appears incorrectly as one word instead of two. I'm attempting to fix that, but it's a styling error that existed before this PR. Claude thinks it's fixed now, but if the updated file hasn't fixed it, I'll try again.

PR #3491, opened 2025-07-15
Add FAQ schema support to blog posts

sumitshinde-84:
## Description

This PR Updates the blog post layout to include an FAQ component if meta.faq is present in the front matter. This component should render a list of frequently asked questions and answers at the end of the blog post.

As an example implementation, added the Serial Port blog post to include an FAQ section by adding a faq field in its front matter.

## Checklist


src/_includes/layouts/post.njk
```diff
@@ -59,6 +59,9 @@
                     </div>
                     {{ content | safe }}
                 </div>
+                {% if meta.faq %}
```
Yndira-E, line 62: I think the FAQ section could use a title, and it feels a bit odd to have it come after the "Conclusion." Maybe just adjusting the titles could help it flow better.

How about:
- Replacing "Conclusion" with "Key Takeaways"
- And giving the FAQs section a proper heading. Either something general like "Frequently Asked Questions", or more specific to the article, like "Common Questions About Connecting Serial Devices"

Yndira-E:
It looks like the FAQs section is part of the conclusion...

Screenshot 2025-07-15 at 18 24 16

Was that the intention?

Also, what about the suggestion of changing the "Conclusion" section title?
It feels a bit off to wrap up the article and then have more content right after.

sumitshinde-84:
> content

Ah, then I think the FAQ title should be an H2 instead of the current H3. I’ll update that, along with changing both 'Conclusion' sections to 'Takeaways'

sumitshinde-84:
I’ve just updated it

src/blog/2025/07/connect-legacy-equipment-serial-flowfuse.md
```diff
@@ -224,7 +242,7 @@ Each time a message is received, the node also outputs the current port configur
 
 In short, the **serial control** node adds a layer of resilience and flexibility that is often essential in real-world deployments—where devices come and go, ports are never quite consistent, and downtime is not an option.
 
-## Conclusion
+## Takeaways
```
Yndira-E, line 245:
```suggestion
## Key Takeaways
```

src/_includes/layouts/post.njk
```diff
@@ -59,6 +59,12 @@
...
+                    <div class="prose mt-12">
+                        <h2 class="mb-4">Frequently Asked Questions</h2>
```
Yndira-E, line 64:
```suggestion
                        <h2 class="mb-1">Frequently Asked Questions</h2>
```

Yndira-E (approved):

PR #3600, opened 2025-08-08
/pricing: Add additional instance cost

Yndira-E:
## Description

This PR adds pricing details for additional instances to the Starter and Pro cards, with the latter covering both cloud and self-hosted.

## Checklist

### DIFF
--- a/src/pricing/index.njk
+++ b/src/pricing/index.njk
@@ -80,7 +80,7 @@ hubspot:
                                     </ul>
                                     <span class="hidden md:inline"><br></span>
                                     <br>
-                                    <note class="font-semibold">Includes 1 hosted instance.
+                                    <note class="font-semibold">Includes 1 hosted instance. Additional instances are $20/month.
                                     </note>
                                 </div>
                             </div>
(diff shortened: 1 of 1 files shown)


Yndira-E:
@MichaelBDavis I'm moving your comment to the PR

> @Yndira-E @gstout52 Wondering if we can put this information in the Feature Comparison section to keep the pricing boxes less cluttered. Caveat: I don't have any designer dna.
> 
> Also, we should show the cost for Hosted instance and Remote instance.

Is there a difference between hosted and remote instance pricing?

If the information in this PR is accurate, I’d move forward with this version and we can iterate next week when Greg is back. I’m not entirely sure about all the items in the table that refer to this, or about the difference in pricing.

knolleary:
@MichaelBDavis @Yndira-E the current pricing is the *same* for the 'default' hosted instance type and the remote instance. $20 in the case of starter, $35 in the case of team/pro. So the information in this PR is technically accurate to the current plans.

As @Yndira-E suggests, happy for us to move on this PR and let you iterate as needed with @gstout52 next week.

knolleary (approved):

MichaelBDavis:
@knolleary Just to be doubly sure. @gstout52 provided this to sales.

Starter: $20 for both
Team: $35 for remote, $65 for hosted

knolleary:
@MichaelBDavis okay - I can see the mix up.

Starter only has access to Small instances: $20/each. Same as Remote. So we can be definitive on that.

Pro has access to Small ($35) / Medium ($65) / Large ($250).  It gets 5 Small included in the price. This is why the wording say for pro say 'additional instances *from* $35'.

PR #3607, opened 2025-08-12
Blog: Building Historical Data Dashboard with FlowFuse Tables

sumitshinde-84:
## Checklist


src/blog/2025/08/time-series-dashboard-flowfuse-postgresql.md
```diff
@@ -0,0 +1,231 @@
...
+    );
+
+    CREATE INDEX idx_timestamp ON sensor_readings(timestamp DESC);
+    CREATE INDEX idx_sensor_timestamp ON sensor_readings(sensor_id, timestamp DESC);
+    ```
+
+**Schema Breakdown:**
+
+  * `TIMESTAMPTZ`: We use this data type to store timestamps with timezone information. This is critical for applications with sensors spread across different geographical locations, ensuring data is always consistent.
+  * **Indexes**: The indexes are vital for query speed. The `DESC` (descending) keyword in our `idx_timestamp` is a specific optimization. It pre-sorts the data from newest to oldest. When you runs a common query like `SELECT * FROM sensor_readings ORDER BY timestamp DESC LIMIT 10` to fetch the latest readings, PostgreSQL can use this index to retrieve the data instantly without performing a costly sort operation.
```
PabloFilomeno83, line 64:
```suggestion
  * **Indexes**: The indexes are vital for query speed. The `DESC` (descending) keyword in our `idx_timestamp` is a specific optimization. It pre-sorts the data from newest to oldest. When you run a common query like `SELECT * FROM sensor_readings ORDER BY timestamp DESC LIMIT 10` to fetch the latest readings, PostgreSQL can use this index to retrieve the data instantly without performing a costly sort operation.
```

src/blog/2025/08/time-series-dashboard-flowfuse-postgresql.md
```diff
@@ -0,0 +1,231 @@
...
+
+You have successfully built a historical data dashboard using FlowFuse Tables and Node-RED. By implementing efficient batch inserts and optimized query patterns, you have created a solution that is both powerful and scalable for demanding Industrial IoT environments.
+
+With FlowFuse Tables now part of the platform, you can build complete industrial applications without juggling external databases or leaving the FlowFuse environment. FlowFuse is now a comprehensive data platform with the ability to collect, connect, transform, store, and visualize data. Combined with FlowFuse's enterprise features—team collaboration, version control, device management, and secure deployments—you have everything needed to take your IIoT projects from prototype to production within one integrated platform.
+
+This means less complexity and faster time to value for your industrial data initiatives. Your historical dashboards, real-time monitoring, and OEE dashboards can all live in the same ecosystem, managed by the same team, with consistent security and governance controls.
+
+[Start using FlowFuse today](https://app.flowfuse.com/account/create) to make your IIoT data work for you.
```
PabloFilomeno83, line 231:
```suggestion
Ready to build your own time-series dashboard? [Get started with FlowFuse Tables] or [explore more industrial blueprints].```

**Review by PabloFilomeno83 — CHANGES_REQUESTED** [review]
The tutorial doesn't mention what happens if the database connection fails or if there's no data for the selected time range. I think it would be worthy mentioning it? wdyt?
I've reviewed with a couple suggestions and typos.

**Comment by sumitshinde-84** [comment]
@andreguerra7 Could you please review this blog post?

File: src/blog/2025/08/time-series-dashboard-flowfuse-postgresql.md [hunk 3]
```diff
@@ -0,0 +1,266 @@
...
+```
+
+3.  Add another **Change node** to add a precise timestamp to each reading.
+
+      - Set `msg.payload.timestamp`
+      - To the value type **timestamp**.
+
+4. Add a **Function node** and name it **"Batch Accumulator"**. Paste the following JavaScript code into the node — it already includes inline comments explaining each step. This function will accumulate incoming readings in batches until the specified batch size is reached, and then create the SQL query to perform batch inserts into the database.
+
+```javascript
```
andreguerra7, line 106: The "Batch Accumulator" function's code in the flow at the end of the article, does not match the correct version shown here.

The flow's code incorrectly generates placeholders for 7 values per reading `(i*7+1)`, but the `sensor_readings` table is only created with 4 columns, which will possibly cause the batch insert to fail

src/blog/2025/08/time-series-dashboard-flowfuse-postgresql.md
```diff
@@ -0,0 +1,266 @@
...
...,"disabled":"false"},{"id":"6d8bff5f3fded5c2","type":"ui-theme","name":"FF Theme","colors":{"surface":"#5046e5","primary":"#5046e5","bgPage":"#ffffff","groupBg":"#ffffff","groupOutline":"#d4d1ff"},"sizes":{"density":"default","pagePadding":"15px","groupGap":"15px","groupBorderRadius":"4px","widgetGap":"12px"}},{"id":"a1e6e28bf926d3f8","type":"group","z":"a9e721cc8fbe46fe","style":{"stroke":"#b2b3bd","stroke-opacity":"1","fill":"#f2f3fb","fill-opacity":"0.5","label":true,"label-position":"nw","color":"#32333b"},"nodes":["da0e835628f95a0f","f7dde47e73470bfa"],"x":774,"y":219,"w":392,"h":82},{"id":"da0e835628f95a0f","type":"catch","z":"a9e721cc8fbe46fe","g":"a1e6e28bf926d3f8","name":"","scope":null,"uncaught":false,"x":860,"y":260,"wires":[["f7dde47e73470bfa"]]},{"id":"f7dde47e73470bfa","type":"debug","z":"a9e721cc8fbe46fe","g":"a1e6e28bf926d3f8","name":"debug 6","active":true,"tosidebar":true,"console":false,"tostatus":false,"complete":"false","statusVal":"","statusType":"auto","x":1060,"y":260,"wires":[]},{"id":"01fd548fa46a547c","type":"global-config","env":[],"modules":{"@flowfuse/nr-tables-nodes":"0.1.0","@flowfuse/node-red-dashboard":"1.26.0","node-red-contrib-moment":"5.0.0"}}]
```
andreguerra7, line 255: There is a data mismatch in the flow. The "Simulate Sensor" node generates humidity, pressure, and quality_score data, but the CREATE TABLE query doesn't create columns to store them from what I can see, so this data is discarded.

Also, I can't fully see the rest of the code as mentioned, but please double check , maybe you are setting the same timestamp for all records, by using `New Date()`

andreguerra7 (changes requested):
@sumitshinde-84 
I've added a few points..

I can't fully review the functionality and fact check as I don't have an instance with FF tables, if this is needed I'm happy to do it, but could you ping someone that can give me this access?

If it's not the case, these are the only changes I recommend

andreguerra7 (changes requested):

sumitshinde-84:
@andreguerra7 Thanks for the suggestions! I’ve applied all of them—could you please take a look and approve? FYI, it's scheduled for today.

PR #3625, opened 2025-08-20
Changelogs for Direct SSO and CORS

hardillb:
## Description

Do not merge until 2.21.0 release day as both will only be available as part of the release

Includes:

- Direct SSO
- HTTP CORS config


joepavitt:
Not merging until release as requested

PR #3636, opened 2025-08-21
handbook: Add JD for Dashboard engineer

ZJvandeWeg:
## Description

While this is not a role we have capacity to hire for now, it's one to consider in the future. Adding the JD to the handbook is one step.

Before this role can be added to the headcount planning, we would need more details in the handbook and a better understanding on the ROI for this role.

## Checklist

### DIFF
--- a/src/handbook/peopleops/job-descriptions/dashboard-engineer.md
+++ b/src/handbook/peopleops/job-descriptions/dashboard-engineer.md
@@ -1,0 +1,163 @@
+---
+navTitle: FlowFuse Dashboard Engineer
+navGroup: Job Descriptions
+---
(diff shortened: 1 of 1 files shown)


joepavitt (approved):

joepavitt:
> Is there an intent with that or just an old draft being worked off?

My interpretation of this was that it's a _founding_ role of a greenfield project, in the sense of complete freedom/project ownership. Not necessarily founding in terms of FF as a company/product, but I agree that could be scale back.

joepavitt:
"Greenfield", "Product Owner", "Independence" are phrases that spring to mind

PR #3730, opened 2025-09-03
Fix data-zoomable in docs template

Yndira-E:
## Description

This PR removes the redundant inclusion of `medium-zoom.min.njk`, which was already part of `common-js.njk`. The duplication was causing image rendering issues, and removing it resolves the problem.

## Checklist


Steve-Mcl (approved):
works perfectly.

PR #3822, opened 2025-09-22
Company Values: Add "Innovate" and "Knowledge is Power" subpoints in company values

joepavitt:
## Description

I was reviewing some old IBM docs on how performance reviews were done, and two key points I think we're explicitly missing in our own values against which we do reviews are:

- Innovate: Explicitly calling out the drive to innovate in everything we do encourages forward momentum and improvement
- Knowledge is Power: It's loosely pointed at in other points, but calling out explicitly the value in sharing your knowledge, encouraging internal (and external) knowledge sharing of best practices, how-to's and ensuring knowledge isn't silo'd is another important point that should be added.

### DIFF
--- a/src/handbook/company/values.md
+++ b/src/handbook/company/values.md
@@ -30,6 +30,9 @@ it down again. Get the minimal amount of work down, but no less, find the [golde
 1. Everything is a Draft: Today won’t be perfect, tomorrow will be better. Perfection is the death of progress.
 1. Decide where good is good enough; understand where excellence is our company's
 edge. Apply the [pareto principle](https://en.wikipedia.org/wiki/Pareto_principle).
+1. Innovate: Be willing to try new things, and be willing to fail. Learn from your mistakes,
+and be willing to try again. This covers innovation for new features and public-facing projects, but also
+innovation for how we do things internally, and always seeking to improve our own processes and procedures.
 
 ## 👥 Collaborative Community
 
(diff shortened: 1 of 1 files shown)


knolleary (approved):

src/handbook/company/values.md
```diff
@@ -30,6 +30,9 @@ it down again. Get the minimal amount of work down, but no less, find the [golde
 1. Everything is a Draft: Today won’t be perfect, tomorrow will be better. Perfection is the death of progress.
 1. Decide where good is good enough; understand where excellence is our company's
 edge. Apply the [pareto principle](https://en.wikipedia.org/wiki/Pareto_principle).
+1. Innovate: Be willing to try new things that could help the company, and be willing to fail. Learn from your mistakes,
+and be willing to try again. This covers innovation for new features and public-facing projects, but also
+innovation for how we do things internally, and always seeking to improve our own processes and procedures.
```
ZJvandeWeg, line 35: Same argument as earlier, we shouldn't adopt this as a sub-value. `Bias towards action` covers the willingness to trying (and thus be willing to fail). Innovation should never be a goal, the result should be, which is why it's the top value.


```suggestion
```

For this PR, `Knowledge is Power` is the chance we should merge, solely.

ZJvandeWeg:
@joepavitt How is `Knowledge is Power` different from `Write things down`? I think there's large overlap, and maybe it should be merged into one? I do think your take on `Knowledge is Power` is better worded that the original.

src/handbook/company/values.md
joepavitt: Bias for Action is me coding and getting it done. 

Innovation is me experimenting with Claude and trying new ways of working, seeing if things work, sticking with it if it does.

PR #3823, opened 2025-09-22
Team Broker Schema Capture changelog

hardillb:
## Description

Add Changelog for Team Broker Schema collection

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/5951


joepavitt (changes requested):

joepavitt:
It was very technical, and din't really communicate the _why_, so i've made amendments to cover that and focus on the value to the reader/customer.

joepavitt (approved):

PR #3835, opened 2025-09-24
Blog: Node-RED vs n8n: Choosing the Right Automation Platform for Your Enterprise

sumitshinde-84:
## Checklist


src/blog/2025/09/node-red-vs-n8n.md
```diff
@@ -0,0 +1,133 @@
+---
```
ZJvandeWeg, line 1: If we build a competitive page, the URL will be `flowfuse.com/vs/<other-solution>`

cc @Yndira-E @PabloFilomeno83

PR #3841, opened 2025-09-25
Update offer email.md

PabloFilomeno83:
Second offer email could be personalized (otherwise if feels very impersonal)

## Checklist


ZJvandeWeg:
@PabloFilomeno83 I disagree, keep it formal and standard. There's one way to do it right, I don't want to open up legal trouble because someone forgot to mention a key detail.

PR #3853, opened 2025-09-26
Node-RED Doc: Integrating GraphQL APIs in Node-RED

sumitshinde-84:
## Checklist


sumitshinde-84:
@andreguerra7 could you please review this PR?

andreguerra7 (changes requested):
@sumitshinde-84 This looks good, I added a few minor suggestions on the file. Below some blockers:

- Method 1 (Mustache Templates)
As written, the example always returns null. To make it work, I had to move the variable to the root of the message `msg.countryCode = "JP`";.

- Mutations section
The draft jumps from querying the Countries demo API straight into mutations with `DeviceInput`. That API is read-only, so readers will inevitably hit `message: "Unknown type "DeviceInput"."`

Either (a) provide another public API that supports mutations, or (b) explicitly state that the code is illustrative and won’t work against the demo endpoint.

PR #3896, opened 2025-10-02
home: Clearer end-to-end story

ZJvandeWeg:
## Description

I think this makes a good alternative to the PR merged today.

## Checklist

### DIFF
--- a/src/index.njk
+++ b/src/index.njk
@@ -117,31 +117,89 @@ hubspot:
         </div>
     </div>
 </div>
-<!-- Status Quo section -->
+<!-- Industrial Data Platform section -->
 <div class="about w-full pt-16 px-6">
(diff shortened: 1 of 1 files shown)


PabloFilomeno83:
I like this - but why deleting the block of Why Traditional Industrial Integration Can't Scale (And How We Use AI to Fix It)? I think we can use both.

Yndira-E:
I'd recommend we wait until all the PRs are in and merged, as more changes are coming in, and we'll see where we can integrate this part if it doesn't overlap with the ones already in place.

PabloFilomeno83:
Fair point @Yndira-E makes sense for me.

ZJvandeWeg:
@Yndira-E @PabloFilomeno83 It's not too great to see how little iteration since this PR has actually been done.

PR #4092, opened 2025-11-13
CTA Template

sumitshinde-84:
## Description

CTA template for blogs that can be used for most FlowFuse posts

## Checklist

### DIFF
--- a/src/_includes/components/ai-chat-interface.njk
+++ b/src/_includes/components/ai-chat-interface.njk
@@ -2,7 +2,7 @@
     <div class="ai-chat-box">
         <div class="textarea-wrapper relative mb-4 h-40 p-4 pb-12 rounded-lg overflow-hidden">
             <textarea
-                placeholder="Tell us what you are trying to build and the FlowFuse Expert Assistant will create a guide to get you up and running quickly"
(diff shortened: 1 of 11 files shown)


Yndira-E (changes requested):
Hi @sumitshinde-84,

I've created the `{% blueCard %}` shortcut for this kind of thing. It works directly in `.md` files, so you can tailor the message for each article.

It looks like this:
Screenshot 2025-11-17 at 12 10 44

If you want to turn it into a partial, you can use the `ff-blue-card` class to style it the same way. If you prefer that approach, I’d suggest adding a variable to the partial so you can customise the message for each article.

The issues with the version you're proposing are that the text styles don’t match the rest of the article, there’s an odd space at the top, and the copy would always be the same.

sumitshinde-84:
> Thanks Sumit. A question: why is everything centered?

I’m not sure which alignment would be best. Since it includes a heading, CTA text, and two buttons, centering seems like the best option to me,  but could you please provide some suggestions?

Yndira-E (approved):
I'm merging as is; we can iterate later

Yndira-E:
@PabloFilomeno83 I thought you were aware of Sumit working on this. Here's the PR to revert it: https://github.com/FlowFuse/website/pull/4115

PR #4104, opened 2025-11-16
Support chat api v4

Steve-Mcl:
## Description

Support additional responses including flows json

Additional, adds DomPurify

### **IMPORTANT** 
This is a work in progress (for hand over)

**TODO**
- [x] Write code for copy to clipboard (might also need to add exceptions for dom purify sanitization or re-arrange code to add click handlers after dom purification)
- [x] Make the "copy flow" buttons "look nice"
- [x] Update URL to non `-dev` prod URL

### prerequisites 
- [x] deploy v3/v4 flows on the production instance

## Checklist

### DIFF
--- a/src/js/ai-expert-modal.js
+++ b/src/js/ai-expert-modal.js
@@ -823,7 +823,7 @@ document.addEventListener('DOMContentLoaded', function() {
         // Setup Guide label for guide messages
         let headerText = 'Result'
 
-        if (richAnswer.kind === 'guide') headerText = 'Guide'
+        if (richAnswer.kind === 'guide') headerText = 'Setup Guide'
         if (richAnswer.kind === 'resources') headerText = 'Resources';
 
         html += `<div class="bg-indigo-100 text-indigo-700 text-sm px-3 py-2 rounded-full inline-block mb-3">${headerText}</div>`;


dimitrieh:
@Steve-Mcl regarding the flow buttons:

> 2 columns
This is a configuration currently for "resources" on the FlowFuse expert on the website. This is not yet configured as such in the app.  My suggestions here would be to not worry too much about making it 2 columns because of:
- Separation with "resource" buttons.
- We do not yet have the double column layout in FlowFuse expert in the app
- We might need some additionally horizontal space for the clipboard icon? (double column layout will make this harder, rather than easier)
- Might we want to show other information? For resource buttons we have:
  - favicon
  - title
  - url

CleanShot 2025-11-17 at 14 43 41@2x

dimitrieh:
Originally we had something like this, which also showed the json:

CleanShot 2025-11-17 at 14 48 24@2x

It included some instructions regarding how to use the flows. If we are going to put something in the clipboard I suggest we have:
- Indication of clicking will do an action rather than opening up page (clipboard button most likely in this case)
- Instructions? for what to do after
- Validation that the user has something copied into their clipboard after clicking the button? (for copying to clipboard, inline verification is often best, rather than a toast or something else). e.a.:

CleanShot 2025-11-17 at 14 51 12

dimitrieh (approved):
Approving during live review

PR #4356, opened 2026-01-08
Generate Integration Pages

sumitshinde-84:
## Description

This PR generates individual integration pages for each integration listed under `/integrations`. The integration page data is fetched from the npm registry, and it currently generates pages for all **FlowFuse-certified nodes** as well as the **top 50 most downloaded nodes**.

Both certified and non-certified node pages display the node’s README along with key details such as **author**, **last updated date**, **published date**, **npm link**, **maintainers**, and **GitHub stats**.

* **Certified node pages** display a FlowFuse Certified mark at the top, along with a banner highlighting the node’s certification status and a link to learn more about certified nodes.
* **Non-certified node pages** also include a banner, but with messaging that encourages node authors to apply for certification, including a link to the certification application form.

https://github.com/user-attachments/assets/d582af30-666e-4f4a-8397-8a5a5f7e6898

### SEO & Traffic Rationale

The primary goal of generating these pages is to **increase organic traffic** to the FlowFuse website. Today, when users search for documentation on specific Node-RED nodes, they typically land on the Node-RED node README pages. According to Node-RED search analytics, these pages receive a **significant number of clicks daily**.

There was an initial concern around **duplicate content** and the risk of Google penalizing these integration pages. However, this should not be an issue because:

* We add certification badges and banners, which provide trust and authority that Google values.
* We include **additional links and contextual information**

These elements meaningfully differentiate our pages from the original Node-RED documentation.

### Expected Impact

With this strategy, when users search for any popular node, our integration pages are likely to rank **highly in search results**, potentially in the **second position**.

Based on existing data:

* Highly downloaded nodes receive ~**100 clicks/day** on the Node-RED website
* Capturing just **10–15 clicks per node** would result in approximately:

**~500 additional clicks per day**
(50 nodes × 10 clicks per node)

### Note on Implementation & Design

> **Note:** Claude was used during parts of this PR for content and implementation assistance. The current design is intentionally basic and **not final**. The structure is in place, and we can **iterate and improve the UI/UX** in follow-up PRs based on feedback and performance data.

## Checklist

### DIFF
--- a/src/handbook/peopleops/hiring/star-questions.md
+++ b/src/handbook/peopleops/hiring/star-questions.md
@@ -11,7 +11,7 @@ Below are some questions that can be used and how answers are evaluated.
 
 ## STAR Interview Questions
 
-#### 1. How do you build rapport with teammates if you don't work in the same physical location?
+#### 1. How do you build connection and strengthen relationships on a distributed team, beyond just getting the work done?
 
 **Values**
 - Collaborative Community


src/_data/integrations.js
```diff
@@ -2,46 +2,329 @@ const EleventyFetch = require("@11ty/eleventy-fetch");
...
+                            }
+                        } catch (err) {
+                            // Examples folder doesn't exist or API error - this is fine, just skip
+                            node.examples = [];
+                        }
+                    }
+                }
+
+                if (nodeDetails.readme) {
+                    // List of valid Prism.js languages (common ones)
```
ZJvandeWeg: You'd expect Prism.JS to expose this in their package, and we don't need to manually maintain this list?

sumitshinde-84:
update: new examples tab is added

PR #4489, opened 2026-02-01
search: Migrate to PageFind

ZJvandeWeg:
I found the results in PageFind better. Now I'm slowly trying to migrate to astro I found PageFind, and it works quickly and locally on a branch.

## Checklist

 - [ ] Post merge -- remove Algolia index and account


Yndira-E:
We're not moving to Astro anymore and Algolia is currenty working correctly. I'm closing it for now, we can review alternatives for Nuxt in case Algolia isn't a good fit in that setup.

PR #4893, opened 2026-04-16
Update to `Qualifying questions`

MichaelBDavis:
Remove the question "Is this an active or planned initiative (not just general research)?". 

The prospect will most often not be able to answer this question at this stage (especially Z2s). This will be a qualification question in the next Discovery meetings.

### DIFF
--- a/src/handbook/sales/hubspot.md
+++ b/src/handbook/sales/hubspot.md
@@ -42,7 +42,6 @@ MQLs will be engaged by a Sales Representative, according with the appropriate g
 - Does the prospect and company fit the target industry / company size / role?
 - Are they working on an initiative related to connecting systems, industrial data, or operational workflows?
 - Can they describe a specific problem, use case or goal they’re trying to address?
-- Is this an active or planned initiative (not just general research)?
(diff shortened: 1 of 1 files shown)


src/handbook/sales/hubspot.md
```diff
@@ -42,7 +42,6 @@ MQLs will be engaged by a Sales Representative, according with the appropriate g
 - Does the prospect and company fit the target industry / company size / role?
 - Are they working on an initiative related to connecting systems, industrial data, or operational workflows?
 - Can they describe a specific problem, use case or goal they’re trying to address?
-- Is this an active or planned initiative (not just general research)?
```
ZJvandeWeg, line 45: @MichaelBDavis Shouldn't we identify "Why now"? And if not through this question, how?

MichaelBDavis: @ZJvandeWeg Yes, we will definitely want to ask this question in Discovery but I wouldn't want to disqualify an opportunity at the very initial stage (think SDR) because they don't yet fully understand what we do.

PR #4949, opened 2026-05-05
handbook: Asana tracks deal desk work

ZJvandeWeg:


UnicornGunnerz (commented):
I verified the links and can confirm that the Asana project is not visible to non-members. 

Would be great to see if we can also get the confirmation from someone who currently doens't have access to Asana

PR #4979, opened 2026-05-08
Restore image optimization cache

Yndira-E:
## Summary

- Preserve `nuxt/public/img` during Nuxt build cleanup so restored Netlify cache is not deleted before Eleventy runs.
- Re-enable Eleventy image optimization in the Nuxt production build by removing `SKIP_IMAGES=true`.

## Notes

The first build after this change is expected to be slow because it needs to regenerate optimized images. After Netlify caches `nuxt/public/img`, subsequent builds should reuse the generated image outputs instead of recompressing everything.

For the first deploy, use **Clear cache and deploy site** in Netlify to avoid carrying over unoptimized images from previous `SKIP_IMAGES` builds.

## Checklist


Yndira-E:
@knolleary I can deal with merging and the deploy details in Netlify after you review/approve. No need to worry about that.

knolleary (approved):

knolleary:
@Yndira-E all yours :)

PR #4993, opened 2026-05-12
Clarify customer data usage guidelines for AI

knolleary:
Revised conditions for customer data usage with AI systems, including clarifications on permitted and prohibited uses.

There was duplication in the 'do not use' list - not training models was state more than once in different ways. The language style didn't align between list header and list contents leading to ambiguity. This, hopefully, makes it clear and focussed on the model training aspect.

Removes a contradiction around use of anonymised data for general product improvements.


dimitrieh:
closing this for now as per the slack discussion

PR #4994, opened 2026-05-12
Revert "Update travel booking guidelines with new standards"

abby-flowp:
Reverts FlowFuse/website#4989


ZJvandeWeg:
@abby-flowp why revert? I agree the text is quite sharp and might need some polish, but a revert seems too much into the other direction?

abby-pie:
> @abby-flowp why revert? I agree the text is quite sharp and might need some polish, but a revert seems too much into the other direction?

@ZJvandeWeg  My bad, the revert was unintentional, I misclicked.

PR #5064, opened 2026-05-26
nuxt: Implement cookie banner and analytics tracking

Yndira-E:
## Description

This PR enables Nuxt-rendered pages to share the same cookie banner and analytics stack (GTM, PostHog, HubSpot) as Eleventy pages.

- Extracts all analytics scripts (GTM, Google Consent Mode, PostHog, HubSpot tracking and chat, cookie banner) from Eleventy's `base.njk` into shared HTML partials at `src/_includes/analytics/`
- A Nitro server plugin injects the same partials into Nuxt-rendered pages, giving a single source of truth for both frameworks during the migration period
- Fixes LinkedIn ad cookies (`_li_dcdm_c`, `_li_ss`) persisting after consent withdrawal on production — `autoClear` only cleared from the current host, not from the root `.flowfuse.com` domain
- Removes redundant PostHog cookie-clearing code that was already handled by `autoClear` in `cookieconsent-config.js`
- Adds `dev:nuxt` script that loads the root `.env` consistently with other dev scripts

When the migration to Nuxt is complete, the shared HTML partials and Nitro plugin should be replaced with a native Vue plugin using vanilla-cookieconsent's official Vue integration. 

The current approach was chosen to avoid duplicating the current cookie consent configuration (categories, callbacks, GTM/PostHog/HubSpot integrations) across both frameworks during the migration period. One source of truth, zero risk of the two implementations drifting apart.

## Checklist


Yndira-E:
Hey @ZJvandeWeg, before we merge #5065, do you think we could get this PR reviewed and merged?

It brings over the cookie banner and analytics stack (HubSpot, PostHog and GTM) from the Eleventy site, so all Nuxt pages share the same tracking and consent setup.

package.json
```diff
@@ -8,7 +8,8 @@
     "author": "Nick O'Leary",
     "workspaces": ["nuxt"],
     "scripts": {
-        "dev": "concurrently \"npm run dev:eleventy\" \"npm run dev:postcss\" \"npm run dev:postcss-nuxt\" \"npm run dev --workspace=nuxt\"",
+        "dev": "concurrently \"npm run dev:eleventy\" \"npm run dev:postcss\" \"npm run dev:postcss-nuxt\" \"npm run dev:nuxt\"",
+        "dev:nuxt": "dotenv -- npm run dev --workspace=nuxt",
```
ZJvandeWeg: @Yndira-E Why is this a seperate task? Why not expose this in the `npm run dev` task?

src/_includes/analytics/head.html
```diff
@@ -0,0 +1,61 @@
+<!-- Google Consent -->
```
ZJvandeWeg, line 1: @Yndira-E I don't quite understand why we wouldnt depend on the SEO package from Nuxt?

ZJvandeWeg (changes requested):

package.json
Yndira-E: Updated. It was originally separated to match the pattern already used by other scripts in this project that use `dotenv --`, but I’ve inlined it into the main dev script.

src/_includes/analytics/head.html
Yndira-E: The main reason for this approach is to keep a single source of truth for consent and analytics (GTM, Consent Mode, PostHog and HubSpot) while Eleventy and Nuxt coexist.

Once the migration is complete and Eleventy is no longer part of the stack, we can revisit the implementation and evaluate whether Nuxt-native solutions make more sense.

ZJvandeWeg (approved):

PR #5076, opened 2026-05-28
Use Cases section + floating Expert dock (structure only)

dimitrieh:
## Summary

Skeleton structure for a new **Use Cases** content shape on the marketing site. Structure only; page content is intentionally placeholder.

- `/use-cases/` hub + per-pattern pages, sharing a new `use-case.njk` layout.
- Floating FlowFuse Expert dock pinned to every use-case page, reusing the existing AI Expert chat.
- Reusable `related-resources.njk` block (case studies / white papers / blog) wired to a `usecase:slug` tag namespace using existing collection logic.

**No nav changes in this PR.** Follow-up PRs will target this branch and fill each use case with content from sales.


dimitrieh:
Re-opening as a same-repo PR on this repo so child PRs (Expert dock, sales content) can target this branch directly. See new PR below.

PR #5081, opened 2026-05-28
Add use case page: Production Monitoring & OEE

dimitrieh:
## Summary

First per-pattern content page on top of the **Use Cases structure** PR (#5079). Adds `/use-cases/production-monitoring-oee/` using the new `use-case` layout.

- Placeholder copy only (no client names, no real numbers, no attributed quotes).
- Mirrors the shape defined by `src/use-cases/_template.njk` in the structure PR.
- Renders into the `/use-cases/` hub grid automatically via the `use-case` collection tag.

**Base:** `feat/use-cases-structure`. Merge order: structure first, then this. To be replaced with real sales content when available.

### DIFF
--- a/src/use-cases/production-monitoring-oee.njk
+++ b/src/use-cases/production-monitoring-oee.njk
@@ -1,0 +1,21 @@
+---
+layout: use-case
+tags:
+  - use-case
+title: "Production Monitoring & OEE"
+meta:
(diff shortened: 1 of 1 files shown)


dimitrieh:
Renaming the example use case to 'Equipment & Process monitoring' per request. New PR replacing this one.

PR #5141, opened 2026-06-04
nuxt: Migrate ebooks, whitepapers and publications index

ZJvandeWeg:
## Description
Moving over 2 pages to Nuxt. Reusing some components we already have.


nuxt/pages/ebooks/[slug].vue
```diff
@@ -0,0 +1,80 @@
...
+              <li
+                v-for="(item, index) in page.contentTable"
+                :key="index"
+                v-html="item"
+              />
+            </ul>
+          </div>
+          <div class="product">
+            <div class="w-80 w-full mx-auto">
+              <HubSpotForm
```
Comment (ZJvandeWeg) [inline nuxt/pages/ebooks/[slug].vue:69]: @Yndira-E just wanted to call out how we can now create these components, and I like how 'neat' this is.

Yndira-E:
Hey! One thing to flag before merging: the ebooks are currently part of a shared Eleventy collection called `publications` that powers the [/resources/publications](https://flowfuse.com/resources/publications/) index page, it lists both ebooks and whitepapers together.

The collection is built from `src/ebooks/*` (deleted in this PR) and `src/whitepaper/*`. Once the ebooks move to Nuxt, they'll disappear from the publications index since 11ty won't know about them anymore.

This PR doesn't include a Nuxt version of the publications index either, so the page would silently lose the ebooks entries.

The folder structure doesn't make this obvious at all, totally understandable to miss it! But before merging, we'd need to also migrate `/resources/publications` and  `/whitepapers` to Nuxt and rebuild the index for ebooks + whitepapers

Yndira-E (approved):

ZJvandeWeg:
Thanks @Yndira-E !

PR #5163, opened 2026-06-08
nuxt: Blog overview migration

ZJvandeWeg:
## Description

Picking up another about 20 pages to migrate. 

Blog posts weren't move (yet) and that makes the blog migration a bit interesting soon, but for now this is a small change we can likely merge quickly.

## Checklist


src/blog/dashboard.njk
n-lark, line 1: Deleting these tag templates removes `/blog/how-to/`, `/blog/dashboard/`, etc. In prod those URLs now render an empty "Blog 1 of 1" page with no posts. They still look fine locally only because `11ty` is serving stale `_site` output.

We also still link to a few of these internally, like the "See All PLC Articles" CTA on the PLC landing page → `/blog/plc/`, plus the dashboard platform/ebook pages → `/blog/dashboard/`, so those now dead-end on the empty page.

Image

Is dropping them intentional? If so we'll likely want redirects to `/blog/`.

nuxt/pages/blog/index.vue
```diff
@@ -0,0 +1,141 @@
+<script setup lang="ts">
```
n-lark, line 1: The old blog template had an search box right under the title so blog search goes away with this change, same for tabs. Just confirming that's intentional?

nuxt/server/api/blog.get.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    const total = all.length
+    const pageCount = Math.ceil(total / PAGE_SIZE)
+
+    const posts: BlogPost[] = all.slice((page - 1) * PAGE_SIZE, page * PAGE_SIZE).map(p => ({
+        title: String(p.title),
+        date: p._date,
+        url: String(p.url),
+        description: String(p.description || ''),
+        image: p.image ? String(p.image) : null,
```
n-lark: So the image value is passed through as-is here, but ~26 posts have a relative path like `blog/2024/01/images/…png` with no leading slash. 

On a paginated URL like `/blog/12/` those resolve to `/blog/12/blog/2024/…png` and break, that's the bulk of the CI failures. Probably want to normalize to an absolute path here so we don't have to fix every post.

nuxt/pages/blog/[[page]].vue
```diff
@@ -0,0 +1,141 @@
...
+
+        </ul>
+
+        <nav aria-label="Pagination" class="pagination mt-4">
+            <ol class="flex flex-row w-full justify-between text-gray-600">
+                <li
+                    class="flex md:flex-initial w-40 justify-start pl-2 ff-nav-blog-p"
+                    :style="currentPage <= 1 ? 'opacity: 0; pointer-events: none;' : ''"
+                >
+                    <NuxtLink :href="prevHref">Previous</NuxtLink>
```
Comment (n-lark) [inline nuxt/pages/blog/[[page]].vue]: The prev/next links are always rendered and just hidden with opacity:0, so the out-of-range hrefs (/blog/0 on page 1, /blog/21 on the last page) stay in the HTML and fail the link checker. 

Could we v-if these so the link isn't emitted at all when there's no prev/next page? That clears the build error and also avoids the hidden-but-focusable link.

n-lark (changes requested):
Looks good, a couple things to confirm below. The broken build is partly the tag-template deletion I flagged below, but it's also catching two separate bugs in the new blog page, left comments for both.

nuxt/pages/blog/index.vue
ZJvandeWeg: Yeah, I looked in posthog, no one cares about search there.

nuxt/server/api/blog.get.ts
ZJvandeWeg: Yeah, that's a fair shout. I was trying to keep the diff as small as possible. I'll work on this

PR #5177, opened 2026-06-10
Migrate docs to Nuxt via sparse clone of flowfuse/flowfuse at build time

ZJvandeWeg:
Replaces the live-branch pipeline with a docs-source Nuxt module that sparse-clones docs/ from the public flowfuse/flowfuse repo on every build. Adds DocsLeftNav, useDocsNav, and docs/[...slug].vue reusing the handbook layout. Removes copy_docs.js, src/docs/, and all related 11ty wiring. Also adds ChecklistItem MDC component for upstream docs migration.

## Checklist


dimitrieh (changes requested):
Some errors it seems:

CleanShot 2026-07-15 at 12 01 03@2x

and we might want to adjust the colors here for less contrast

CleanShot 2026-07-15 at 12 02 11@2x 

Also search and ToC doesn't seem to work yet

CleanShot 2026-07-15 at 12 04 30@2x

Yndira-E:
Really like being able to see docs locally without cloning the repo, and in deploy previews, where it's currently not possible at all.

Seconding what @dimitrieh flagged, dug into the "errors" screenshot a bit further: the root cause is that `docs/README.md` (and a few other pages) embed raw HTML plus an inline `<script>` that defines a custom element (`icon-chevron-right`) for the card grid. `markdown-it` (11ty) passes that straight through, but `@nuxt/content`'s parser doesn't render it the same way, the whole block falls back to literal escaped text, which is why it shows up as a misaligned "code block" instead of the tile grid.

A couple of other observations:

- Right now, if the git clone in `docs-source.ts` fails (network hiccup, GitHub blip), the entire production build fails with no fallback. Worth adding some resilience here, either caching `nuxt/content/docs` as a fallback if the clone fails, or a simpler retry around the clone call, which would cover most transient network issues.
- If we go for the cache route, incremental updates become possible too, but with the current doc volume (~112 files) this isn't critical, more of a nice-to-have.
- Similarly, incremental refresh in local dev (right now it's "delete the folder or set `FLOWFUSE_DOCS_LOCAL`") would be nice for DX, but also not critical.

Once Algolia search, the right-hand ToC, and the other rendering issues Dimitrie called out are sorted, I'm happy to do a final CSS pass to make sure everything's visually consistent with the rest of the site.

.eleventy.js
```diff
@@ -55,7 +55,6 @@ console.info(`[11ty] Image build profile: ${IMAGE_BUILD_PROFILE}`)
 module.exports = function(eleventyConfig) {
 
     eleventyConfig.addDataExtension("yaml", contents => yaml.load(contents)); // Add support for YAML data files
-    eleventyConfig.setUseGitIgnore(false); // Otherwise docs are ignored
```
Yndira-E, line 58: We should restore this, it wasn't only needed for docs, `blueprints/` is also gitignored, which is why they're showing up in the failing tests.

Yndira-E:
~~The rest of the failing tests have to do with hyperlink being unable to resolve the docs links from old Eleventy content when the URL ends in `.md`, I guess it relies on `website/src` to map those, and the docs source no longer lives there.~~

This is not it, that .md url is not in the website pages. Needs a deeper dig.

dimitrieh:
Pushing some changes into this to resolve it. Would love for this to hit main sooner rather than later. Has been an outstanding request from me for a while now.

dimitrieh:
Pushed fixes for the failing link check and merged main in; `test_website` is green now.

Regressions from the docs move, all resolved:

- `/docs/*.md` links: `remark-docs-links` bailed on absolute paths and only rewrote ``, not `<a>`, so upstream docs links kept `.md` and 404'd. On the "deeper dig": those `.md` URLs aren't in the website pages, they're in the docs content itself (e.g. `hardware/introduction.md`), so hyperlink's `--sources` attribution pointed at unrelated blog files.
- Section anchors: `@nuxt/content`'s slugger differs from the old `markdown-it-anchor` scheme (which kept `? & . ( )` and repeated dashes), so links like `#device-agent---hardware` missed. Anchors are now slugified to match, plus the few legacy links left in site content.
- Blueprints: the migration dropped `setUseGitIgnore(false)` from `.eleventy.js` (added there for docs), but `src/blueprints/` is generated into a gitignored path too, so every blueprint page stopped building. Restored it.

Verified with a full production build and the same hyperlink check. Still open from the review, and not addressed here: the card-grid rendering (raw HTML plus an inline `<script>` custom element in `docs/README.md` falling back to escaped text), Algolia search, the right-hand ToC, and the contrast tweak. The clone-resilience idea in `docs-source.ts` looks like a sensible follow-up too.

dimitrieh:
Links, anchors, and the build all pass CI now, so what's left is docs quality rather than plumbing. Remaining before this is merge-ready, roughly in priority order:

1. Card-grid rendering: the tile grid in `docs/README.md` (and a few other pages) relies on raw HTML plus an inline `<script>` defining a custom element, which `@nuxt/content` escapes rather than renders, so it falls back to a literal code block. Converting that grid to an MDC component (same approach as `ChecklistItem`) is the durable fix, and likely needs a matching tweak in the upstream `flowfuse/flowfuse` docs.

2. Algolia search and the right-hand ToC: not wired up yet on the Nuxt docs pages.

3. Clone resilience in `docs-source.ts`: docs are cloned from `flowfuse/flowfuse` at build time with no retry or fallback, so a transient GitHub or network blip fails the entire production deploy. A simple retry around the clone (optionally caching `nuxt/content/docs` as a fallback) closes that. Worth treating as a blocker rather than a follow-up, since it gates prod deploys.

4. Contrast and a light CSS pass once the above render correctly, to line the docs up visually with the rest of the site.

Lower priority: incremental docs refresh for local-dev DX, and incremental updates if we go the cache route (not critical at the current doc volume).

Since docs are a high-traffic surface, I'd hold merge until at least the rendering, search, and ToC are in, so we don't ship a visible downgrade from the current docs.

---

This is now fixed in this stacked PR: https://github.com/FlowFuse/website/pull/5390

.eleventy.js
dimitrieh: Restored in 85362bc, with the comment updated to name the real reason (`src/blueprints/` is generated into a gitignored path). That was exactly what the blueprint pages failing to build came down to.

Yndira-E:
Thanks @dimitrieh for implementing all the suggestions from my review.

I did a couple of tweaks on the styling, I'd say is ready to merge.

Yndira-E (approved):

KristopherLeads (approved):

ZJvandeWeg:
I appreciate us wanting to merge this, I'll hold off for a day or so. 

I'll recheck the visuals etc after the product line changes are all merged.

dimitrieh:
Also, conflicts

dimitrieh:
Conflicts fixed

PR #5180, opened 2026-06-10
Upgrade Tailwind CSS from v3 to v4

ZJvandeWeg:
- Replace tailwindcss@3 + autoprefixer + postcss-import with tailwindcss@4 + @tailwindcss/postcss
- Delete tailwind.config.js; migrate all theme config to @theme{} in style.css
- Load typography plugin via @plugin directive instead of require() in config
- Replace all @tailwind base/components/utilities directives with @import "tailwindcss"
- Replace all theme(colors.x.y) calls with var(--color-x-y) across all CSS files
- Port custom utilities (grid-cols-header, transition-*, text-shadow-header) to @utility
- Port safelist to @source inline()
- Remove style.base.css (was only @tailwind base;)
- Remove TAILWIND_MODE=watch from dev scripts (v3 JIT-only flag)


ZJvandeWeg:
@Yndira-E This change creates some weird artifacts on the new homepage, and I have not figured out yet why.

Yndira-E:
I'll check it out in the morning

Yndira-E:
Ok, homepage "fixed". This affected everything, not just the homepage and I'm sure I'll run into more things to fix, but I'll deal with those in separate PRs as I spot them.

The thing is, in v3, specificity determined which styles won. In v4, layers take precedence over specificity, and the layer order wasn't set up correctly for our overrides to work.

Yndira-E (approved):

PR #5214, opened 2026-06-16
Blog: Fixing PLC Communication Latency: Where the Milliseconds Really Go

sumitshinde-84:
## Checklist


dgatti0213:
@sumitshinde-84 Can you help get the page preview going? I think it failed, right? Or am I clicking the wrong link?

dgatti0213:
This is really solid, the "follow the time" spine and the concrete numbers carry it well. 

One thing before it ships in the protocol section. We shouldn't name EtherNet/IP as the surest pick. The honest framing is that there's no universal best, because the protocols you reach a PLC/device through at the integration layer (Modbus TCP, OPC UA, MQTT, EtherNet/IP explicit messaging, API) are not real-time protocols. The deterministic behavior lives below that, in the controller's own I/O and scan, which isn't what the runtime polls. So the recommendation becomes a method, not a ranking: pick a protocol the device supports and that's well implemented on both ends, apply the techniques in the article, and only switch if the timing still won't hold. Worth scoping that to the northbound/supervisory layer so the real-time fieldbuses below (EtherCAT, PROFINET IRT) don't become a gotcha. This also ties straight into the "build it at the source" section, so it makes the piece more coherent, not less. Same change needs to flow into the HowTo step and the FAQ answer so the page doesn't contradict itself.

Nice work man!

sumitshinde-84:
> @sumitshinde-84 Can you help get the page preview going? I think it failed, right? Or am I clicking the wrong link?

yes, there are some errors at Nuxt, will take look at it

sumitshinde-84:
> This is really solid, the "follow the time" spine and the concrete numbers carry it well.
> 
> One thing before it ships in the protocol section. We shouldn't name EtherNet/IP as the surest pick. The honest framing is that there's no universal best, because the protocols you reach a PLC/device through at the integration layer (Modbus TCP, OPC UA, MQTT, EtherNet/IP explicit messaging, API) are not real-time protocols. The deterministic behavior lives below that, in the controller's own I/O and scan, which isn't what the runtime polls. So the recommendation becomes a method, not a ranking: pick a protocol the device supports and that's well implemented on both ends, apply the techniques in the article, and only switch if the timing still won't hold. Worth scoping that to the northbound/supervisory layer so the real-time fieldbuses below (EtherCAT, PROFINET IRT) don't become a gotcha. This also ties straight into the "build it at the source" section, so it makes the piece more coherent, not less. Same change needs to flow into the HowTo step and the FAQ answer so the page doesn't contradict itself.
> 
> Nice work man!

Thanks, Drew! 🙌 I've applied all the suggestions you provided. Could you please take another look and approve the PR if everything looks okay?

PabloFilomeno83 (approved):

dgatti0213 (approved):
Nice work!

PR #5266, opened 2026-06-30
Remove Pablo from team page

allthedoll:
## Description

This PR:
- Removes Pablos' entry from the team page
- Removes old fuser headshots that were never removed (clean up), including Pablo's 

## Related Issue(s)

N/A

## Checklist


knolleary:
Rather than delete, we should move over to `src/_data/guests` - so that any author-links in blog posts don't break.

Also, the headshot tidy-up mentioned in the description isn't present in the commits.

allthedoll:
@knolleary I forgot to push the commit for the images; it's in there now. Also TIL on the moving to guests... I should undo the delete for Greg??

knolleary:
Some of those headshots are guest authors on the blog so shouldn't be removed. Can you revert that commit?

allthedoll:
@knolleary i've reverted the wrong commit and now git is misbehaving and i'm lazy so i'll restart with another pr

PR #5311, opened 2026-07-10
Correct install/restart note and device terms

allthedoll:
## Description

This PR corrects the install note on the certified nodes pages. It prev read:

> Existing devices and hosted instances will not pick up newly installed nodes until they are restarted.

That's inaccurate. Installing a node into the runtime does **not** require a restart, only *updating* an already-installed node does. The note now reads:

> Newly installed nodes are picked up automatically — no restart needed. Restart is only required when you update a node that's already installed: restart any remote instance or hosted instance running the previous version.

Also replaces the retired "devices" term with "remote instances."

## Related Issue(s)

N/A; @dimitrieh flagged this to me in a DM and @knolleary confirmed the behavior. ✨ 

## Checklist


knolleary (commented):
I think you have some stray files included in this - `.fuse_hidden...` etc

allthedoll:
@knolleary yeah i was just fixing it while you peeped this. they're gone now. ignore my sins.

knolleary (approved):

sumitshinde-84:
@allthedoll @knolleary @dimitrieh I think the original note is correct here.

> Existing devices and hosted instances will not see newly added nodes until they are restarted. Restart any instance you plan to install nodes on so it picks up the updated catalogue.

The PR description says "newly installed nodes," but the docs say "newly added nodes." I agree the note could use a bit more clarity, but I think what it says is correct.

I also confirmed this with @hardillb. It refers to newly added nodes means in  the catalogue and existing devices and hosted instances won't pick up those new catalogue entries until they are restarted.

sumitshinde-84:
I think this is much clear:
> If your device or hosted instance was already running before node was enabled for your team, it won't show the package in the Install tab search. Restart the instance or device first, then repeat the steps above.

PR #5316, opened 2026-07-10
Fix/hubspot form fallback seo

sumitshinde-84:
## Description

The HubSpot consent fallback block (shown when a form or meeting scheduler
fails to load due to privacy/cookie settings) was rendered visible in the
markup by default. This meant its placeholder copy, e.g. "Hmm… there was
supposed to be a form here", was present in the initial DOM and could be
picked up by search engine crawlers, polluting indexed page content.

This change adds the hidden Tailwind class to the fallback container so it
is hidden by default and only revealed via JS when the HubSpot embed genuinely
fails to load.

## Checklist


Yndira-E:
@sumitshinde-84 if it's hidden without conditions, then it never shows the fallback message. Check out the book a demo page.

Screenshot 2026-07-13 at 10 17 30

Yndira-E (approved):

sumitshinde-84:
I've updated the PR to exclude the HubSpot calendar,  this only applies to forms now.

I dug into this a bit more, and you're right that there's no logic in that file itself. consent-fallback.njk just sets a class for default behavior, it doesn't do anything on its own.

The actual show/hide happens in `base.js`, in a function called `ffCreateHubSpotForm`. That's what every form embed calls. It tries to load the HubSpot script, and:
- if the script fails to load or breaks somehow (ad blocker, strict browser privacy, whatever), it removes the hidden class and shows the fallback
- if the form loads fine, it just leaves it hidden

PR #5321, opened 2026-07-13
add auto-calculated read time and "preferred source" button

sumitshinde-84:
## Description

This PR adds two reader-facing enhancements to blog article pages.

### 1. Auto-calculated read time
**What:** Each article now displays an estimated "N min read" in the meta line, next to the author and date.

a visible read-time estimate sets reader expectations up front, which improves engagement and reduces bounce, readers are more likely to start an article when they know the time commitment. It's calculated automatically from the article content, so authors don't have to set or maintain it manually and it can never drift out of sync with the actual content.

A new `readingTime` Eleventy filter (`.eleventy.js`) strips HTML from the rendered content, counts the words, and divides by 200 wpm (rounded up, minimum 1 min). Rendered in `post.njk` as `{{ content | readingTime }} min read`.

### 2. **"Make us a preferred source" button** ( really important) 
**What:** A "See more FlowFuse in Google Search" prompt in the right-hand sidebar of each article, just below the Table of Contents. It's a text link, "Click to make us your preferred source", pointing to https://www.google.com/preferences/source?q=flowfuse.com (opens in a new tab), with a short note underneath on the next step: "Then check the box next to flowfuse.com."

blog-pages

Google now lets users mark sites as preferred sources, which boosts how prominently our content surfaces for those users in Search/Top Stories. Adding a one-click button on our blog articles makes it easy for engaged readers to opt in, increasing repeat organic reach to FlowFuse content.

### 2. Replace separator "." with "|" in top

It replaces the separator, which was previously a dot (·), with a pipe (|). I think the pipe looks good with the Author, Real-time, and Make Us a Preferred Source buttons.

## Checklist

### DIFF
--- a/src/_includes/layouts/post.njk
+++ b/src/_includes/layouts/post.njk
@@ -29,27 +29,27 @@
                 {% if subtitle %}
                     <h4>{{ subtitle | safe }}</h4>
                 {% endif %}
-                <div class="flex flex-wrap items-center gap-x-2 gap-y-2 text-sm text-gray-500 mt-4">
+                <div class="flex flex-wrap items-center gap-x-3 gap-y-2 text-sm text-gray-500 mt-4">
                     <span class="flex flex-wrap items-center gap-1">
+                        <span>By</span>
                         {% set comma = joiner() %}
                         {%- for author in authors %}
                             {{ comma() }} <span class="font-medium">{{ people[author].name if people[author] else 'FlowFuse' }}</span>{% if people[author] and people[author].title %}<span class="text-gray-400">, {{ people[author].title }}</span>{% endif %}
                         {%- endfor %}
-                        <span class="text-gray-500">•</span>
+                        <span class="text-gray-300">|</span>
                         {%- if lastUpdated -%}
                         <time datetime="{{ lastUpdated | dateToRfc3339 }}">Updated {{ lastUpdated | shortDate }}</time>
                         {%- else -%}
                         <time datetime="{{ date | dateToRfc3339 }}">{{ date | shortDate }}</time>
                         {%- endif -%}
-                        <span class="text-gray-500">•</span>
+                        <span class="text-gray-300">|</span>
                         <span>{{ content | readingTime }} min read</span>
                     </span>
-                    <a href="https://www.google.com/preferences/source?q=flowfuse.com"
+                    <a class="ff-btn ff-btn--highlight !font-normal !text-sm !py-1 !px-3 max-sm:mx-auto"
+                       href="https://www.google.com/preferences/source?q=flowfuse.com"
                        target="_blank"
-                       rel="noopener"
-                       class="inline-flex items-center gap-1.5 rounded-md border border-gray-300 px-3 py-1.5 text-xs font-medium text-gray-700 transition-colors hover:border-indigo-400 hover:text-indigo-600">
-                        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="currentColor" class="w-4 h-4 text-yellow-500"><path fill-rule="evenodd" d="M10.788 3.21c.448-1.077 1.976-1.077 2.424 0l2.082 5.007 5.404.433c1.164.093 1.636 1.545.749 2.305l-4.117 3.527 1.257 5.273c.271 1.136-.964 2.033-1.96 1.425L12 18.354 7.373 21.18c-.996.608-2.231-.29-1.96-1.425l1.257-5.273-4.117-3.527c-.887-.76-.415-2.212.749-2.305l5.404-.433 2.082-5.006z" clip-rule="evenodd" /></svg>
-                        Make us a preferred source
+                       rel="noopener">
+                        <span class="items-center">Make us a preferred source</span>
                     </a>
                 </div>
             </div>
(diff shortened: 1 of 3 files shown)


Yndira-E:
Hey @sumitshinde-84 love the read time calculator! 

Regarding the preferred source link, I like the idea, but I think this might work better as a secondary CTA rather than a primary one in the article header.

If the target audience is engaged readers, I’d expect it to perform better after they’ve had a chance to read the article, or in the right-hand sidebar. That also avoids competing with the article title.

I’d also consider making it a text link instead of a button, with copy that explains the user benefit. Something like:

> Liked this article? Want to see more FlowFuse tutorials in your Google search results? Make FlowFuse your preferred source in Google.

At the moment, “Make us a preferred source” is quite company-centric and doesn’t explain where or why someone would want to do it.

sumitshinde-84:
Thanks @Yndira-E ! I agree with not making it a primary CTA and with using a text link instead of a button. However, I think moving it away from the top could mean missing an opportunity. Google has introduced this feature to let users choose the sources they want to see more often, so giving it good visibility could potentially help us gain more organic exposure from readers who find our content valuable.

I think the best approach would be to keep it in its current position but make it a subtle text link rather than a prominent button. I also think “Make us a preferred source” is already fairly clear in this context, readers who find our content valuable can choose FlowFuse as a preferred source to see more content like this in Google, especially since our content focuses on a specific niche.

I like the idea of making the copy more user-focused and clearly explaining the benefit, but I think the longer copy would add too much visual clutter to the article header. That’s why I’d prefer keeping a short, clear text link at the top.

Also, if we appear more frequently in the search results of people who have made us a preferred source, there’s a higher chance that they’ll eventually visit our high-intent pages as well. Over time, they may not even consciously remember that they previously selected us as a preferred source, but the increased visibility and familiarity with FlowFuse could still influence them to click on primary cta's.

Yndira-E:
I still think the right-hand sticky sidebar would be a better fit. It keeps the link visible throughout the reading experience without competing with the article title, and it gives us enough space to explain the benefit.

My main concern with keeping it in the header isn’t visibility, it’s timing and context. The current copy assumes readers already know Google’s Preferred Sources feature. “Make us a preferred source” doesn’t really answer _preferred source where?_ or _source of what?_. In the sidebar, we could add a short explanation like “Want to see more FlowFuse tutorials in your Google Search results?”, which makes the value much clearer while keeping the article header focused on the content.

I also tried the flow myself, and I found it a little confusing. The link doesn’t actually make FlowFuse a preferred source—it opens Google’s Source Preferences page, where you then have to discover that the next step is checking the box on the right. Since the action isn’t as direct as the CTA suggests, I think it’s even more important to provide a bit of context about what users should expect after clicking.

sumitshinde-84:
> I still think the right-hand sticky sidebar would be a better fit. It keeps the link visible throughout the reading experience without competing with the article title, and it gives us enough space to explain the benefit.
> 
> My main concern with keeping it in the header isn’t visibility, it’s timing and context. The current copy assumes readers already know Google’s Preferred Sources feature. “Make us a preferred source” doesn’t really answer _preferred source where?_ or _source of what?_. In the sidebar, we could add a short explanation like “Want to see more FlowFuse tutorials in your Google Search results?”, which makes the value much clearer while keeping the article header focused on the content.
> 
> I also tried the flow myself, and I found it a little confusing. The link doesn’t actually make FlowFuse a preferred source—it opens Google’s Source Preferences page, where you then have to discover that the next step is checking the box on the right. Since the action isn’t as direct as the CTA suggests, I think it’s even more important to provide a bit of context about what users should expect after clicking.

@Yndira-E Done! I’ve updated the PR, please check the PR description. I think that if we’re moving it to the sidebar, we should make it a secondary button to give it more visibility. But I’ll leave the decision to you. Feel free to merge it!

Yndira-E:
@sumitshinde-84 Could you please push the latest changes? The preview is still showing the first version

sumitshinde-84:
> @sumitshinde-84 Could you please push the latest changes? The preview is still showing the first version

Done Updated

src/_includes/layouts/post.njk
```diff
@@ -101,6 +104,16 @@
...
                     {% include "layouts/toc.njk" %}
+                    <div class="mb-6 pt-6 border-t-2">
+                        <h3 class="mb-3">Prefer FlowFuse in Search</h3>
+                        <a class="text-sm font-medium text-blue-700 hover:text-blue-800 hover:underline"
+                           href="https://www.google.com/preferences/source?q=flowfuse.com"
+                           target="_blank"
+                           rel="noopener">
+                            Add flowfuse.com as a preferred source
+                        </a>
+                        <p class="text-xs text-gray-500 mt-1 mb-0">On the page that opens, check the box next to flowfuse.com to see more of our articles in your results.</p>
```
Yndira-E: @sumitshinde-84 What do you think about this version? I think the heading should focus on the reader rather than the brand. Then the link introduces FlowFuse, and the supporting text provides the context and explains what happens after clicking.

```suggestion
                        <h3 class="mb-3">Like what you’re reading?</h3>
                        <a class="text-sm font-medium text-blue-700 hover:text-blue-800 hover:underline"
                           href="https://www.google.com/preferences/source?q=flowfuse.com"
                           target="_blank"
                           rel="noopener">
                            Add FlowFuse as a preferred source
                        </a>
                        <p class="text-xs text-gray-500 mt-1 mb-0">On the page that opens, check the box next to flowfuse.com to see more of our articles in your Google Search results.</p>
```

sumitshinde-84: Yes I like it

Yndira-E (approved):

PR #5336, opened 2026-07-16
Blog: Handling Clock Drift in Distributed Edge Devices

sumitshinde-84:
## Checklist


dgatti0213:
Its on point but unless you are doing motion control, PLCs doent care about time. The system talking to it does. 

Generally PLCs arent the ones doing the data time stamping and the protocol that is used to talk to the PLC doesnt have a time association. The point is valid but you need to bring it up a level. 

Scada, HMI, Historian, and data loggers are who experience this. The only time a PLC feels this is if the data you want to log is faster than the protocol you are using and you have to buffer in an array the data capture.

dgatti0213 (approved):
Nice work!

PR #5371, opened 2026-07-23
Holistic redesign PoC: industries, AI page, nav + interaction polish

dimitrieh:
Holistic proof-of-concept, not intended to be merged as-is. Sharing for feedback. Can be broken out into multiple PRs.

Scope: industry pages (animated video heroes, card images), AI page rebuild (sticky capability sections), site-wide button/nav/footer interaction polish, homepage copy.

PR #5067, opened 2026-05-27
Add a top utility bar to the homepage header

dimitrieh:
## Description

Adds a slim top utility bar above the main navigation on the website header.

CleanShot 2026-06-15 at 12 24 22@2x

- Left: a rotating announcement that cycles through live events (replacing the old `events-banner`, whose rotation behaviour now lives in the bar) plus standing promos for FlowFuse Expert and installing the device agent.
- Right (desktop): quick links: Blog, Docs, Support, Download, Sign In.
- Mobile: Sign In stays visible in the bar, a Contact Us CTA sits next to the hamburger menu button, and the menu close button stays above the open menu overlay.

Docs and Sign In move into the bar on desktop and remain reachable on mobile. The old `events-banner.njk` is removed; its rotation behaviour is preserved in the bar.

## Related Issue(s)

Issue to be added: the tracking issue for what this is part of still needs to be created and linked.

## Checklist

- [ ] I have read the contribution guidelines
- [x] I have considered the performance impact of these changes
- [ ] Suitable unit/system level tests have been added and they pass (N/A: static Eleventy template/markup, no test harness)
- [ ] Documentation has been updated (N/A)
- [ ] For blog PRs, an Art Request has been created (N/A: not a blog PR)

PR #5375, opened 2026-07-24
Add mega navigation, top utility bar, and rebuilt footer

dimitrieh:
## Description

**Outcome:** site-wide wayfinding is rebuilt. A full-width mega navigation (hover intent, animated underlines, and a marketing-configurable highlight card per dropdown), a top utility bar, and a footer whose sections mirror the nav. The Nuxt header, footer, and utility bar are brought to parity with the Eleventy site, including the By Integration footer column and the Certifications link.

**Builds on:** the Shop Floor Communication page beneath it, so the new "By Use Case" nav and footer entries resolve to a live page.

**Provides for later PRs:** the shared `style.css` and `base.njk` layout foundation that every subsequent PR in this stack relies on, plus the nav/footer structure those pages plug into.

Also removes the retired events banner.

### Updated after review

- **Footer links and FAQ headers underline again on hover.** Reserving the sweeping underline for the main navigation removed the plain underline along with the motion, which left both with a colour change and nothing else. Footer links now underline statically instead of sweeping, and FAQ headers underline alongside their colour shift. Only the animation is navigation-only, not the underline itself.
- **The mega panel hover underline hugs the label text.** It was anchored to the `<a>` box, and a mega link is a full-width flex row holding an icon and a label, so the line ran the width of the dropdown column. It now hangs off the `.ff-nav-label` span, which is what the top-level nav items already did for the same reason.
- **The logo hover effect is fully out of this PR now.** The animation moved to #5388 earlier, and the five-stop gradients it animates have now followed it, so the wordmark and square SVGs here carry plain fills again. With all five stops set to one colour they rendered as a flat fill and did nothing here, so nothing inert is left behind and #5388 stays independently acceptable or declinable.
- **Header icons stay inline here.** Converting the Heroicons ones to Nuxt UI `<UIcon>` needs a new Iconify collection and would silently break CSS that reaches inside some of the SVGs, so it moved to #5389.

## Related Issue(s)

Split from #5371. Implements the intent of #5067 (utility bar).

## Checklist

---

```mermaid
flowchart LR
    main([main])
    P75 --> P89["#5389 nav icons"]
    class P75 here;
```


Yndira-E:
@dimitrieh looking good! Thanks for your work on this. A few notes:

- The new “Download” CTA that leads to the Device Agent docs, I’m not sure I understand the reasoning behind this. Could you elaborate? Also, “Download” doesn’t seem to set the right expectation. I’m not sure why this CTA needs to appear on every page. Would someone visiting our website for the first time really be looking to download the Device Agent? At what stage of their journey would they actually need this?

- Nav on Eleventy pages doesn't match the nav on Nuxt pages, columns, items, icons, font size...
Eleventy
Screenshot 2026-07-24 at 15 28 38

Nuxt
Screenshot 2026-07-24 at 15 28 46

- The mobile nav is also different in terms of layout and items.
Eleventy
Screenshot 2026-07-24 at 15 31 53

Nuxt
Screenshot 2026-07-24 at 15 32 17

- The AI image under "platform" on the main nav looks cropped
Screenshot 2026-07-24 at 15 17 46

- The Food & Beverage item has the wrong icon (on Eleventy pages)
Screenshot 2026-07-24 at 15 18 17

- The icon for the "Certification" item is missing
Screenshot 2026-07-24 at 15 21 44

- There’s some inconsistency in the responsive behaviour. When the image disappears, the columns expand to fill the available width. However, at narrower widths, they stop doing that. I’d expect the opposite behaviour.
Screenshot 2026-07-24 at 15 23 51
Screenshot 2026-07-24 at 15 23 58

- The top border on the footer feels unnecessary. The background already provides enough separation from the content.
Screenshot 2026-07-24 at 15 23 34

- I think there’s an opportunity to improve how the footer links are grouped. Is there any rationale behind the current grouping? I tried to understand the relationship between the grouped items, but it wasn’t obvious to me.
Screenshot 2026-07-24 at 15 33 50

- I’m not sure the underline needs to extend across the full width of each menu item. I think it would feel cleaner if it only matched the width of the text.

- I don’t think the hover effect on the logo adds much. I’d probably keep it static.

dimitrieh:
@Yndira-E Thanks! looking at this asap. 

Argh, I had not tested with the nuxt pages, just made sure the links were there as well. Looking at this in a bit

> The new “Download” CTA that leads to the Device Agent docs, I’m not sure I understand the reasoning behind this. Could you elaborate? Also, “Download” doesn’t seem to set the right expectation. I’m not sure why this CTA needs to appear on every page. Would someone visiting our website for the first time really be looking to download the Device Agent? At what stage of their journey would they actually need this?

I am welcoming suggestions. But we need to lead people directly to use the device agent (Nick is making this a conversion funnel this release, people can start using FlowFuse from the terminal without first needing to login to the website, they will get guided through that journey starting with the terminal). The easiest first step was leading them towards the download docs for the device agent.  Basically, this should be considered an equal opportunity to "login or free trial". 

See for more info https://flowfuse.slack.com/archives/C032Q63FGG1/p1784901796597059?thread_ts=1784898894.095579&cid=C032Q63FGG1

dimitrieh:
**FYI:** Rebased onto the updated split/shop-floor, and onto current main underneath (main had moved 14 commits since Friday). Deploy preview is rebuilding, the link above will update.

`.eleventy.js` and `src/css/style.css` had both changed on main from the Starter/Pro removal, so those two are resolved in favour of main and the removed tiers stay removed.

No review feedback is addressed yet, this push was mechanical only. All the notes above are still on my list.

dimitrieh:
@Yndira-E on the Food & Beverage icon: agreed it is wrong, it is currently using `snowflake` (cold chain).

I went looking for a replacement and could not find a good one. Nothing in `src/_includes/components/icons/` is a reasonable food or beverage metaphor, and the set is Heroicons based (24x24, stroke-width 1.5, `currentColor`), which has no food or beverage icon either. The closest existing options are `factory`, `circle-stack` or `funnel` and none of them really reads as Food & Beverage, so swapping to one of those would not actually fix your comment.

So this needs a specific SVG, and I would rather you decide which:

- there is an icon you have in mind that I have missed, in which case tell me which and I will wire it up
- or a new one needs to be created in the same style, the way `wind-turbine.svg` was made for Renewables

Leaving `snowflake` in place and this item open until you say which way to go.

dimitrieh:
On the Download CTA, updating this: the label is changed rather than kept.

It now reads **Install on edge** instead of Download, in both the announcement line and the quick links, and in both the Eleventy include and the Nuxt component so the two stay consistent. Your point stands that "Download" sets the wrong expectation, since the link goes to installing the device agent on hardware rather than downloading a product.

Keeping the placement as is. It links to the edge first onboarding, which is likely releasing with 2.33, and the intent is that it sits as an equal peer to login and free trial: the terminal first path becomes a conversion funnel this release, where people can start using FlowFuse from the terminal without logging into the website first and get guided from there.

Still open to suggestions on both. The thing I would most like to explore as a follow up is giving the homepage an option to copy the install command directly, which sets the expectation better than any single label can.

dimitrieh:
On the footer grouping: there is a rationale, but it is a layout one rather than a semantic one, which is why it does not read as a relationship between the items.

The three upper sections mirror the mega nav panels exactly: Platform, Solutions and Resources, with the same group labels in the same order. The last row then reuses the same `[2fr_3fr_1fr]` track geometry so its columns line up with the columns above it: the logo sits under Platform, the next three columns land under Solutions' three sub-columns, and the final pair lands under Resources. The number of links per column is whatever fills those four aligned positions, so the split is driven by the alignment, not by grouping the items by meaning.

I did try labelling that row and grouping it by meaning instead (Company, Partners and Services, Support, matching the Company nav panel). It reads better as a list but it empties the column that lines up under Resources and breaks the alignment down the right side of the page, so I have reverted it.

Happy to revisit if you would rather have meaningful grouping than the alignment, but that is a trade between the two and I would want your call on which matters more here. Left as it was for now.

One unrelated fix while I was in this file: the nav said "Aviation Aerospace" while the footer link and the page's own title both use the ampersand. The nav now matches.

dimitrieh:
**Review feedback addressed.** Going through the list in order:

| Item | Status |
|---|---|
| Download CTA label | Relabelled to **Install on edge**, placement kept. See the updated comment above. |
| Eleventy vs Nuxt nav mismatch (desktop) | **Fixed.** See below, this was worse than it looked. |
| Mobile nav differs | **Fixed** by the same change: both share one responsive nav, and the mobile CTA block was already structurally identical. |
| AI image cropped under Platform | Fixed. Highlight images now use `contain` rather than `cover`, since they come from each page's own front matter and are not all 16/9. |
| Food & Beverage wrong icon | **Open**, needs a specific SVG from you. See my separate comment. |
| Certifications icon missing | Fixed. It was the only nav icon wrapping its artwork in a clip path, a design-tool export leftover, and nested inside the wrapper svg the shortcode generates that clip resolved to nothing. The clip rectangle enclosed every path anyway, so removing it changes no geometry. |
| Responsive column behaviour inverted | Fixed. The panel now spreads its columns across the md range as it already did above 1024px, instead of clustering them centrally the narrower the viewport got. |
| Footer top border | Removed. |
| Footer link grouping | **Explained rather than changed**, see my separate comment: the grouping is a column-alignment constraint, not semantics. Happy to revisit if you would rather have meaning than alignment. |
| Underline wider than the text | Fixed. It is now anchored to the label span rather than the menu item box, so it matches the text width. That also removed the need for a per-breakpoint inset. |
| Logo hover effect | Removed, in the header and the footer. |

On the Nuxt side, the drift was larger than the screenshots showed. Measured from the rendered DOM of both:

- the whole **By Integration** column was missing from Solutions, so Redis, OPC UA, RTSP, EtherNet/IP, MQTT, ONNX, LLM Nodes, MCP and the see-all link did not appear at all
- **Certifications** was missing from the Company column
- Platform split its columns 5/4/7 instead of 3/6/7, with Integrations and Blueprint Library under Product rather than Components
- the Company panel reserved five grid rows for six
- a `Docs` item existed at top level on mobile that the Eleventy nav does not have
- no item carried the label span the underline is anchored to, so once the underline moved to that span it stopped rendering on Nuxt pages entirely
- the Free Trial CTA was missing the class that gives it its rest colour and hover underline
- the header CTAs carried no `capture()` calls, so those clicks went untracked on Nuxt pages

Both now render the same 58 labelled links and an identical grid structure. Font size was never the issue, both are 14px from the shared stylesheet.

Separately, the utility bar interaction states differed: the announce-rotation and hover-underline rules lived in an inline `<style>` in the Eleventy include only, so Nuxt links never underlined on hover. Those rules moved into the shared stylesheet both sites load, so the two cannot drift again.

KristopherLeads (approved):

nuxt/components/AppFooter.vue
```diff
@@ -1,70 +1,231 @@
...
+            <section class="mt-12 border-t border-gray-300 pt-9">
+                <!-- Mirrors the upper section's [2fr_3fr_1fr] tracks so the logo lines up
+                     under Platform, the three groups under Solutions' sub-columns, and the
+                     last group under Resources. -->
+                <div class="grid grid-cols-1 lg:grid-cols-[2fr_3fr_1fr] gap-x-8 gap-y-8 text-sm">
+                    <div>
+                        <a href="/" class="ff-logo-link ff-footer-logo block w-40 no-underline hover:no-underline" aria-label="FlowFuse home">
+                            <svg class="h-full w-full" viewBox="0 0 402 70" xmlns="http://www.w3.org/2000/svg">
+                              <defs>
+                                <linearGradient id="ffWaveWordmarkFooter" gradientUnits="userSpaceOnUse" x1="0" y1="0" x2="402" y2="0">
```
Yndira-E: @dimitrieh isn't this what was moved to a different pr? I don't see the hover effect on the preview but what's the use of `linearGradient` here?

nuxt/components/AppHeader.vue
```diff
@@ -122,94 +188,179 @@ onMounted(() => {
...
+              <li><a class="flex items-center gap-2" href="/platform/features/"><svg class="ff-icon ff-icon-sm" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor"><path stroke-linecap="round" stroke-linejoin="round" d="M11.48 3.499a.562.562 0 011.04 0l2.125 5.111a.563.563 0 00.475.345l5.518.442c.499.04.701.663.321.988l-4.204 3.602a.563.563 0 00-.182.557l1.285 5.385a.562.562 0 01-.84.61l-4.725-2.885a.563.563 0 00-.586 0L6.982 20.54a.562.562 0 01-.84-.61l1.285-5.386a.562.562 0 00-.182-.557l-4.204-3.602a.563.563 0 01.321-.988l5.518-.442a.563.563 0 00.475-.345L11.48 3.5z"/></svg><span class="ff-nav-label">Features</span></a></li>
+              <li><a class="flex items-center gap-2" href="/platform/security/"><svg class="ff-icon ff-icon-sm" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor"><path stroke-linecap="round" stroke-linejoin="round" d="M16.5 10.5V6.75a4.5 4.5 0 1 0-9 0v3.75m-.75 11.25h10.5a2.25 2.25 0 0 0 2.25-2.25v-6.75a2.25 2.25 0 0 0-2.25-2.25H6.75a2.25 2.25 0 0 0-2.25 2.25v6.75a2.25 2.25 0 0 0 2.25 2.25Z" /></svg><span class="ff-nav-label">Security Statement</span></a></li>
```
Yndira-E: Instead of inline `svg`, could we use [Vue icons](https://ui.nuxt.com/docs/components/icon)?

Yndira-E:
I still see the line taking all of the space instead of the text lenght

Screenshot 2026-07-28 at 09 25 35

nuxt/components/AppFooter.vue
dimitrieh: The gradient does nothing in this PR, so this is a fair catch. All five stops hold the same colour, which renders exactly like a flat fill. It only becomes meaningful once the keyframes shift the stop colours, and those moved to #5388.

Fixed: the wordmark and square SVGs go back to plain fills on both front ends, and the wave-only `ff-logo-link` and `ff-wave-square` hooks are gone with them. `ff-footer-logo` stays, because `style.css` uses it to exempt the footer logo from the footer link underline rules.

On why it was split that way: CSS can only animate `stop-color` on `<stop>` elements that already exist, so the wave needs the gradient present even at rest. That is what kept #5388 to a stylesheet-only diff. Leaving the markup down here was the wrong half of the split, so the gradients now travel with the keyframes that need them. I checked that what #5388 restores is byte-identical to what this PR carried before, so nothing is lost and that PR is now self-contained.

nuxt/components/AppHeader.vue
dimitrieh: Worth doing, but it is not a drop-in swap, so I scoped it out of this PR and into #5389 rather than grow the nav diff:

- These are Heroicons, and the only Iconify collection in the tree is `@iconify-json/lucide`, a Nuxt UI transitive dependency. Without adding `@iconify-json/heroicons`, Nuxt Icon resolves the names over the network at runtime. With it added, dev startup reports `Nuxt Icon server bundle mode is set to local`, so nothing is fetched from api.iconify.design.
- `@nuxt/icon` defaults to `mode: "css"`, so `UIcon` renders a `<span>` with a mask rather than an `<svg>`. That is fine for the dropdown icons, because `ff-icon-sm` sets `stroke-width: 1.5px` and Heroicons outline is drawn at 1.5 on the same viewBox, so the stroke renders identically. It does silently break the rules that reach inside the SVG: `.ff-icon--gh path`, `.ff-icon--down path` and `.ff-icon--solid { fill }`. The GitHub mark, the chevrons and the burger stay inline for that reason.
- A dozen of the inline SVGs are brand and vendor marks that no icon set carries: Node-RED, MQTT, the certificate and wind turbine icons among them.

One thing worth knowing before that PR is reviewed: the header currently mixes Heroicons releases (it carries two differently formatted copies of the same `sparkles` icon), so converting settles it on one release and a few glyphs pick up the current artwork instead of the older drawing. I matched every icon on path data rather than by eye so the names are not guesses.

dimitrieh:
Reproduced and fixed. The line was anchored to the `<a>` box, and the mega link is a full-width flex row holding an icon and the label, so it stretched across the whole dropdown column.

The top-level nav items already solved this by hanging the underline off the `.ff-nav-label` span that wraps just the text. The mega panel rule was never brought over to that pattern, so it now does the same. Checked first that every mega link carries the label span on both front ends (the 11ty `navoption` shortcode and the Nuxt header both emit it, and no other anchors live inside the panels), so nothing loses its underline in the swap. The stale mega selector in the `prefers-reduced-motion` block is gone too, since the `.ff-nav-label` entry beside it already covered both cases.

Verified on the running Eleventy build against the same item as your screenshot: hovering Renewables now ends the underline with the final "s" instead of running it to the column edge. The deploy preview will show it once it rebuilds.

Yndira-E:
The footer still has the border above all the items
Screenshot 2026-07-28 at 16 32 05

Yndira-E:
Regarding the food icon, what about this one?
Screenshot 2026-07-28 at 16 35 55

You can personalise it as needed and download it from https://icon-sets.iconify.design/?query=food

dimitrieh:
> The footer still has the border above all the items Screenshot 2026-07-28 at 16 32 05

@Yndira-E did you compare with and without, both desktop and mobile? I think it works better currently with the borders with it giving some structure/leading the eyes to grouping the columns. We can adjust, though I am for scoping it out ideally.

Yndira-E:
The background already provides enough separation from the content, at least for the first block it feels redundant.

On mobile, only the first one feels redundant.

Yndira-E:
And the image thing, cover was better, it just needed cropping the image so it matched the proportion and didn't get cut... If you can reverse that, I can take care of the image as a follow up.

Yndira-E:
@dimitrieh on the nuxt version it's flipping the icons on hover, and some items are not properly aligned, please check it out as soon as possible.

Screenshot 2026-07-28 at 19 44 57

dimitrieh:
@Yndira-E fix is up: https://github.com/FlowFuse/website/pull/5413

Both symptoms turned out to be one cause, and it only affects Nuxt. Nuxt UI renders the header's Heroicons as a masked `<span>`, where Eleventy still emits an inline `<svg>`. Nav rules written to target spans, back when no icon could ever match them, now match 45 icons.

- The flipping is `header a > span:not(.ff-nav-label)`, which sets `scale(0.75)`, paired with a hover rule adding `rotate(180deg)`. Both now exclude `.ff-icon`.
- The alignment is a `padding: 8px 16px 16px` landing on the same spans. With `border-box` that floors the box at 32x24, while the inline SVGs beside them are 20x20, so the two sizes sit in the same row.

Measured on the live Nuxt header rather than guessed: the icon span computed `scale(0.75)` and rendered 24x18 against a 20x20 sibling. With the fix applied to that page it renders 20x20, identical to the sibling.

Worth flagging as a pattern rather than a one-off: the mega highlight card already carries its own `transform: none` to defend against that same rule, so this is the second time it has caught something. The rule now has no legitimate target I can find on either front end, so it may be safe to delete outright, but I have left it in place rather than remove something on a hunch.

PR #5374, opened 2026-07-24
Add Shop Floor Communication use-case page

dimitrieh:
## Description

Adds the `/use-cases/shop-floor-communication/` operational use-case page.

**Outcome:** a new entry point for shop-floor messaging and response scenarios, discoverable from the use-case grid and linked from the new navigation.

**Stacking:** base of the feature stack (off `main`). It sits first so the mega-navigation PR that links to this page has a live target to point at. Placeholder review copy from the proof-of-concept has been removed.

### Updated after review

- The related resources section is held back on use case pages. It is commented out in `use-case.njk` rather than deleted, with a note on how to restore it, because with the resources currently tagged it rendered more empty space and "More coming soon" placeholders than actual links. The component is untouched and still works; it needs more tagged resources, not code changes. This affects every use case page, not only the one this PR adds.

## Related Issue(s)

Split from #5371.

## Checklist

 - [x] I have read the contribution guidelines
 - [x] I have considered the performance impact of these changes
---

**Stack** (everything below #5375 is independent and can merge in any order, except #5378 which needs #5376):

```mermaid
flowchart LR
    main([main])
    main --> P74["#5374 shop-floor"] --> P75["#5375 nav"]
    P75 --> P76["#5376 use-cases"] --> P78["#5378 industry + AI + homepage"]
    P75 --> P81["#5381 content polish"]
    P75 --> P87["#5387 punctuation"]
    P75 --> P88["#5388 logo wave"]
    classDef here stroke-width:4px;
    class P74 here;
```

Changes to the stack since it was opened:

- `#5373` (punctuation) was independent, off `main`, and is merged.
- `#5377` (hero video assets) is **closed**. The video heroes were dropped after review, so the assets are not needed.
- `#5379` (AI page) and `#5380` (homepage copy) were merged into the stack rather than into `main`, so their content now travels in `#5378`.

PR #5376, opened 2026-07-24
Add use-case and outcome lens pages and cross-link includes

dimitrieh:
## Description

**Outcome:** visitors can navigate the platform by use case and by outcome. Adds use-case chips and links, a derived use-case grid for industry pages, the use-case layout, refreshed use-case pages, and systematic cross-links between related content. The industry-use-cases heading falls back to the page eyebrow so industry pages read cleanly rather than showing the raw SEO title.

**Builds on:** the navigation PR's shared stylesheet and components.

**Provides for later PRs:** the reusable includes (`use-case-links`, `use-case-chips`, `industry-use-cases`) that the industry pages and the AI page consume downstream.

### Updated after review

- **Text links underline on hover again.** The related resource titles and the blog "See all" link had `hover:no-underline`, which cancels the site-wide rule and left them with a colour change only. The titles drop the class; "See all" underlines its label span so the trailing arrow stays out of the underline. Card links are unchanged and still have no underline, since the card hover carries that interaction.
- The related resources section itself is held back in #5374 until there is enough tagged content, so the change above is not visible yet on use case pages.

## Related Issue(s)

Split from #5371.

## Checklist

---

```mermaid
flowchart LR
    main([main])
    class P76 here;
```


src/_includes/explore-more-content.njk
```diff
@@ -22,9 +22,9 @@
             </a>
             {%- endfor -%}
         </div>
-        <a href="/blog/{{ collectionName or '' }}" class="w-full text-right flex flex-row items-center justify-end self-end justify-self-end gap-1">
-            See all
-            {% include "components/icons/arrow-long-right.svg" %}
+        <a href="/blog/{{ collectionName or '' }}" class="group hover:no-underline w-full text-right flex flex-row items-center justify-end self-end justify-self-end gap-1">
+            <span class="ff-animated-link">See all</span>
```
Yndira-E: Introducing it here creates an inconsistency with the rest of links across the website. I would reserve this for the main nav exclusively.

Yndira-E:
What do you think about adding a CTA that links to /ai, so they can get started with the expert?

Screenshot 2026-07-24 at 17 09 52

src/_includes/explore-more-content.njk
Yndira-E: I can see you've been adding that class to other elements as well, lets reserve this class for the main nav only.

Yndira-E: For instance, here, the hover effect on the card is enough, no need for that effect there
Screenshot 2026-07-24 at 17 20 09

Yndira-E:
This button shouldn't be full width at that screen size
Screenshot 2026-07-24 at 17 18 45

dimitrieh:
**FYI:** Rebased onto the updated split/nav, and onto current main underneath (main had moved 14 commits since Friday). Deploy preview is rebuilding, the link above will update.

dimitrieh:
**Review feedback addressed.**

- The animated underline is removed from `explore-more-content.njk`, and from the use-case index, related-resources, industry use-cases and use-case-links includes. It is now reserved for the main navigation. On the cards the existing hover state carries the interaction, as you suggested.
- The Device Agent button no longer stretches the full width of the edge-connectivity section. The button class is a flex container, so it filled its block parent; it is now wrapped the way the other page CTAs are, and measures 212px in a 1024px parent instead of spanning it. Verified locally.
- Added the CTA through to the AI page from the build-it-with-AI section, so readers can go straight to the Expert from the step cards.

Yndira-E (approved):

PR #5377, opened 2026-07-24
Add hero video backgrounds and posters

dimitrieh:
## Description

Adds the hero video background assets (9 clips + matching poster stills) used by the industry pages and the AI page.

**Outcome:** brand-quality motion visuals available to the site's hero sections.

**Stacking:** isolated in its own PR so the media reviews on its own and any clip can be swapped without touching page structure. It sits beneath the pages that reference it (industry, AI), so those pages resolve their assets.

**Provides for later PRs:** the clips and posters the industry and AI PRs wire into their layouts.

All clips are from Pexels and are free for commercial use, no attribution required (https://www.pexels.com/license/). Sources:

| File | Source |
|------|--------|
| industries/hero-video/automotive.mp4 | https://www.pexels.com/video/32386532/ |
| industries/hero-video/aerospace-components.mp4 | https://www.pexels.com/video/19791092/ |
| industries/hero-video/aviation-aerospace.mp4 | https://www.pexels.com/video/36971603/ |
| industries/hero-video/electronics-appliances.mp4 | https://www.pexels.com/video/6755170/ |
| industries/hero-video/food-beverage.mp4 | https://www.pexels.com/video/8633311/ |
| industries/hero-video/life-sciences.mp4 | https://www.pexels.com/video/8863359/ |
| industries/hero-video/renewables.mp4 | https://www.pexels.com/video/32939545/ |
| industries/hero-video/semiconductors.mp4 | https://www.pexels.com/video/32386615/ |
| ai/ai-hero-bg.mp4 | https://www.pexels.com/video/32750415/ |

Each clip is under 1 MB; posters are the matching stills.

## Related Issue(s)

Split from #5371.

## Checklist

 - [x] I have read the contribution guidelines
 - [x] I have considered the performance impact of these changes
---

**Stack** (left to right is merge order; #5373 punctuation is independent, off `main`):

```mermaid
flowchart LR
    main([main])
    main --> P73["#5373 punctuation"]
    main --> P74["#5374 shop-floor"] --> P75["#5375 nav"] --> P76["#5376 use-cases"] --> P77["#5377 hero-media"] --> P78["#5378 industry"] --> P79["#5379 AI"] --> P80["#5380 homepage"] --> P81["#5381 polish"]
    classDef here stroke-width:4px;
    class P77 here;
```

PR #5378, opened 2026-07-24
Rework industry page heroes, rebuild the AI page, refresh homepage copy

dimitrieh:
## Description

**Outcome:** industry pages lead with a contextual eyebrow and their own product or testimonial visual, keep their customer logo strip directly below the hero with per-page copy, and gain a derived grid of relevant use cases. The AI page is rebuilt around sticky capability sections and shows FlowFuse Expert in action before it explains it. The homepage leads with the edge-native message.

Originally this PR **added** animated video heroes. Those were dropped after review: the clips were redundant against the contextual visual already in each hero, largely hidden behind the semi-transparent readability card, low quality, and visibly jumping on loop rather than looping seamlessly. The video markup, front matter, and sizing scripts are removed here, and `#5377` (the asset PR) is closed.

What this PR now contains:

- industry layout and automotive page: eyebrow with icon, no video hero, customer logo strip restored to its original position with its per-page copy rather than a relocated block with hardcoded copy
- derived use-case grid per industry, consuming the includes from `#5376`
- AI page rebuilt with sticky capability sections, and the FlowFuse Expert screenshot returned directly under the hero, which also closes the empty white band that separated the hero from the page
- homepage tagline and hero copy, in title case, with the manual line break in the heading removed
- a dev image-handler fix that disambiguates output by source path, so two source images sharing a basename no longer overwrite each other

**Builds on:** the use-case includes from `#5376` (the derived grid) and the navigation PR's shared styles.

### Updated after review

- **Hero CTA markup restored.** The arrow wrapper span is reverted: `arrow-right.svg` already carries `w-6 h-6` on its root, so it sizes itself and the wrapper shrank it from 24px to 20px rather than fixing anything. Both hero CTAs are back to the original markup.
- **Problem marker sizes restored** to a 48px circle with a 24px mark.
- **Homepage AI text links underline on hover again.** The "Learn more about FlowFuse AI" and "Book a demo" links underline their label spans, so the trailing arrow stays clear.
- **Punctuation partly split out.** `src/landing/plc.njk` was purely punctuation and moved to #5387. The industry markdown files here still carry their em dash replacements, because each one interleaves them with the hero `eyebrow` and `eyebrowIcon` front matter this PR adds; reverting those files takes the eyebrow with them. #5373 missed these files, so the replacements are not duplicates of it.
- Still open: the marker alignment on the problem lists. The `li` carries both `items-start` and `items-center`, so the winner depends on stylesheet order rather than the attribute. Left alone pending a decision on whether the marker should align to the first line or centre on the paragraph.

## Related Issue(s)

Split from #5371. Absorbs #5379 (AI page) and #5380 (homepage copy), both merged into this stack.

## Checklist

---

```mermaid
flowchart LR
    main([main])
    class P78 here;
```


src/industries/automotive.njk
```diff
@@ -161,12 +185,9 @@ cta:
                 {% endif %}
             </div>
         </div>
-        <div class="max-w-screen-lg mx-auto mt-16 text-center">
-            <p class="text-gray-500 text-lg font-semibold">{{ hero.trustedByText }}</p>
-            <div class="mx-auto text-center -mt-0.5 -mb-10">
-                {% include "social-proof.njk" %}
-            </div>
```
Yndira-E, line 168: @dimitrieh this section needs to be where it is, it was one of @JoycePlaysFootball's request that the logos where right below the hero section, and also, we had a more meaningful copy above the icons than the one proposed in this PR.

Yndira-E:
I like the eyebrow above the title, it helps provide more context.

The videos… I’m not sure we need them. The current layouts already have visuals that support the message, so adding background videos feels redundant, and I don’t think they add enough value in their current state:

* The content already sits alongside a contextual visual, either a product image or a testimonial (only automotive), so introducing a full-width background video means we now have two competing visuals mostly telling the same story.

Screenshot 2026-07-24 at 17 32 30

* Since the content sits inside a large semi-transparent card for readability, most of the video ends up obscured, leaving only the edges visible. At that point it feels more like decoration than content that supports the message.
* The quality isn’t good enough; they look blurry/pixelated. I suspect that’s the tradeoff of compressing them enough to keep their size reasonable, which makes me question whether they’re adding enough value to justify introducing them in the first place.
* They’re not seamless loops, so there’s a noticeable jump when they restart, which makes the experience feel less polished.
* I’m not convinced the additional motion adds enough value to justify the extra visual complexity.

Regarding the animated hover effect for the links, same feedback as in the previous PRs: I’d keep it only for the main navigation.

dimitrieh:
**FYI:** Rebased onto the updated split/hero-media, and onto current main underneath (main had moved 14 commits since Friday). Deploy preview is rebuilding, the link above will update.

This branch now also carries the AI page (#5379) and the homepage copy (#5380), both of which were merged into the stack rather than into main.

dimitrieh:
**Review feedback addressed.** The title and description of this PR are updated, because what it does has changed materially.

- **Video heroes removed.** All five of your objections are accepted rather than worked around: redundant against the contextual visual already in each hero, largely hidden behind the readability card, low quality, not seamless on loop, and the added motion not worth the complexity. `#5377`, which carried the assets, is closed.
- **Logo strip restored.** You were right and I had misread it. On the automotive page the strip sat inside the hero, directly below the hero grid and above the metrics, using the per-page `hero.trustedByText` copy. This PR had moved it near the bottom with hardcoded "Trusted by our customers". It is back where it was with its own copy, and the layout-driven industry pages get their per-page `socialProofText` back too.
- **The eyebrow stays**, since you liked it for the added context.
- Animated underline removed from the hero secondary CTAs here, per the nav-only decision.

Carried in from the two PRs that merged into this stack:

- `#5379` (AI page): the hero video is gone, and the FlowFuse Expert screenshot is back directly under the hero. That also closes the abrupt hero-to-section transition and the large white gap you flagged, and answers the missing product imagery: the page now shows the product before it explains it. The sticky capability sections you liked are unchanged, and the three copy suggestions from that review are applied.
- `#5380` (homepage): the tagline is in title case as suggested, and the manual line break in the heading is removed.

Note the deploy preview is queued behind a large Netlify backlog, so the link above may still show Friday's build, videos included, for a while.

src/_includes/layouts/industry.njk
```diff
@@ -52,13 +59,10 @@ layout: default
                         <a class="ff-btn flex items-center gap-2 text-base uppercase text-indigo-600 hover:text-indigo-800"
                             href="{{ hero.secondaryCtaLink }}"
                             onclick="capture('cta-contact', {'position': 'hero-mobile', 'page': '{{ slug }}'})">
-                            <span class="text-base uppercase items-center flex gap-2 uppercase items-center text-indigo-600 hover:text-indigo-800">
-                                {{ hero.secondaryCta }}
-                                {% include "components/icons/arrow-right.svg" %}
-                            </span>
+                            <span>{{ hero.secondaryCta }}</span>
+                            <span class="w-5 h-5 shrink-0 flex items-center [&>svg]:w-full [&>svg]:h-full">{% include "components/icons/arrow-right.svg" %}</span>
```
Yndira-E: @dimitrieh I'm not sure this should be part of this pr?

src/_includes/layouts/industry.njk
```diff
@@ -75,8 +79,8 @@ layout: default
                     <ul class="mt-6 flex flex-col gap-6 list-none p-0 m-0">
                         {% for problem in problems %}
                         <li class="flex items-start gap-3 items-center">
-                            <span class="flex-shrink-0 w-12 h-12 bg-red-300 rounded-full flex items-center justify-center text-white">
-                                <span class="w-6 h-6">{% include "components/icons/x-mark.svg" %}</span>
+                            <span class="flex-shrink-0 w-8 h-8 bg-red-300 rounded-full flex items-center justify-center text-white">
+                                <span class="w-4 h-4">{% include "components/icons/x-mark.svg" %}</span>
                             </span>
```
Yndira-E: @dimitrieh why change the size of the icons?

src/industries/aerospace-components.md
```diff
@@ -33,32 +35,32 @@ outcomes:
   items:
     - svgPath: "components/icons/bolt.svg"
       title: "Cut FAI cycle time"
-      description: "by assembling dimensional results, process parameters, and operator sign-offs continuously during production — submission-ready without manual reconstruction."
+      description: "by assembling dimensional results, process parameters, and operator sign-offs continuously during production, submission-ready without manual reconstruction."
     - svgPath: "components/icons/clipboard-document-check.svg"
       title: "Eliminate traveler revision escapes"
-      description: "with digital travelers always issued from a single controlled source — the current approved revision, every time."
+      description: "with digital travelers always issued from a single controlled source, the current approved revision, every time."
```
Yndira-E, line 41: @dimitrieh weren't the em dashes replaced in a different pr that's already merged?

src/_includes/layouts/industry.njk
Yndira-E: They look misaligned too... I'd keep them as they were...

Screenshot 2026-07-27 at 19 17 18

src/_includes/layouts/industry.njk
```diff
@@ -26,22 +35,20 @@ layout: default
                             <a class="ff-btn flex items-center gap-2 text-base uppercase text-indigo-600 hover:text-indigo-800"
                                 href="{{ hero.secondaryCtaLink }}"
                                 onclick="capture('cta-contact', {'position': 'hero', 'page': '{{ slug }}'})">
-                                <span class="text-base uppercase items-center flex gap-2 uppercase items-center text-indigo-600 hover:text-indigo-800">
-                                {{ hero.secondaryCta }}
-                                {% include "components/icons/arrow-right.svg" %}
-                                </span>
+                                <span>{{ hero.secondaryCta }}</span>
+                                <span class="w-5 h-5 shrink-0 flex items-center [&>svg]:w-full [&>svg]:h-full">{% include "components/icons/arrow-right.svg" %}</span>
```
Yndira-E: I'm not sure we need those new clases, it was rendering as expected without these changes

Yndira-E:
@dimitrieh I thinks this pr needs some clean up too, to keep it focused on what we actually want to update, there are many files fixing/replacing em dashes that I believe were already addressed in this pr: https://github.com/FlowFuse/website/pull/5373

src/_includes/layouts/industry.njk
dimitrieh: You are right. Restored the original markup here and on the mobile copy of the same CTA below.

`arrow-right.svg` already carries `class="w-6 h-6"` on its root element, so it sizes itself and never needed a wrapper. My `w-5 h-5` span with `[&>svg]:w-full` actually shrank it from 24px to 20px, so it was a size change dressed up as a fix. Including the icon bare is also the existing convention in `abm-landing.njk` and `index.njk`. Verified on the rebuilt page: the arrow is back to 24x24 and rendering from its own class.

One thing I noticed while reverting and deliberately did not change: the original inner span repeats classes already present on the anchor (`text-base uppercase`, `flex gap-2`, `text-indigo-600 hover:text-indigo-800`) and carries `uppercase` and `items-center` twice over. Harmless, and happy to tidy it separately if you want, but not in this PR.

src/_includes/layouts/industry.njk
dimitrieh: Same as the other thread on this file: reverted. This is the mobile copy of the same hero CTA, so both are back to the original markup.

src/_includes/layouts/industry.njk
dimitrieh: Reverted to the 48px circle with the 24px mark. Verified on the rebuilt page: 48x48 and 24x24.

On the misalignment, I think I found the cause, and it predates this PR. The `li` carries both `items-start` and `items-center`. Two `align-items` utilities on the same element, and which one wins is decided by the order Tailwind emits them in the stylesheet, not the order they appear in the attribute. It currently resolves to `flex-start`, so the marker pins to the very top of what is a five line paragraph on most of these pages.

I have left that alone rather than guess, because it is a visual call. Which would you prefer: the marker aligned to the first line of text, or centred against the whole paragraph? I will remove the duplicate and set it explicitly either way.

src/industries/aerospace-components.md
dimitrieh: Checked this properly. The replacements here are not duplicates of #5373, though you are right that they do not belong in this PR.

#5373 merged on 24 July across 100 files, but it missed these. Main still carries 9 em dashes in this file, 38 in `src/landing/plc.njk` and 3 in `src/platform/why-flowfuse.njk`.

I tested every file in the PR by comparing its alphanumeric content stream, so punctuation changes are invisible to the comparison and word changes are not. Exactly one file was purely punctuation: `src/landing/plc.njk`. That is now split out as #5387.

These six industry markdown files cannot be split the same way. Each one interleaves the punctuation with the hero `eyebrow` and `eyebrowIcon` front matter that this PR adds. I tried reverting them wholesale and it took the eyebrow with it, so I put that back. If you would rather they were strictly separated, I can move the punctuation lines individually into #5387; it is just more surgical than a file revert.

dimitrieh:
**Today's review feedback addressed.** Replies are in the individual threads, summary here.

| Item | Status |
|---|---|
| Arrow wrapper classes not needed | **Reverted.** You were right: `arrow-right.svg` already carries `w-6 h-6` on its root, so my wrapper shrank it from 24px to 20px rather than fixing anything. Both hero CTAs are back to the original markup. |
| Why change the icon sizes | **Reverted** to the 48px circle with the 24px mark. |
| Icons look misaligned | **Cause found, waiting on you.** The `li` carries both `items-start` and `items-center`, so the winner is decided by stylesheet order rather than the attribute. It resolves to `flex-start`, pinning the marker to the top of a five line paragraph. Details and a question in the thread. |
| Em dashes already done in #5373 | **Partly.** #5373 missed these files, main still has them, so the replacements are not duplicates. But they are in the wrong PR: `plc.njk` was purely punctuation and is now #5387. The six industry markdown files interleave punctuation with the eyebrow front matter this PR adds, so they cannot be split by reverting the file. Detail in the thread. |
| PR needs clean up to stay focused | **Partly done.** The punctuation split above is the piece that could be cleanly separated. |

On the wider cleanup point: I agree with the principle and it applied to the stack more broadly than this PR. The logo hover effect has also moved out of #5375 into its own PR at the top of the stack, so it can be declined on its own rather than riding along with the navigation rebuild.

Verified locally after the reverts: problem markers 48x48 with a 24x24 mark, hero secondary arrow back to 24x24 from its own class, and the eyebrow and use case section unaffected.

Yndira-E:
@dimitrieh can you reverse the changes that affect the FAQs? On hover they look like a link, but they're not.
Screenshot 2026-07-28 at 17 07 42

The homepage, copy-wise, I'll leave that to ZJ. If we keep this version, I'd like it to be centered in the available vertical space. Since it's now a two line sentence on large screens, it looks misaligned:

Screenshot 2026-07-28 at 17 06 33

When it breaks into three lines it does look center.
Screenshot 2026-07-28 at 17 12 59

Yndira-E (approved):
I don't want to be a blocker, most of it is ok, and I'll defer the new tagline review to @ZJvandeWeg 

If we keep this version there are some adjustments to implement to the homepage hero section to better align it. Also the FAQs style should remain as they currently are on main.

If it's necessary to move forward, these last two points can be addressed in a follow up PR.

PR #5379, opened 2026-07-24
Rebuild AI page with sticky capability sections

dimitrieh:
## Description

**Outcome:** the AI page presents its capabilities as sticky-header scrolling sections over a video hero, a clearer and more engaging AI story than the previous static layout.

**Builds on:** the AI hero clip from the hero-media PR and the use-case link cards from the use-case PR (the featured production-monitoring and shop-floor cards), plus the navigation PR's shared styles.

## Related Issue(s)

Split from #5371.

## Checklist

 - [x] I have read the contribution guidelines
 - [x] I have considered the performance impact of these changes
---

**Stack** (left to right is merge order; #5373 punctuation is independent, off `main`):

```mermaid
flowchart LR
    main([main])
    main --> P73["#5373 punctuation"]
    main --> P74["#5374 shop-floor"] --> P75["#5375 nav"] --> P76["#5376 use-cases"] --> P77["#5377 hero-media"] --> P78["#5378 industry"] --> P79["#5379 AI"] --> P80["#5380 homepage"] --> P81["#5381 polish"]
    classDef here stroke-width:4px;
    class P79 here;
```

PR #5381, opened 2026-07-24
Polish content page links, buttons, and copy

dimitrieh:
## Description

**Outcome:** link and button interactions on the remaining content pages follow the site conventions, and their labels underline on hover like any other body link.

Originally this PR **applied** the animated underline across the site. Review reserved that effect for the main navigation, so it is gone from here. Removing it initially took the plain underline with it, which left several links with a colour change and nothing else. That is corrected: labels underline on hover again, while the sweeping motion stays navigation-only.

What this PR contains:

- link and button labels wrapped in their own span, so a hover underline hugs the text instead of dragging the adjacent icon or chevron into it
- the back links on the post and story layouts restructured, with an invalid `align-center` class corrected to `items-center`, and their labels underlining on hover
- a chevron given `shrink-0` in the FAQ include so it cannot be squashed
- button conventions applied to the comparison pages and the why-FlowFuse CTA
- punctuation normalized in the why-FlowFuse and comparison page copy
- a social image added to the customer stories index

The FAQ header and footer link underlines are restored in #5375, since that is where they were over-removed. The PLC landing page punctuation pass moved to #5387 so this PR stays on interactions. Underline work for pages introduced earlier in the stack lives in those PRs.

**Builds on:** the whole stack, this is the finishing layer applied last.

```mermaid
flowchart LR
    main([main])
    class P81 here;
```

## Related Issue(s)

Split from #5371.


Yndira-E (commented):
I still have the same concern I raised in the previous PRs about the animated underline treatment.

My impression is that it’s being applied quite selectively rather than consistently, so it doesn’t yet feel like a cohesive interaction pattern. Some links use it, others don’t, and in places like the FAQ it adds motion to interactions that I don’t think benefit from it.

I’d still recommend keeping this effect reserved for the main navigation. Used more sparingly it feels like a deliberate interaction pattern, whereas applying it across many different templates starts to dilute its impact and adds more motion throughout the site than I think is necessary.

dimitrieh:
**FYI:** Rebased onto the updated split/industry. Two things changed here that are worth knowing.

The diff is now 10 files instead of 13, and the conflict is gone. #5379 was squash merged into split/industry, so this branch was carrying a second copy of the AI page and homepage changes and conflicting on `src/ai.njk`. Those two commits are dropped now, so what is left is only the actual interaction polish change.

The three `src/ai.njk` copy fixes from the #5379 review (`Built-In` in the meta title and hero title, and "build industrial applications" in the meta description) were still at their pre-review wording on this branch and would have been reverted on merge. They are restored.

dimitrieh:
**Review feedback addressed.** The title and description are updated, because this PR now does the opposite of what it originally did.

Your point is taken in full: the underline is reserved for the main navigation. Removed from the FAQ headers, the post, story and ABM landing layouts, the comparison pages, and the why-FlowFuse CTA. FAQ headers keep their colour shift on hover and focus, since it was the motion you objected to rather than the feedback itself.

Two things worth knowing about this PR specifically:

- The diff shrank from 13 files to 10 and the conflict is gone. `#5379` was squash merged into `split/industry`, so this branch was carrying a second copy of the AI page and homepage changes and conflicting on `src/ai.njk`. Those commits are dropped, so what is left is only the interaction work this PR owns.
- The three `src/ai.njk` copy fixes from the `#5379` review were still at their pre-review wording on this branch and would have been reverted on merge. They are restored.

On the follow-up ZJ asked for, standardising CTA links as a component when the page is ported to Vue: happy to leave that to a separate PR rather than fold it in here, so this one stays a removal.

KristopherLeads (approved):

Yndira-E:
Hey @dimitrieh, thanks for the changes.

I’m seeing quite a few files with changes that don’t seem to be related to the animated underline. Could you please clean them up so the PR only includes the relevant changes?

Yndira-E:
@dimitrieh I'm a bit confused about this PR, why not reverse all of those files to their state on main? and the commas and dashes can be an unrelated pr to the stacked ones here...

The style you were implementing now only applies to the main nav, so reversing the rest should be fine, or am I missing something?

PR #5388, opened 2026-07-27
Add the logo hover wave to the header and footer

dimitrieh:
This just adds a simple logo animation in alignment with: https://flowfuse.slack.com/archives/C02TG4WH5PC/p1785262574563559?thread_ts=1784793026.150549&cid=C02TG4WH5PC

> For what it's worth, I quite like the singular instance colour flow across the logo. If it's limited to first time only, it reads as a surprise and delight opportunity rather than noise

The logos link back to the homepage. This gives them a fun interaction style to indicate they are clickable beyond just the cursor

CleanShot 2026-07-28 at 10 48 30

CleanShot 2026-07-28 at 10 48 19


KristopherLeads (approved):

Yndira-E (approved):
Not entirely convinced here, the blue to red gradient is not part of FlowFuse identity, we've only used it specifically for the expert, which I think is where it makes sense.

I'll merge and iterate on it when I can.

PR #5390, opened 2026-07-28
Fix docs card grid, table of contents, search and contrast on the Nuxt docs

dimitrieh:
## Description

Stacked on https://github.com/FlowFuse/website/pull/5177, which migrates the docs to Nuxt. This closes out the four items left open on that review, so it targets `zj-docs-deploy-nuxt` rather than `main`.

**Card grid rendered as escaped code blocks.** CommonMark ends a raw HTML block at the first blank line. The upstream docs have stray blank lines inside the tile grids, and because the markup is indented, everything after each one was reparsed as an indented code block, which is why fragments like `</div>` and `<label>` appeared as dark boxes overflowing the content column. `markdown-it` tolerated this; the Nuxt content parser does not. The markdown transforms move into a plain-JS sibling module so they can be unit tested, and two are added: one removing blank lines inside top-level raw HTML blocks, one stripping the inline scripts that define custom elements the parser never executes. Ordering matters, since callout conversion deliberately inserts blank lines, and a test pins that. An `IconChevronRight` component makes the chevron tag resolve the same way `ChecklistItem` already does.

**Table of contents never rendered.** The docs page passed no `links` prop to `HandbookToc`, which only renders when it has links, so the right column was always empty.

**No search on docs pages.** The Algolia autocomplete is extracted into a component taking a filter and placeholder, leaving `HandbookSearch` as a wrapper so handbook behaviour is unchanged. The indexer needs no change: it scans the built output and derives the category from the first path segment.

**Inline code contrast.** The color mode module ships with the UI library and defaults to following the operating system, so visitors preferring dark got the dark class on the document and the UI colour tokens flipped. Hand-styled page chrome stayed light while token-styled elements like inline code became near-black pills with white text. Tokens are locked to light until the site has a dark theme. Code blocks also now scroll rather than clip, and the critical callout variant is styled (emitted by the docs module, never defined in the stylesheet).

**Clone resilience.** The build-time clone is retried with backoff, so a transient network failure no longer fails the whole production deploy. Both git calls now take argument arrays instead of shell strings, since one interpolated a filename taken from the cloned repo.

Left deliberately out of scope: the broader visual consistency sweep, which was already offered on the base PR.

## Related Issue(s)

Stacked on https://github.com/FlowFuse/website/pull/5177.

## Checklist


dimitrieh:
@Yndira-E @ZJvandeWeg this merges into https://github.com/FlowFuse/website/pull/5177 as it fixes a subscope as mentioned in https://github.com/FlowFuse/website/pull/5177#issuecomment-5070047546

Fixing CI in a bit.

Yndira-E (approved):

PR #5395, opened 2026-07-28
Show the FlowFuse AI and MCP nodes on /integrations

dimitrieh:
## Description

The `/integrations` page reads three feeds: the Node-RED library mirror at `ff-integrations.flowfuse.cloud/api/nodes`, plus the Hub and Edge certified catalogues. The FlowFuse AI nodes and the MCP server nodes are in none of them, because both packages are published only to the private registry and so never reach the public library.

A fourth catalogue does list them, `ff-certified-nodes.flowfuse.cloud/ff-catalogue.json` ("FlowFuse Nodes", 13 modules). This adds it as a source and filters it down: eleven of its modules are `@flowfuse` packages the library API already returns with download counts and author data, so those richer entries win and only modules missing from both other sources are taken from the new feed. Today that is `@flowfuse-nodes/nr-ai-nodes` and `@flowfuse-nodes/nr-mcp-server-nodes`.

They are classified as `recommended`, which sorts them into the block directly after the certified entries and gives them a generated detail page. Both carry a `flowfuse.com` docs URL in the catalogue, so the existing docs-path handling links them to `/node-red/flowfuse/ai/` and `/node-red/flowfuse/mcp/` without an override.

The two shared helpers were generalised to take the tier and an optional collection, since the fetch and normalise logic is identical across all three catalogue feeds. A feed that fails still degrades to an empty list rather than taking the page down.

## Related Issue(s)

Closes FlowFuse/engineering#212

## Checklist

### DIFF
--- a/nuxt/types/integrations.ts
+++ b/nuxt/types/integrations.ts
@@ -109,3 +109,11 @@ export const INTEGRATION_CATEGORIES: Record<string, string> = {
 export const INTEGRATIONS_API = 'https://ff-integrations.flowfuse.cloud/api/nodes'
 export const CERTIFIED_HUB_API = 'https://ff-certified-nodes.flowfuse.cloud/ff-it.json'
 export const CERTIFIED_EDGE_API = 'https://ff-certified-nodes.flowfuse.cloud/ff-ot.json'
+
+/*
+    The FlowFuse-published node packages, which are a different set from the Hub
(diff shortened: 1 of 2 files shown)


dimitrieh:
CleanShot 2026-07-28 at 15 32 16@2x

n-lark (approved):

dimitrieh:
enabling auto-merge, thanks @n-lark

Issue #4264, opened 2025-12-19
Website accessibility issues


knolleary:
Reported by the community, our website scores poorly for accessibility according to https://www.accessibilitychecker.org/audit/?website=https://flowfuse.com/blog/2025/12/flowfuse-release-2-25/&flag=uk#

I suspect there are some quick wins to be had to improve it.

allthedoll:
cc @n-lark for 👀

n-lark:
## Accessibility Audit Scope - #4264

> **TL;DR:** 8 issues found, 24 total instances. 4 are high priority (keyboard access, focus visibility, skip nav, modal ARIA). Mostly small fixes — estimated ~1-2 days total. No issues with images, form labels, or lang attribute.

cc @allthedoll @knolleary 

Audited against WCAG 2.1 AA. Nick's header anchor fix (#4266) is already merged. Here's what remains:

### Issues & Effort

| # | Issue | Why | Files Affected | Effort | Priority |
|---|-------|-----|---------------|--------|----------|
| 1 | **No "Skip to main content" link** | WCAG 2.4.1 - Keyboard users must tab through entire nav on every page load | `_includes/layouts/base.njk` | S | ✅ See https://github.com/FlowFuse/website/pull/4883 |
| 2 | **Interactive divs not keyboard-accessible** - file drop zone, pagination `<li>` elements, dialog overlay all use `onclick` without `tabindex`, `role`, or key handlers | WCAG 2.1.1 - All functionality must be operable via keyboard; screen readers can't announce non-semantic elements | `_includes/migration.njk`, `integrations/index.njk`, `_includes/feature_lists/features-table-base.njk` | M | High |
| 3 | **Modal missing ARIA attrs** - no `role="dialog"`, `aria-modal`, `aria-labelledby`, no focus trap, no Escape key close | WCAG 4.1.2 - Screen readers don't announce it as a dialog; focus can escape behind the modal | `_includes/components/ai-expert-modal.njk`, `js/ai-expert-modal.js` | M | High |
| 4 | **Focus outlines removed without alternatives** - `outline: none` on search, range input, textareas with no visible replacement | WCAG 2.4.7 - Keyboard users lose track of which element is focused | `css/algolia-theme.css`, `css/style.css`, `_includes/components/ai-chat-interface.njk` | S | ✅ See https://github.com/FlowFuse/website/pull/4883 |
[...]

PR #5402, opened 2026-07-28
Stop the FAQ header looking like a link on hover

dimitrieh:
## Description

FAQ headers are buttons that expand an answer in place, but hover added an underline, so they read as links. Underline removed, colour shift kept.

## Related Issue(s)

Review feedback on #5378.

## Checklist

### DIFF
--- a/src/css/style.css
+++ b/src/css/style.css
@@ -3004,18 +3004,17 @@ lite-youtube::before {
     }
 }
 
-/* FAQ / accordion header interaction: colour shift plus a plain underline on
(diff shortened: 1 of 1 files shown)


Yndira-E (approved):

dimitrieh:
@Yndira-E the original change was made so we don't only rely on a small color change for interaction styling. Still think this is necessary if we merge this PR. We should do a solid WCAG review of our website.

Yndira-E:
@n-lark has done extensive [work on WCAG](https://github.com/FlowFuse/website/issues/4264) already

PR #5405, opened 2026-07-28
Give Food & Beverage its own nav icon

dimitrieh:
## Description

Replaces `snowflake` (cold chain) with `fast-food-outline` from Ionicons, the icon suggested in review. MIT licensed. Applied to both front ends.

Kept at its native 512 viewBox: Ionicons draws it at stroke-width 32, which is 1/16 of the viewBox, the same ratio Heroicons gets from 1.5 on 24, so it matches its neighbours without adjustment.

## Related Issue(s)

Review feedback on #5375.

## Checklist


dimitrieh:
This needs a follow up, one sec

dimitrieh:
@Yndira-E updated: this now uses the exact icon you suggested, `fast-food-outline` from Ionicons, pulled from the set rather than redrawn. MIT licensed.

The first push had a hand-drawn fork and glass instead, because the shape wasn't identifiable from the screenshot alone. That stand-in is gone.

It needed no adaptation to fit the set: Ionicons draws it at `stroke-width="32"` on a 512 viewBox, which is 1/16, the same ratio Heroicons gets from `1.5` on 24. So it sits at the same weight as the icons either side of it. Checked in both front ends, since 11ty nests the file inside the wrapper the shortcode emits and Nuxt inlines it directly.

One thing worth your eye on the preview: the other marks in that panel are plant and equipment metaphors (wind turbine, factory, building), and a burger with a soda cup reads more quick-service restaurant than food manufacturing. Happy either way, it's your call, but flagging it now rather than after it ships.

Yndira-E (approved):

PR #5412, opened 2026-07-28
Reduce duplicate CI/Netlify builds per PR push

ZJvandeWeg:
## Summary
- `test.yml` now cancels in-progress runs for the same ref (matching `sast-scan.yml`), so pushing multiple commits to a PR no longer queues redundant full builds that just get superseded.
- `netlify.toml` skips the `branch-deploy` build context, since deploy previews already build every PR commit — branch deploys were building the same commit a second time.

## Test plan
- [ ] Push a couple of quick commits to this PR branch and confirm only the latest "Tests" run executes (earlier ones show as cancelled).
- [ ] Confirm Netlify shows only a deploy-preview build for this PR's commits, no branch-deploy build.


dimitrieh:
Both changes do what the description says, and the Netlify half is correctly scoped: `ignore` under `[context.branch-deploy]` does not cascade into `[context.deploy-preview]`, and Netlify treats exit 0 as "skip", so this kills only branch-deploy builds. One thing I'd want changed before this goes in.

**The concurrency block also cancels runs on `main`.** `test.yml` is `on: push` with no branch filter, so the group resolves to `Tests-refs/heads/main` for merge commits too. `main` has no branch protection, so "Tests" is not a required check, which makes the post-merge run the only thing that ever builds the site and runs the hyperlink check with `--check-anchors` over `nuxt/dist`. Yesterday five PRs merged inside six minutes. Under this config each merge cancels the previous main run, so most of those commits would never get link-checked, and any eventual failure attaches to whichever commit happened to survive.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```

That keeps the intended win, deduping pushes to PR branches, which is where the queue cost actually comes from.

Related: the "matching sast-scan.yml" framing in the commit message isn't quite equivalent. `sast-scan.yml` filters `push` to `branches: [main]` and picks up PR coverage from a separate `pull_request` trigger, so its group keys differ from an unfiltered `on: push`.

Two smaller notes:

* I'd keep `command = "npm run build:nuxt"` alongside the new `ignore` in the branch-deploy block. If anyone later makes the ignore conditional, branch deploys silently fall back to the root `[build]` command, which also runs `index:algolia` and would write the search index from a branch.
* Branches with no open PR lose previews entirely, since deploy-preview only fires for PRs. Worth confirming nobody relies on pushing a branch and sharing its URL before opening the PR.

Worth being explicit in scope: this doesn't touch the two things that actually hurt, the ~100 minute builds when the image cache misses, and the 3-build concurrency cap. Roughly halving builds for branches that have a PR is still a real win, but queue depth on a stacked set remains.

Also, both test-plan boxes are still unticked. The Netlify one matters here, because a misconfigured `ignore` fails silently by producing no preview at all.

dimitrieh:
TL;DR: ship it, with one change.

* **Blocking:** `cancel-in-progress: true` also cancels merge runs on `main`, since `on: push` has no branch filter. That run is the only full build and link check we get, because `main` has no required checks. Make it `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}`.
* **Nits:** keep `command = "npm run build:nuxt"` next to the new `ignore`; branches with no open PR lose previews entirely; the `sast-scan.yml` comparison in the commit message isn't equivalent (it filters `push` to `main`).

dimitrieh (approved):
@ZJvandeWeg see comments above

dimitrieh:
@ZJvandeWeg Merged main in and took main's `test.yml` wholesale. The concurrency block already landed in #5601, guard included, and the conflict here was only against the shared build workflow from #5515.

What's left against main is the branch-deploy skip, with `command` kept next to the `ignore`.

- Trade-off, noted in the commit: a branch with no open PR now gets no preview at all.
- Worth ticking the Netlify test-plan box before merge. A wrong `ignore` fails quietly, by producing no preview.

PR #5425, opened 2026-07-29
PostHog analytics fix

KristopherLeads:
## Description
Gate analytics injection on import.meta.dev instead of process.env.NODE_ENV, which read as non-production in the server bundle and suppressed PostHog/GTM on all Nuxt pages. Changes `nuxt/server/plugins/analytics.ts`.

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass

PR #5433, opened 2026-07-30
fix: correctly gate Nuxt analytics/cookie banner injection to production

Yndira-E:
## Summary
Follow-up to #5425 (initial PostHog fix). That PR fixed the guard using `import.meta.dev`, but that alone would inject analytics on every build, which isn't what we want; it should only fire in real production.

1. `process.env.CONTEXT` isn't reliable at request time in Netlify's Function runtime, so the guard could still misfire. Now captured at build time via `runtimeConfig`.
2. The `serverAssets` path for analytics HTML was wrong. Before #5128 it worked fine as `'../src/_includes/analytics'`, resolved relative to `process.cwd()`. #5128 switched to Nitro's `serverAssets`, which resolves that same path against `nuxt/server` instead — one level short of the real folder. Fixed to `'../../src/_includes/analytics'`.

Both were needed: analytics/cookie banner never actually loaded on Nuxt pages (handbook, docs, changelog, pricing) because of this second bug, even after #5425.
## Checklist

### DIFF
--- a/nuxt/nuxt.config.ts
+++ b/nuxt/nuxt.config.ts
@@ -41,6 +41,13 @@ export default defineNuxtConfig({
     devtools: { enabled: true },
     modules: ['@nuxt/ui', '@nuxt/content', '@nuxtjs/seo', 'nuxt-studio', '@nuxt/image', './modules/docs-source'],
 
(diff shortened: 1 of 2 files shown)


KristopherLeads:
Thanks for the follow-up @Yndira-E!

PR #5439, opened 2026-07-30
Fix invisible headings when a stale color-mode preference is stored

dimitrieh:
## Description

Headings rendered near-white on the white page background (contrast ratio 1.00) for any visitor whose stored `nuxt-color-mode` was not `light`. `colorMode.preference` only covers first-time visitors: @nuxtjs/color-mode's inline head script reads the stored value ahead of the config and puts `.dark` on `<html>` before first paint, and its own `app:mounted` hook then copies that stored value back into the reactive state and re-persists it, so the state never settles on light and the class is never removed.

Renaming `storageKey` retires every stale value, because that key is exactly what the inline script looks up. Nuxt UI's dark tokens are additionally re-pointed at its own light values, so the class cannot change how anything renders even if it ever does appear.

This also removes `force-light-mode.client.ts`, which was verified in the browser to be a no-op: it assigned `preference` a value the hydrated state already held, so no watcher fired, and the module's `app:mounted` hook overwrote it afterwards regardless. Happy to keep it if you would rather.

Verified with Playwright against production and against this branch's config: with no stored value (the state every visitor lands in after this change) `.dark` is absent and `light` is persisted under both light and dark OS preferences; and with `.dark` forced on, the token override reproduces the light rendering identically across all 603 coloured elements on the page.

## Related Issue(s)

Follow-up to #5406 and #5422.

## Checklist

### DIFF
--- a/nuxt/assets/css/theme.css
+++ b/nuxt/assets/css/theme.css
@@ -21,6 +21,39 @@
     --ui-text-highlighted: var(--color-gray-700);
 }
 
+/* Single-theme safety net. colorMode is pinned to light in nuxt.config.ts, but
+   the class that drives Nuxt UI's dark palette is applied by an inline script
+   from client state, so a stale or hand-set value can still put `.dark` on
+   <html>. The page background comes from the 11ty stylesheet and has no dark
+   variant, so when that happened, headings resolved to near-white on white and
+   became invisible. Re-point Nuxt UI's dark tokens at its own light values so
+   the class cannot change what anything looks like. Keep this in sync with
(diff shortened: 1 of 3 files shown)


Yndira-E:
@dimitrieh have you checked what it looks like when dark mode is forced from the browser?

dimitrieh:
@Yndira-E no this was an in between meetings, hoping that the preview url would be up by the time we would look at it. I'll look into it

Yndira-E:
This other pr was the solution https://github.com/FlowFuse/website/pull/5446

PR #5451, opened 2026-07-31
Remove education program

Yndira-E:
## Description

As per title. It redirects to `/node-red/learn/` (open to suggestions here)

## Related Issue(s)

https://app.asana.com/1/1213818720452348/project/1214857503357659/task/1216967574736471?focus=true

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5460, opened 2026-07-31
Render the nav and footer from one shared data file

dimitrieh:
## Description

The mega nav and footer link lists were maintained by hand in two places, `src/_includes/layouts/base.njk` and the Nuxt `AppHeader`/`AppFooter` components, and nothing failed when the two drifted. Both now render from `src/_data/chrome.json`, so adding or removing a page is one edit. Nuxt draws icons from the same SVG files Eleventy loads, via `NavIcon` and an explicit import map, giving the glyphs a single source too.

Verified as a rendering no-op on the Eleventy side: served header markup and the header plus footer link sets are byte identical across five sampled pages, with only three HTML comments dropped. The Nuxt render needs checking on the deploy preview.

`nuxt/lib/chrome.test.mjs` guards what can still drift, including nav entries that point at a redirected path. That is the gap that let the removed education page keep a live nav link: adding a 301 makes `hyperlink` accept the old URL, so a stale entry passes CI.

## Related Issue(s)

Follows the leftover nav link found reviewing #5451. That PR removes the two education entries from the old markup; once it lands they come out of `chrome.json` instead.

## Checklist


dimitrieh:
An optimisation for us to consider

Yndira-E (approved):

PR #5473, opened 2026-08-03
Resolve product docs from the caller's checkout

dimitrieh:
## Description

Docs resolution moves into `nuxt/lib/docs-sync.mjs`, which looks for the docs in this order: `FLOWFUSE_DOCS_LOCAL`, a sibling `flowfuse` checkout, the snapshot committed on `live`, then a download from `main`.

The sibling rule is the one #5177 removed. It means a local checkout is picked up with no setup, and the flowfuse doc check can build a docs PR against its own changes instead of against `main`.

Three smaller fixes ride along: clones keep full history, so each page gets its own "last updated" date instead of every page sharing one; `npm run docs` is back; and `build.yml` commits the docs onto `live`, so production deploys stop downloading them at deploy time.

This does not change whether docs publish. That is FlowFuse/flowfuse#8029.

Review steps are in the comment below.

## Related Issue(s)

Part of #5432

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [x] Suitable unit/system level tests have been added and they pass
 - [x] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5474, opened 2026-08-03
Show the docs update time as relative, exact stamp on hover

dimitrieh:
## Description

The docs sidebar prints `Updated: 2026-03-18 15:07:47 +0000`. This shows `Updated: 5 months ago`, with the exact stamp on hover:

```html
<time datetime="2026-03-18T15:07:47+00:00" title="2026-03-18 15:07:47 +0000">5 months ago</time>
```

The label is worked out in the browser, since a prerendered page would freeze it until the next build. The server still renders the exact stamp, so it survives without JS.

Formatting lives in `nuxt/lib/relative-time.mjs` with 7 unit tests. Pairs with #5473, which gives each page its own date, but does not depend on it.

## Checklist

### DIFF
--- a/nuxt/lib/relative-time.mjs
+++ b/nuxt/lib/relative-time.mjs
@@ -1,0 +1,54 @@
+// Formats the docs `updated:` stamp. Kept free of Nuxt and Vue imports so it can be
+// unit tested with `node --test`.
+
+// Git writes commit dates as "2026-03-18 15:07:47 +0000". That is not a valid HTML
+// datetime value, and engines are not required to parse it, so normalise it here
+// rather than handing it to `new Date()` and hoping.
+const GIT_DATE = /^(\d{4}-\d{2}-\d{2})[ T](\d{2}:\d{2}:\d{2})(?:\s*([+-])(\d{2}):?(\d{2}))?$/
+
+const MINUTE = 60
+const HOUR = MINUTE * 60
+const DAY = HOUR * 24
+
+/**
+ * Normalise a git commit date to an ISO 8601 string, or return null if it is not one.
+ */
+export function toIso (raw) {
+    if (typeof raw !== 'string') return null
+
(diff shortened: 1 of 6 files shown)


nuxt/lib/relative-time.test.mjs
ZJvandeWeg, line 1: Are these executed? I vote to not test it actually

nuxt/lib/relative-time.mjs
```diff
@@ -0,0 +1,54 @@
+// Formats the docs `updated:` stamp. Kept free of Nuxt and Vue imports so it can be
```
ZJvandeWeg, line 1: I bet there's a library to do all this?

ZJvandeWeg (changes requested):

nuxt/lib/relative-time.test.mjs
ZJvandeWeg: They do run locally (`npm test` covers `nuxt/lib/*.test.mjs`), but I found `test.yml` never actually calls `npm test` in CI — it only builds and link-checks. Added a unit test step in c4af147d, and while at it fixed a flaky test in `handbookChanges.test.mjs` that hardcoded a specific PR as "most recent" and was bound to break as soon as the handbook changed again. Keeping these tests.

nuxt/lib/relative-time.mjs
ZJvandeWeg: Checked — no date library (date-fns/dayjs/moment/luxon) is a dependency anywhere in the repo. The formatting itself already uses the built-in `Intl.RelativeTimeFormat`; the only custom bit is `toIso()`, a small regex to normalize git's non-ISO `YYYY-MM-DD HH:MM:SS +ZZZZ` stamp. Didn't think it's worth pulling in a dependency just for that.

ZJvandeWeg (approved):

PR #5487, opened 2026-08-04
Add author pages, linked bylines, and sidebar author card for E-E-A-T

sumitshinde-84:
## Description

Add dedicated author pages, linked bylines, a sidebar author card, and an Expert Reviewed link to improve E-E-A-T.
Every published author now has a /blog/author/{slug}/ page with their bio, expertise, and all published articles. The byline links to that page, the sidebar shows the author's credentials at the top, the Expert Reviewed link points to our handbook blogging process, and the Person JSON-LD now points to a real author page.

https://github.com/user-attachments/assets/0b438fa4-144a-418a-9898-d31f7fc7975e

This makes it easier for readers and search engines to see who wrote the content, why they are qualified, and how the article was reviewed before publication. It also improves internal linking by connecting older articles with newer ones through author pages.

**It also matters for human review, not just crawlers. Google employs thousands of Search Quality Raters,  real people, not bots, who assess pages against the public Search Quality Rater Guidelines. Those guidelines ask raters to look for exactly this: who the author is, what makes them qualified on the topic, and whether the site is transparent about who stands behind the content. A byline with no destination gives a rater nothing to verify. An author page with a bio, credentials, and a body of work does.**

This follows the same best practice used by SEO leaders such as Ahrefs and Semrush.

- [semrush blog](https://www.semrush.com/blog/eeat/)
- [ahref blog](https://ahrefs.com/blog/eeat-seo/)

## Checklist


JoycePlaysFootball (approved):

Yndira-E:
Hey @sumitshinde-84, this looks great. Just a couple of notes:

* This will probably need a refactor since we migrated the handbook to Nuxt yesterday, so all the 11ty syntax will need to be replaced with Vue.
* I read through the Semrush article you shared, and it definitely supports having a clear byline and linking to the author’s page. I also noticed that both Semrush and Ahrefs display an author card in the sidebar, so I can see where the inspiration comes from. My only hesitation is that I couldn’t find any recommendation (either in the article or in Google’s guidance) that duplicating the full author bio is what provides the SEO/E-E-A-T benefit. It seems more like a design decision those sites have made rather than something the guidance explicitly recommends.

Since you’re adding a linked byline and we already have a full author bio at the end of the article, I’m not convinced the additional card adds any SEO or UX value. To me, it makes the sidebar feel a bit busier without a clear benefit.

sumitshinde-84:
@Yndira-E The reason we should emphasize the author is that, as I mentioned in the DM, Google now places a strong focus on E-E-A-T and relies on human evaluators, not bots, to assess who is behind an article. In addition, the EU AI Act places greater emphasis on transparency around authorship.

Regarding Semrush and Ahrefs, both follow a very similar practice. I don’t think this is simply a design decision, because I’ve been following their content for a long time. They closely observe each other and adopt ideas from one another. 

Another example is that both prominently indicate when an article has been reviewed, either through an “Expert Reviewed” badge or by displaying the reviewer’s name, especially after the recent EU AI Act. To me, that is a strong signal that they consider authorship really important.

That said, if you still think it isn’t needed and that it clutters the design, I’m happy to remove it from the right sidebar.

sumitshinde-84:
One more thing I’d like to address during the Nuxt migration is that we should move away from the `/blog/2023/09/blog-name/` URL structure, as it is generally considered a bad practice. Almost no one uses this structure anymore, and it can signal to Google that an article is old. If we want to align with Google’s emphasis on content freshness, we should signal recency through the Last Updated date rather than through the URL, because even after an article is updated, the URL still makes it appear old.

sumitshinde-84:
see this : https://yoast.com/what-is-a-permalink/#:~:text=as%20it%20grows.-,Configuring,-permalinks%20in%20WordPress

Yndira-E:
Thanks for the explanation. I’m definitely in favour of the linked byline and author page.

Have you found any guidance suggesting that duplicating the full bio provides additional SEO value? I get the inspiration from Semrush and Ahrefs, and there are definitely things I like about their layouts. But our right column already has more sections than theirs, so unless there’s a clear benefit, I’d prefer to keep the linked byline at the top and the full author bio at the end, rather than repeating it.

As for the permalink discussion, I’d rather keep this PR focused on the author changes and discuss URL structure separately, since that’s a broader site-wide decision.

Let me know once this has been refactored for Nuxt and I’ll take another look.

sumitshinde-84:
@Yndira-E, I’ll remove the author byline from the sidebar and convert it into a Nuxt component.

sumitshinde-84:
@Yndira-E, I have moved the implementation to Nuxt and also removed the byline from the right sidebar. Could you please re-review it?

nuxt/pages/blog/[...slug].vue
```diff
@@ -128,16 +131,22 @@ if (routeInfo.value.kind === 'post') {
         <h4 v-if="page.subtitle">{{ page.subtitle }}</h4>
         <div class="flex flex-wrap items-center gap-1 text-sm text-gray-500 mt-4">
           <span>By</span>
-          <span v-for="(author, i) in authorMembers" :key="i">
+          <span v-for="(author, i) in authorMembers" :key="author.slug">
             <span v-if="i > 0">, </span>
-            <span class="font-medium">{{ author ? author.name : 'FlowFuse' }}</span>
+            <NuxtLink :to="authorPath(author.slug)" class="font-medium hover:underline">{{ author.name }}</NuxtLink>
```
Comment (Yndira-E) [inline nuxt/pages/blog/[...slug].vue]: Could you please add back a fallback to "FlowFuse" for when authorMembers is empty? Right now the byline just renders nothing after "By" in that case, while the JSON-LD schema above still falls back to "FlowFuse", so they're inconsistent.

nuxt/composables/useTeam.ts
```diff
@@ -17,18 +22,52 @@ function keyFor(path: string): string {
...
+    return `/blog/author/${slug}/`
+}
+
+// Person node shared by the blog post byline (article author) and the author page itself,
+// so both resolve to the same @id and Google can join them up.
+export function authorSchema(member: TeamMember) {
+    const sameAs = [
+        member.linkedin && `https://www.linkedin.com/in/${member.linkedin}`,
+        member.github && `https://github.com/${member.github}`,
+        member.twitter && `https://twitter.com/${member.twitter}`,
```
Yndira-E, line 55: @sumitshinde-84 how come we use it here but don't render it on the card or author page?

nuxt/pages/blog/author/[slug].vue
```diff
@@ -0,0 +1,98 @@
...
+            .select('path', 'title', 'date', 'authors', 'description', 'meta', 'image')
+            .order('date', 'DESC')
+            .all()
+        return all.filter(entry => !isFuturePost(entry.date) && (entry.authors || []).includes(slug.value))
+    }
+)
+
+// Only sumit-shinde carries `expertise` today; `knowsAbout` is the older key and
+// stands in for it on anyone who has one but not the other.
+const topics = computed(() => author.value?.expertise || author.value?.knowsAbout || [])
```
Comment (Yndira-E) [inline nuxt/pages/blog/author/[slug].vue]: @sumitshinde-84 I checked and you're the only author with either key (`expertise `or `knowsAbout`), nobody else has data yet. Also, `authorSchema()` only reads `knowsAbout` for the JSON-LD, so `expertise` doesn't feed SEO at all. Worth unifying to just `knowsAbout` so there's one source of truth for both the UI chips and the schema?

nuxt/pages/blog/[...slug].vue
Comment (Yndira-E) [reply nuxt/pages/blog/[...slug].vue]: Also, would be worth a build-time check (e.g. in collectAuthorRoutes) that warns when an authors slug has no matching data file, so typos surface in the build log instead of shipping silently. Not blocking for this PR, just flagging for a follow-up.

nuxt/composables/useTeam.ts
sumitshinde-84: We are rendering these handles on the author card, also this actually used for the JSON schema profile items added to the author page schema.

Yndira-E: @sumitshinde-84 I checked `nuxt/pages/blog/author/[slug].vue` and `nuxt/components/BlogAuthorCard.vue` and twitter is not included. It wasn't on main either...

I also noticed you don't have twitter and you're the biggest contributor so I guess is not a big deal, but might be worth adding it for those who do.

sumitshinde-84: ahh you are talking about twitter, understood !

sumitshinde-84:
@Yndira-E done, applied all the suggestions !

Yndira-E (approved):

PR #5488, opened 2026-08-04
Add the FlowFuse Application Guide at /application-guide

robmarcer:
## Description

Adds the **FlowFuse Application Guide** as a new top level `/application-guide` section, and
ports Drew's prototype guides into it: the **FlowFuse Guide** (11 pages) and the **Node-RED
Guide** (7 pages), 41 tabbed pattern panels in total.

This is the "get the content live so we can see it and share it" step agreed in Slack. It is a
first iteration, so it is deliberately **unlinked**: nothing anywhere on the site points at
`/application-guide`. The pages are reachable by URL only, which is enough to share them for review
without changing the site for everyone. Wiring the section into the site navigation (Michael's
Resources / Learn suggestion) is a separate, later PR.

Source material is Drew's prototype on `demo-apps.flowfuse.cloud` (`flowfuse-guide-overview`,
`node-red-guide-overview` and their sub pages). The copy is ported close to verbatim, including
his phrasing and punctuation, so he can check it against what he has been using with prospects.

The Use case Registry is **not** included. It is an interactive app rather than page content,
and Dimitrie is separately working on the use case pages that will point at this material.

### The section has its own chrome, not the site's

This is the main thing to look at. `/application-guide` does not use `AppHeader` / `AppFooter` and
does not look like the rest of flowfuse.com. It keeps the look of the prototype so it reads as
its own thing under the FlowFuse brand:

* A dark bar (`#1F2937`) with the FlowFuse wordmark and the section name, and the 3px red rule
  under it, taken from the prototype.
* A light grey canvas with a centred white reading column.
* The guide sidebar beside the content on every page in the section, the landing page included.
* Violet accent (`#5048E5`) on the flagged intros, tab pills, "use it when" blocks and kickers.

**The bar holds only the logo and the words "FlowFuse Application Guide".** There is no site nav, no utility
bar, no footer and no other links, so the FlowFuse logo is the single route back out to the rest
of the site.

### What this adds

* **Content**: `nuxt/content/application-guide/{flowfuse,node-red}/*.yml`, one file per
  page. Every page is the same structure (a list of blocks, and most blocks are a tabbed set of
  patterns with a diagram, a summary, "use it when", "how it works" and component lists), so the
  guides are a `data` collection rather than prose markdown, following the same pattern as
  `certified-nodes.yml` and `plans/*.yml`. A copy edit is a YAML edit, no Vue involved.
* **Pages**: `/application-guide/` (section landing), `/application-guide/flowfuse/{slug}/` and `/application-guide/node-red/{slug}/`.
* **Layout and styling**: `nuxt/layouts/application-guide.vue` and
  `nuxt/assets/css/application-guide.css`. The stylesheet is namespaced under `.ag-root`,
  because the global site stylesheet still loads
  and styles bare elements such as `a` and `h1`.
* **Components**: `GuideBlocks` (dispatches a page's block list), `GuideTabs`, `GuidePanel`,
  `GuideDiagram`, `GuideLeftNav`, plus shared types in `nuxt/types/application-guide.ts`.
* **Diagrams** are rebuilt as one data driven `GuideDiagram` component (zones, boxes, labelled
  connectors) rather than 30 odd hand drawn SVGs, so they stay consistent and remain editable.
  They carry the same structure as the prototype but they are not a pixel copy.
* **Deep links**: the selected tab is reflected in the URL hash and each panel carries that slug
  as a real element id, so a single pattern can be linked to directly, for example
  `/application-guide/flowfuse/hardware-apps/#configurable-app`. The overview pages use those
  links, which is how Drew shares one idea at a time on a call.

### What it changes outside the section: three plumbing lines, no links

29 of the 32 files are new and live entirely inside the new section. The three existing files
that change get plumbing only:

| File | Change |
|------|--------|
| `nuxt/content.config.ts` | Registers the new `applicationGuide` collection. Existing collections untouched. |
| `nuxt/nuxt.config.ts` | Adds a helper that lists the new routes and spreads them into the prerender list, and adds the section to the existing sitemap `exclude` array. Nothing else in either list changes. |
| `nuxt/server/middleware/legacy.ts` | One line adding `/application-guide` to the dev proxy prefixes so `npm run dev` does not send it to 11ty. That middleware returns early unless `NODE_ENV=development`, so it has no production effect. |

No header, footer, utility bar, menu, `src/_data`, `src/_includes` or 11ty template is touched.

### Deliberately not included

* **No navigation anywhere.** As above, this iteration is unlinked on purpose.
* **No sitemap entries.** The section is excluded from `sitemap.xml` in `nuxt.config.ts`, so
  it is shareable by URL without being advertised to search engines. Worth removing that
  exclusion when the section is linked and ready to be indexed. Note the pages are not
  `noindex`, so a URL that leaks could still be crawled; happy to add a `routeRules` entry if we
  would rather they were not.
* **No audit of docs cross links.** Two pointers come straight from Drew's originals: FlowFuse
  Foundations links to `/docs/user/concepts/` and Node-RED Foundations to `nodered.org/docs`.

## Related Issue(s)

None. This came out of a Slack thread with @MichaelBDavis and Drew Gatti. Happy to open an issue
to track the navigation follow up.

## Checklist

 - [ ] Suitable unit/system level tests have been added and they pass. There is no test framework
   for page content in this repo. Verified by running `npm run dev:nuxt` locally: all 19 new
   routes return 200, an unknown guide slug 404s, all 31 deep link anchors resolve to a real
   element id, tab switching works, no console errors, every page carries the section chrome and
   zero site header markup, the sidebar renders exactly once per page, and `/pricing`,
   `/handbook/` and `/resources/publications/` still render, so the content config and dev proxy
   changes are non breaking.
 - [ ] Documentation has been updated. n/a, this is website content.
 - [ ] For blog PRs, an Art Request has been created. n/a.

### Review notes

The copy is Drew's, ported rather than rewritten, so it keeps his voice and some FlowFuse
internal shorthand.

* **@MichaelBDavis** the section is the **FlowFuse Application Guide** at `/application-guide`,
  with the two guides at `/application-guide/flowfuse/` and `/application-guide/node-red/`. Worth
  knowing it does not wear the site chrome and is not linked from anywhere in this iteration.
* **@dgatti0213** please check the content is faithful, especially the diagrams, since those are
  rebuilt rather than copied. Flag anything that reads wrong outside the dashboard context.

🤖 Built with Claude Code. The copy is Drew's; the layout, components, content structure and
wiring were generated and verified locally, and still want a human read before merge.


MichaelBDavis:
Don't care for the term "best practices". Here are few ideas, in order of my personal preference.
- FlowFuse Patterns Guide
- FlowFuse Architecture Guide
- FlowFuse Field Guide

robmarcer:
I assume those examples are page titles @MichaelBDavis. We also need to agree the path (flowfuse.com/SOMETHING) which ideally is short and easy to read / spell.

MichaelBDavis:
@robmarcer /patterns/ or /architecture/ works

Do you or @dgatti0213 have a preference?

robmarcer:
/architecture/  for me @dgatti0213 @MichaelBDavis

dgatti0213:
Agreed on dropping "best practice."

My vote is `/guides/`, and I'd leave `/architecture/` alone. 

One request either way: the sub-paths are redundant right now. Not `/guides/flowfuse-guide/` but:
- `flowfuse.com/guides/flowfuse/`
- `flowfuse.com/guides/node-red/`

I only did the long links because of a dashboard 2.0 limitation BTW. 

Agreed on leaving the Use case Registry out. Ideally, out use case pages read the way this was intended to. The idea is that the use case pages lead you to the guides and the guides lead you back or to the demo env. 

Ill need some time on the content review but i do want to link the original docs the site was spawned from here for ref: 
https://drive.google.com/drive/folders/1eLI_eQHx7MXJD0JzmfoUoAgx-1f1UxAu

MichaelBDavis (approved):

PR #5489, opened 2026-08-04
Remove leftover Nunjucks artifacts from blog posts

ZJvandeWeg:
## Summary
Follow-up to the blog-to-Nuxt migration. Blog posts now render through Nuxt/`@nuxt/content` (MDC), which doesn't use Nunjucks and won't evaluate any Nunjucks syntax left in the markdown source — it renders as literal template text on the page instead.

- Removed dead `{% raw %}...{% endraw %}` escape tags (only needed to protect `{{ }}` from 11ty's Nunjucks processor) from 11 posts
- Replaced 135 `{% include "sign-up-url.njk" %}` occurrences (117 files) with the literal URL the partial always resolved to (`https://app.flowfuse.com/account/create`)
- Replaced the one live `{% include "hubspot/hs-form.njk" %}` (`certified-nodes-v2.md`) with the existing `HubSpotForm.vue` component via its `::hub-spot-form` MDC shortcode
- Inlined the two customer-story tiles in `building-on-flowfuse-devices.md` as static markup, replacing the `{% from %}` macro import + `{% for %}` loop (no reusable Nuxt component existed for this one-off use)
- Dropped now-unused `templateEngineOverride`/`stories`/`certificationHubspot` frontmatter that only existed to support the removed Nunjucks logic

## Test plan
- [x] Verified via the running Nuxt dev server that `/blog/2025/07/certified-nodes-v2/` renders the HubSpot form div in place of the old include
- [x] Verified `/blog/2024/07/building-on-flowfuse-devices/` renders both story tiles, in the same order the old Nunjucks sort/reverse produced
- [x] Confirmed no `{% %}` tags remain anywhere under `src/blog`
- [x] Spot-checked flow-JSON code fences still parse as valid JSON after the raw-tag removal


Yndira-E:
@ZJvandeWeg there's something off with the story tiles styling on `/blog/2024/07/building-on-flowfuse-devices/`

Screenshot 2026-08-04 at 19 45 27

The image alignment within the cards isn’t quite right, and the hover effect isn’t smooth and still shows the underline decoration.

Not a blocker, just flagging it. If you can address it in this PR, great. Otherwise, I can create a follow-up PR for it, I just can’t get to it right now.

Yndira-E (approved):

PR #5495, opened 2026-08-04
Add a docs watcher for the dev server

### LINKED ISSUE #5432: README out of date with build instructions
### Description

PR #5177 changed how the website incorporates docs from the flowfuse/flowfuse repo. There are some outstanding problems:

`.claude/CLAUDE.md` was updated with information about the new process - but `README.md` was not updated and contains out of date information.

Specifically, the fact it now clones the ff repo by default rather than use a local copy of the docs without manual intervention. This has broken the FF repo build that verifies doc changes.

It has also added docs to `CLAUDE.md` about an npm task called `dev:nuxt`. This task doesn't exist.

It has also removed the ability to watch the docs for changes and trigger automatic rebuild - was previously the `npm run dev:docs` task. When working on product docs, I now have to manually restart the website build every time and wait for it to complete.

I also find I'm routinely hitting https://github.com/nuxt/nuxt/issues/35253 due to the version of nuxt we are using.

In addition to knowing about the `FLOWFUSE_DOCS_LOCAL` env var, I also have to set `TMPDIR=/tmp` to workaround that nuxt issue.

### Epic/Story

_No response_

dimitrieh:
## Description

Editing a product doc currently means restarting the dev server, because `docs-source.ts`
syncs once inside Nuxt's `setup()` and never again. This watches the resolved docs and syncs
each change.

The first version of this ran the existing whole-tree sync on every save. That deletes and
recreates all 130-odd pages, and the dev server ran out of heap re-indexing them, on the
first edit. So the sync gained a single-file path and the watcher uses it: one edit, one
write. A full build still produces byte-identical output, which matters because the CI docs
check and the Netlify deploy both run through it.

chokidar rather than nodemon, because nodemon can only re-run a command and cannot say which
file changed.

Verified against a real flowfuse checkout: 30 consecutive edits with the dev server alive and
its memory flat, plus new page, delete, rename, README and image cases, and the clone case
exiting instead of watching nothing.

## Related Issue(s)

Closes #5432

## Checklist


dimitrieh:
@knolleary for your consideration when you're back

knolleary:
I tried this locally a few times without success - I get out of memory errors every time I change any docs.

> [4] FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

knolleary:
The memory required by nuxt to rebuild the site when docs update is completely unworkable. It should not require GBs of memory to turn some markdown into html.

The only solution I can think of would be to make the script smarter - only copy files that have changed. That is its own resource intensive operation, but, to be frank, I'd rather wait a few seconds for the rebuild to happen, then have to restart the whole stack every time.

dimitrieh:
@knolleary I did another run on it, can you check it again?

knolleary:
Looks much better - thanks @dimitrieh

knolleary (approved):

ZJvandeWeg:
@dimitrieh @knolleary Why a hacky script? I would've expected this to use the Nuxt collection primitive?

I don't think this was ready to be merge, or reviewed in a decent manner.

PR #5498, opened 2026-08-04
Use case pages: add a How It Works section

dimitrieh:
## Description

Use case pages carried only *what* and *why*. This adds a `howItWorks` block to the use case layout so a page can also show **how** FlowFuse builds the thing: the component breakdown, the design pattern chosen, an end-to-end architecture diagram, and the hand-off into the product docs.

Filled in for the two use case pages, Production Monitoring and Shop Floor Communication. Diagrams are `placehold.co` stand-ins pending an art request; the layout renders a dashed border and a "Placeholder diagram" caption while `diagram.placeholder` is set.

Two supporting fixes:

- Section eyebrow numbers were hardcoded while the sticky nav numbers came from `loop.index`, so they desynced on any page omitting a block. Both now derive from one lookup over the nav array.
- `architecture: "edge-connectivity"` on Shop Floor Communication was front-matter the layout never read. It now sits inside `howItWorks` and renders as a link to the architecture page the use case sits on.

Content note for review: the design pattern assigned to each page is a first pass and needs a sales sign-off. Per-pattern docs pages do not exist yet, so `docsUrl` points at the nearest real pages (`/docs/user/devops-pipelines/`, `/docs/user/envvar/`) rather than at pattern docs.

## Related Issue(s)

None.

## Checklist

### DIFF
--- a/src/_includes/layouts/use-case.njk
+++ b/src/_includes/layouts/use-case.njk
@@ -9,9 +9,32 @@ sitemapPriority: 0.7
      problem        hero sub-headline (the pain in one line)
      customerPain   { heading, intro[], cards[ {icon,title,detail} ] }
      outcomeFirst   { heading, intro, dimensions[ {label,title,detail} ] }
+     howItWorks     { heading, intro, designPattern{}, diagram{}, pieces[], architecture, docsCta{} }
      whyItMatters   { heading, intro, points[ {title,detail} ] }
      competition    { heading, intro, traps[ {label,title,detail} ] }
      comparison     { without[ {title,detail} ], with[ {title,detail} ] }
+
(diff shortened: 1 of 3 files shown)


dimitrieh:
Preview affected pages:

- https://deploy-preview-5498--flowforge-website.netlify.app/use-cases/production-monitoring/
- https://deploy-preview-5498--flowforge-website.netlify.app/use-cases/shop-floor-communication/

dgatti0213:
I like the direction but you need to break down the problem like this: https://docs.google.com/drawings/d/1u2xDF5igAhPMYzAmBxOY4eRKJvdKuY7MoGWbPPUIUUU/edit

production monitoring consists of a software app and a hardware app

Main point: Use case does not = 1 app
use-case could mean many apps that sum to the use case.

PR #5499, opened 2026-08-04
Add a Platform page under Platform > Product

dimitrieh:
## Description

The website explains what FlowFuse does and why it matters, but never what the platform is made of or how the pieces fit together. That leaves no page to send a technical evaluator to after the homepage, and the current next step offered is the Expert, which does not answer "what are the major components".

Adds `/platform/`, wired into the top nav and the footer under **Platform > Product**, ahead of Features. Sections:

1. **How code moves.** The two distribution models, shipping a whole application by pipeline versus shipping building blocks via a library. This is the reason the word "platform" applies.
2. **Major components.** Everything FlowFuse does, in five groups: where applications run, how they get built, how code is distributed and managed, what they connect to, how it is governed.
3. **How it fits together.** End-to-end architecture diagram, currently a `placehold.co` stand-in pending an art request.
4. **Design patterns.** The three hardware and three software patterns.
5. **Where Node-RED fits.** Platform first, then Node-RED.

Every component links to an existing product page or docs page; no new docs are assumed.

Content is a first pass and needs a sales sign-off on the pattern names and the component groupings.

## Related Issue(s)

None.

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5500, opened 2026-08-04
Homepage: say what an app platform means

dimitrieh:
## Description

> Stacked on #5499. Base is `feat/platform-page` because this section links to `/platform/`, which #5499 adds; the link checker would fail against `main`. Retarget to `main` once #5499 merges.

The homepage asserts that FlowFuse is a platform and then moves straight to capabilities, so a visitor never learns what the platform is made of or why the word applies. The next step offered is the Expert, which is the wrong place to send someone who does not yet understand the components.

Adds one section after the solution statement that:

- States the claim, that "app platform" is about how FlowFuse distributes and manages code.
- Backs it with the two distribution models, shipping a whole application by pipeline versus shipping building blocks via a library.
- Lists the major components grouped by what they do, as a compact scannable list rather than another card grid.
- Hands off to `/platform/` for the detail instead of dead-ending at the Expert.

No existing homepage content is removed or reordered.

## Related Issue(s)

None.

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5500--flowforge-website.netlify.app/
- https://deploy-preview-5500--flowforge-website.netlify.app/platform/

dgatti0213:
So this is a great start! but IMO the homepage should sit one level above the platform detail page. Homepage in terms visitors already recognize, platform page for our specifics.

Right now both pages carry the same five groups with the same product names, so the homepage is the platform page with the explanations removed. "Instances · Devices · Groups" and "Snapshots · Pipelines · Environment Variables" are our vocabulary, not the visitor's. Anyone who can parse them is already headed to /platform.

**Suggested values:**
Where it runs: edge devices, your infrastructure, or hosted
How it's built: visual flow programming, in the browser
How it's distributed: one app to every site, or a library of shared blocks
What it connects to: OPC UA, Modbus, MQTT, EtherNet/IP, databases, ERP, REST, Limitless! 
How it's governed: SSO, role-based access, audit logs, SOC 2 Type 2

allthedoll:
should we get @JoycePlaysFootball's input on where it goes?

dgatti0213:
These pages have alot of the content im recommending needs to be apart if the customer journey on the homepage and such:
https://flowfuse.com/application-guide/flowfuse/foundations/
https://flowfuse.com/application-guide/flowfuse/app-delivery-methods/

Im not a marketer, im just trying to highlight the minimum content required for someone to understand what we do and how they can envision it in their org. One thing to also think about is the guide doesn't direct you to a product-line so that also needs to be taken into account.

dimitrieh:
cc: @KristopherLeads

PR #5515, opened 2026-08-05
Call the shared website build workflow

dimitrieh:
Contributes to [Website preview on documentation pull requests](https://github.com/FlowFuse/engineering/issues/237).

## Description

Calls the shared `build_website.yml` in FlowFuse/github-actions-workflows instead of keeping a second copy of the website build here. Same steps, one definition, shared with FlowFuse/flowfuse documentation pull requests.

The job keeps its name, so the reported check becomes `test_website / Build and check`. The `protect main` ruleset has to require that context before this merges.

## Related Issue(s)

- Shared workflow: https://github.com/FlowFuse/github-actions-workflows/pull/211
- Other caller: https://github.com/FlowFuse/flowfuse/pull/8070

## Checklist


dimitrieh:
Context at https://github.com/FlowFuse/engineering/issues/237

dimitrieh:
Slack thread: https://flowfuse.slack.com/archives/C032Q63FGG1/p1785930190689399

Yndira-E:
@dimitrieh Could you elaborate on the skip-images point? The way it's worked so far is just skipping image optimization, images are still "real" and visible either way.

The reason we skip it is that it takes longer than necessary for a preview, and with Netlify's recent issue where it drops the cache for no reason, we were getting ~90min builds on random production deploys too, so it's currently disabled for production as well.

.github/workflows/site-preview.yml
```diff
@@ -0,0 +1,216 @@
...
+# FlowFuse/flowfuse documentation pull requests, so a change to the build, the link
+# check or the preview upload is picked up by both callers without being copied.
+on:
+  workflow_call:
+    inputs:
+      website_repository:
+        description: 'Repository holding the website source.'
+        type: string
+        default: 'FlowFuse/website'
+      website_ref:
```
ppawlowski: What is the purpose of the parametrisation of a repository name and ref?

.github/workflows/site-preview.yml
```diff
@@ -0,0 +1,216 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      preview_url: ${{ steps.deploy.outputs.preview_url }}
+    steps:
+      - name: Generate a token
+        id: generate_token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          # Accepts either the App ID or the Client ID as the JWT issuer, so both callers
+          # can pass whichever of the two their own secret holds.
```
ppawlowski:
```suggestion

```

This is not true.

.github/workflows/site-preview.yml
```diff
@@ -0,0 +1,216 @@
...
+          ref: main
+          path: 'blueprint-library'
+          token: ${{ steps.generate_token.outputs.token }}
+      - name: Cache image pipeline output
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          # Rolling key. A static key is only ever written once, so the entry went stale and
+          # never picked up newly processed images. These are the two directories that
+          # netlify.toml's cache plugin keeps warm on Netlify's own builds.
+          key: img-pipeline-${{ github.run_id }}
```
ppawlowski: This is very expensive, the `github.run_id` is unique per run, so the exact lookup can never hit. That means the cache restore step always falls through to restore-keys and cache  save always fires - a fresh full copy every workflow execution.
Please have in mind that there is a 10G storage limit for cache. I am pretty sure we will hit this limit very quickly with this approach.

.github/workflows/site-preview.yml
```diff
@@ -0,0 +1,216 @@
...
+      - name: Generate a token
+        id: generate_token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          # Accepts either the App ID or the Client ID as the JWT issuer, so both callers
+          # can pass whichever of the two their own secret holds.
+          client-id: ${{ secrets.ci_app_id }}
+          private-key: ${{ secrets.ci_app_key }}
+          owner: ${{ github.repository_owner }}
+          repositories: blueprint-library
```
ppawlowski: Scoping the token to the `FlowFuse/blueprint-library` will cause operations on the `flowfuse/website` and `flowfuse/flowfuse` repositories to fail.

.github/workflows/test.yml
```diff
@@ -3,48 +3,26 @@ name: Tests
...
-        run: npm run build:nuxt:skip-images
-        working-directory: 'website'
-      - name: Check links
-        uses: untitaker/hyperlink@9375bc4063712ad490d5eb3d54df0b6aade15e54 # 0.3.2
-        with:
-          args: website/nuxt/dist/ --check-anchors --sources website/src
+      - name: Require the site build to have succeeded
+        run: |
+          echo "build_site result: ${{ needs.build_site.result }}"
+          [ "${{ needs.build_site.result }}" = "success" ]
```
ppawlowski: Just update the ruleset instead of introducing a separate job that spins up a dedicated runner and consumes minutes.

ppawlowski (changes requested):
1. Please move the reusable GitHub Actions workflow to the `FlowFuse/github-actions-workflows` repository
2. Address my comments while moving

.github/workflows/site-preview.yml
dimitrieh: Two callers, two different sources.

- `actions/checkout` defaults to the repo that owns the run. Called from `FlowFuse/flowfuse` that's flowfuse, not the website, so the website source has to be named explicitly. Now that the workflow sits in `github-actions-workflows` that's true for both callers.
- `website_ref` is what lets this repo's own CI build the commit under test (`github.sha`) rather than `main`, while flowfuse docs PRs stay on website `main`.

Regarding `website_repository`, I'm fine hardcoding `FlowFuse/website` and keeping only the ref if you prefer. Fork PRs can't build anyway since they don't get the secrets. Let me know which you'd rather have.

.github/workflows/site-preview.yml
dimitrieh: Removed.

On the mechanism, for the record: `main.js` resolves `core.getInput("client-id") || core.getInput("app-id")` and passes the result straight through as the JWT issuer, and `test.yml` on main already feeds `GH_CI_READONLY_APP_ID` into `client-id` today. What I can't verify is what that secret actually holds, so the comment was claiming more than I can back up. Dropped it.

.github/workflows/site-preview.yml
dimitrieh: Agreed, that key was wrong. `github.run_id` is a guaranteed miss, which means a full save every single run.

Two changes:

- The step now only runs when `deploy_preview` is true. Everything else builds with `SKIP_IMAGES`, which sets `nuxt.config.ts`'s image provider to `none`, so those runs were writing a cache nothing ever read.
- The key is content based now: `hashFiles` over the source images under `src`, with `restore-keys: img-pipeline-`. Unchanged images give an exact hit and no save at all. A run that touches an image restores from the prefix and writes one entry.

That gets it to one entry per branch per image content, rather than one per run. Good catch on the 10 GB, the output tree is in the tens of megabytes so the old version would have filled it fast.

.github/workflows/site-preview.yml
dimitrieh: I don't think it does, but tell me if I'm missing a path.

The app token is referenced exactly once, on the `blueprint-library` checkout. The website and docs checkouts use the caller's default `GITHUB_TOKEN`, and both repos are public, so a caller-scoped token reads them fine. The old `test.yml` listed `website` because it passed the app token to the website checkout, and this version doesn't.

Happy to add `website` back if you'd rather the two lists stay identical, but as written it's scope the workflow never uses. I've put a comment on the step saying so.

.github/workflows/test.yml
dimitrieh: Done, the gate job is gone.

The calling job is named `test_website`, so the required context becomes `test_website / Build and check`. Can you update the "protect main" ruleset to that name? I'll hold the merge until it's in, otherwise every website PR blocks on a check that never reports.

dimitrieh:
Makes sense, that's where it belongs. Both points are done:

- FlowFuse/github-actions-workflows#211 adds the workflow as `build_website.yml`, with all five comments addressed.
- This PR shrinks to the `test.yml` change, calling that workflow.
- https://github.com/FlowFuse/flowfuse/pull/8070 moves to the same ref.

Both callers point at the `add-website-build-workflow` branch until #211 lands, then move to `@main`. Merge order is workflows repo, then here, then flowfuse.

One trade-off worth naming: a build change can't be tested in the PR that makes it any more, since the caller has to point at a merged ref. That's why it started here, but I'd rather follow the convention.

Two things I'd like back from you:

- The "protect main" ruleset needs to move to `test_website / Build and check` before this merges.
- I've pushed back on the token scoping comment, take a look and tell me if I'm wrong there.

dimitrieh:
@ppawlowski moved to FlowFuse/github-actions-workflows#211, comments addressed, replies above. This one is just the caller now.

.github/workflows/site-preview.yml
ppawlowski: Had to refresh my knowledge, and you are correct - currently the`github` context, including token, is automatically inherited by a reusable workflow ([source](https://docs.github.com/en/actions/reference/workflows-and-actions/reusing-workflow-configurations#github-context)). 

Historically, `github.token` had to be explicitly passed to the reusable workflow via a secret.

ppawlowski:
The `build_website` reusable workflow has been released. Please:
1. Resolve conflicts
2. Update the reference tag to `build_website/v1`

dimitrieh:
@ppawlowski done: merged main, ref is now `build_website/v1`. The run reports as `test_website / Build and check`, so the tag resolves.

Two blockers left:

- The "protect main" ruleset still requires the context `test_website`, which no longer reports. It needs to be `test_website / Build and check`.
- Website output content cache (#5562) added an og-image cache step to `test_website` on main. A `uses:` job takes no steps, so the merge drops it. It has to move into `build_website.yml`, and its `github.run_id` key needs replacing with a content-based one.

ppawlowski:
> The "protect main" ruleset still requires the context test_website

I will update it once this PR is merged

> Website output content cache (https://github.com/FlowFuse/website/pull/5562) added an og-image cache step to test_website on main. A uses: job takes no steps, so the merge drops it. It has to move into build_website.yml, and its github.run_id key needs replacing with a content-based one. 

Update the introduced reusable workflow if needed.

dimitrieh:
https://github.com/FlowFuse/github-actions-workflows/pull/213 created

dimitrieh:
@ppawlowski can you check what still needs to be changed? Not sure what needs changing still or what your review refers to

dimitrieh:
@ppawlowski the ruleset has to change before this merges, not after.

`protect main` requires a check named `test_website`. That name is gone. The check is now named `test_website / Build and check`.

So the required check never arrives and the pull request stays blocked. The ruleset has no bypass actors, so there is no way around it.

Point the required context at `test_website / Build and check` and this merges on its own. Everything else is green.

dimitrieh:
The order in which things need to happen is also defined in the issue managing this effort across repos: https://github.com/FlowFuse/engineering/issues/237

ppawlowski (approved):

PR #5107, opened 2026-06-01
Add guideline for FlowFuse Cloud Personal Access Tokens

hardillb:
## Description

Added a guideline regarding the use of FlowFuse Cloud Personal Access Tokens with AI agents.

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1215298579367718

PR #5520, opened 2026-08-06
handbook: allow scoped credentials for AI agents under review

robmarcer:
## Description

Rewrites point 5 of "Internal Use of AI by FlowFuse Team Members" in the AI Development and Customer Data Policy.

Point 5 currently bans giving any credential to an AI tool or agent outright. In practice there is legitimate work that needs an agent to hold a token, and a rule that is routinely inconvenient tends to get quietly ignored rather than followed. This replaces the blanket ban with a conditional that is stricter in the places that actually matter.

Under the new wording, a credential may be given to an AI tool or agent only when all of the following hold:

1. **Discuss it first** with at least one other team member before creating or sharing it.
2. **Document the discussion** by opening an Access / Permission Request issue in the admin repo, recording who was consulted, what the agent will do, the scope granted, and when the credential will be revoked.
3. **Scope it to the job**, for example a single FlowFuse team rather than the whole account.
4. **Read only by default**, with write access only where the task genuinely cannot be done without it.
5. **Rotate and revoke**, treating any credential an agent has held as spent, revoking it when the work is finished, and rotating anything that must persist per the Access Control Policy.

Two things stay prohibited outright, because neither can satisfy "minimum needed":

- Admin-scoped FlowFuse Cloud and Dedicated Instance PATs. This was the original concern behind point 5 and it is preserved, with the reasoning made explicit.
- Passwords and any other secret that cannot be scoped or revoked.

The net effect is a narrower blast radius than today. The current rule says no, which means the cases where someone does need it happen with no discussion, no record, and no scoping. This makes those cases visible and bounded.

## Related Issue(s)

None. Follow-up to #5107, which introduced point 5.

## Checklist

Most of the template checklist does not apply to a handbook policy change.

 - [x] I have considered the performance impact of these changes (n/a, content only)
 - [x] Suitable unit/system level tests have been added and they pass (n/a)
 - [x] Documentation has been updated (this PR is the documentation)
 - [x] For blog PRs, an Art Request has been created (n/a, not a blog post)

## Note for reviewers

@knolleary is the listed policy owner. @hardillb wrote the original point 5 and @allthedoll broadened it during review of #5107, so both have context on the intent being changed here.

The effective date in the header table has deliberately been left at 2026-02-18, matching how #5107 handled an amendment. Happy to bump it if the policy owner would prefer amendments to move that date.


hardillb:
> In practice there is legitimate work that needs an agent to hold a token, and a rule that is routinely inconvenient tends to get quietly ignored rather than followed

This is a problem, if this is the case then it needs to STOP now and be discussed.

As part of that discussion we need to understand what "there is legitimate work that needs an agent to hold a token" is and if it really does meet that criteria.

This can be discussed when Nick and Jam are back, but all work using tokens must stop until then, it is explicitly against stated policy.

nuxt/content/handbook/company/security/ai-development-and-customer-data.md
```diff
@@ -40,7 +40,14 @@ When using AI internally:
...
-5. Never give credentials to AI tools or agents. This covers Personal Access Tokens, API keys, passwords, and any other secret, whether you're pasting it into a chat or handing it to an agent to act with. Admin-scoped FlowFuse Cloud and Dedicated Instance PATs are the highest-risk case: an agent holding one can take destructive, account-wide actions.
+5. Credentials may be given to an AI tool or agent where the work genuinely needs it, provided all of the following hold:
+   - **Discuss it first.** Talk the use case through with at least one other team member before you create or share the credential. If you cannot explain to a colleague what the agent will do and why it needs this access, that is your answer.
+   - **Document the discussion.** Record who you spoke to, what the agent will be doing, the scope granted, and when the credential will be revoked. Open an [Access / Permission Request](https://github.com/FlowFuse/admin/issues/new?template=access-request.md) issue in the admin repo on GitHub, the same place other access decisions are recorded. A verbal or DM conversation nobody can find later does not count.
```
knolleary: Step 1 should be to raise the the permission request - and it isn't just there to document the discussion, it is to get approval from leadership.

knolleary (changes requested):

nuxt/content/handbook/company/security/ai-development-and-customer-data.md
robmarcer: I felt someone should discuss it with a team member before asking for permission so to try to mitigate requests being created where they are not needed.

I didn't take requesting permission to be 'just there to document the discussion'. How about if I refer to the Access Control Policy https://flowfuse.com/handbook/company/security/access-control/#access-control-policy-1 if you feel it reads like just a formality rather than an actual permission request?

knolleary: Nowhere in the steps you've proposed is 'receive permission' - its just chat about it, document it, do it (with the added guidance you've outlined). It should be explicit that permission is required.

robmarcer:
```suggestion
Credentials may be given to an AI tool or agent where the work genuinely needs it, provided the access has been approved in advance and all of the following hold:
   - **Get approval first.** Open an [Access / Permission Request](https://github.com/FlowFuse/admin/issues/new?template=access-request.md) issue in the admin repo on GitHub before you create or share the credential, the same route as any other access request under the [Access Control Policy](./access-control.md). Set out what the agent will be doing, the access it needs, the scope you are asking for, and when the credential will be revoked. The request has to be approved before the credential is created: your manager as a minimum, and the CTO where it touches critical resources such as production. If you are not certain the access is needed at all, talk it through with a colleague before raising the request.
```

nuxt/content/handbook/company/security/ai-development-and-customer-data.md
```diff
@@ -40,7 +40,13 @@ When using AI internally:
 2. Use internal, synthetic, or publicly available data for experimentation whenever possible.
 3. Review AI-generated output before relying on it or including it in the product.
 4. Do not use AI tools to bypass security controls, access restrictions, or approval processes.
-5. Never give credentials to AI tools or agents. This covers Personal Access Tokens, API keys, passwords, and any other secret, whether you're pasting it into a chat or handing it to an agent to act with. Admin-scoped FlowFuse Cloud and Dedicated Instance PATs are the highest-risk case: an agent holding one can take destructive, account-wide actions.
+Credentials may be given to an AI tool or agent where the work genuinely needs it, provided the access has been approved in advance and all of the following hold:
```
robmarcer:
```suggestion
5. Credentials may be given to an AI tool or agent where the work genuinely needs it, provided the access has been approved in advance and all of the following hold:
```

nuxt/content/handbook/company/security/ai-development-and-customer-data.md
robmarcer: Hello, I have addressed these points (hopefully). Not sure if you get an alert when I edit the PR so messaging to be sure.

knolleary (approved):
Thanks for the update.

PR #5365, opened 2026-07-22
Stub release 2.33 blog

allthedoll:
## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5524, opened 2026-08-06
Changelog: Remove tier badges on changelog entries

Yndira-E:
## Description

The badges are outdated against our current packages, I'm disconnecting the changelog from that logic until it gets updated.

## Checklist

### DIFF
--- a/nuxt/components/ChangelogListItem.vue
+++ b/nuxt/components/ChangelogListItem.vue
@@ -1,15 +1,9 @@
 <script setup lang="ts">
-import { findFeatureByChangelog, deriveTierLabel } from '../composables/useFeatureCatalog'
-
 const props = defineProps<{
     entry: { path: string, title: string, description?: string, date: string | Date, authors?: string[] }
 }>()
 
 const authorNames = computed(() => useAuthorNames(props.entry.authors))
-
-const catalogFeature = computed(() => findFeatureByChangelog(props.entry.path))
-const tierCloud = computed(() => deriveTierLabel(catalogFeature.value?.cloud))
-const tierSelfHosted = computed(() => deriveTierLabel(catalogFeature.value?.selfHosted))
 
 const formattedDate = computed(() => new Date(props.entry.date).toLocaleDateString('en-US', { year: 'numeric', month: 'short', day: 'numeric' }))
 </script>
@@ -24,7 +18,6 @@ const formattedDate = computed(() => new Date(props.entry.date).toLocaleDateStri
           <div class="author">{{ authorNames }}</div>
         </div>
       </NuxtLink>
-      <ChangelogTierBadges :cloud="tierCloud" :self-hosted="tierSelfHosted" />
     </div>
     <div class="flex-grow pt-4">
       <div class="prose">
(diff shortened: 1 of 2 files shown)


Yndira-E:
@dimitrieh will you take care of updating it to show the right packages? I didn't delete the entire thing, just disconnected it from the changelog entries.

dimitrieh:
Looks good to me, merge whenever. Disconnecting is the right call: the badges have been rendering "Enterprise" against Cloud/Self-Hosted since before repackaging, so they were actively wrong rather than just stale.

And yes, I'll pick up the catalog work. It's a bigger job than the changelog though, so some context on what I found while looking into it.

### The badges were mostly a docs thing, and they already broke

PostHog has `tier-badge-click` running at roughly 100 to 150 clicks a month from around 30 people through July. Nearly all of those were on `/docs/*` pages, not the changelog. They dropped to zero after the week of 27 July.

Cause is the docs Nuxt migration (#5177, merged 29 July). The `docsFeatureBadges` transform in `.eleventy.js` still runs and still emits HTML, but it writes into `nuxt/public/` where the Nuxt route wins, so the output gets thrown away. Same story for `releaseFeatures` on release blogs, the 2.33 release blog renders zero badges live. So this PR is removing the last remaining slice of a surface that had already gone dark everywhere else.

Worth flagging as a pattern: Eleventy transforms shadowed by a Nuxt route fail completely silently. Nothing errors, nothing warns, the build stays green.

### There are two catalogs and they've drifted apart

- `src/_data/featureCatalog.yaml` has the metadata (docsLink, changelog + release, solutions, subfeature, beta) but the tier axis is Cloud/Self-Hosted x Enterprise, so pre-repackaging.
- `nuxt/content/feature-catalog.yml` has the correct Edge/Hub/Fleet axis but no changelog links, no docs links, no solutions.

Only a minority of feature IDs overlap between the two, so this is a merge rather than an update.

### Two small things on this PR

- `ChangelogTierBadges.vue` and `composables/useFeatureCatalog.ts` have zero consumers once this merges. Nuxt auto-imports, so nothing will error, they'll just sit there. Fine to leave them since I'm rewiring both shortly, just noting it so they don't get forgotten.
- `writing-changelog.md` (Feature catalog and availability) and `release-blogs.md` still tell writers that adding a changelog url to `featureCatalog.yaml` renders tier badges automatically. That isn't true anymore. I'll fix those in the follow-up rather than hold this up.

### What I'm doing in the follow-up

- Making `feature-catalog.yml` the single catalog: extending the schema in `content.config.ts` with docsLink, changelog, solutions, subfeature and beta, then porting the metadata across and reconciling the IDs.
- Badges become plan badges (Edge / Hub / Fleet), dropping the Cloud VS Self-Hosted split since packaging isn't deployment specific anymore.
- Rewiring the changelog badges, the handbook Feature Catalog table, and the docs badges, all Nuxt side this time.
- Deleting `src/_data/featureCatalog.yaml` and the dead transforms in `.eleventy.js`.

One open question I'll need input on: only a handful of changelog posts currently link to a catalog feature, so even after the rewire most entries won't show a badge. I'd rather fix the docs surface first (that's where the clicks were) and treat changelog backfill as separate.

Yndira-E:
[I asked Jamie to not show the badges on the latest release blog.](https://github.com/FlowFuse/website/pull/5365#issuecomment-5132343106) But it was showing them and it probably still shows it on previous release posts...

I spotted these because the newsletter, which points to changelog entries, was sent out today.

Yndira-E:
> I'd rather fix the docs surface first (that's where the clicks were) and treat changelog backfill as separate.

Sounds like a plan!

PR #5526, opened 2026-08-06
Enhance commission payment instructions in handbook

abby-pie:
Added instructions for checking contract terms and payment structures for exported deals. Included guidance on handling multi-year bookings and emphasized the importance of written agreement for commission numbers.

## Checklist


nuxt/content/handbook/operations/commission-payment.md
```diff
@@ -49,6 +49,11 @@ When the deal board is updated with on the won deals of last month,
 click "Export View" and export as CSV. Download this file
 to your machine.
 
+For each exported deal, check the executed Agreement for the contract term and the payment terms. HubSpot reports ARR annually,
```
ZJvandeWeg: @abby-flowp Isn't there a property to include instead? I don't think this the way to do it.

If there's a column on the sheet, we can automate this, but we need to export the correct property.

abby-pie: @ZJvandeWeg I'll correct this. One blocker: ARR/ACV don't populate on every deal in the export. DRV Bund was one of them, which is part of why the term value got missed. So even with the right property added, the export is only as good as what reps fill in

I've already raised it with Kasheef and asked if sales can populate those fields on future deal. I'll check whether there's an existing property that shows if a deal is paid upfront or in installments or if we need a custom one and get back on this.

I'll add the column to the sheet.

abby-pie (commented):
Updated with hubspot deal filters when exporting csv

ZJvandeWeg:
@abby-flowp There's already a handbook update published? https://flowfuse.com/handbook/operations/commission-payment/#calculating-team-commissions

Closing in case this is a duplicate

PR #5540, opened 2026-08-07
Prune stale redirects, move redirect handling to Nuxt

ZJvandeWeg:
## Summary
- Remove redirect entries from `src/redirects.njk` that are older than a year (per `git blame`) — search engines/browsers should have long since picked up the new destinations.
- Move the remaining redirects from the 11ty-generated `_redirects` file into native Nuxt `routeRules` (`nuxt/redirects.ts`), since Nuxt is now the front door for every route per the strangler-fig migration.
- Fix `/docs/user/expert/ -> /docs/user/expert/`, a no-op self-redirect accidentally introduced by a prior "update links" commit — restored to `/docs/user/assistant/ -> /docs/user/expert/`.

## Test plan
- [x] `npx tsc --noEmit` in `nuxt/` shows no new errors vs. main
- [x] `npm run dev:nuxt` and curl-tested exact-path, wildcard (`/certified-nodes/*`), and external-URL (`https://app.flowfuse.com/...`) redirects — all return correct 301s
- [x] Verified `/docs/user/assistant/` now redirects to `/docs/user/expert/`


ZJvandeWeg:
cc @Yndira-E

PR #5546, opened 2026-08-08
nuxt: Sitemap enrichment

ZJvandeWeg:


nuxt/lib/git-lastmod.mjs
```diff
@@ -0,0 +1,78 @@
...
+            currentDate = line.slice(1)
+            continue
+        }
+        if (!line || map.has(line)) continue
+        map.set(line, currentDate)
+    }
+    return map
+}
+
+function buildLastmodMap (repoRoot) {
```
ZJvandeWeg, line 50: So the trick here is to just ask `git` when each file was last updated, and map that file to the path on the website. We do this once, with one `git` invocation. This is the quickest way, costs about 2 seconds total for all pages.

nuxt/lib/git-lastmod.mjs
```diff
@@ -0,0 +1,78 @@
...
+// A NUL byte can't appear in a file path, so it safely marks a commit-date line apart
+// from the `--name-only` file lines that follow it. `%x00` asks git to emit the byte
+// into its output; the argv string itself only ever contains the ASCII text "%x00".
+const NUL = '\u0000'
+const mapCache = new Map()
+
+/**
+ * Pure parser for `git log --pretty=format:%x00%ci --name-only` output - split out from
+ * buildLastmodMap so the newest-first/first-occurrence-wins logic can be unit tested
+ * against fixture strings without shelling out to git or touching a real repo.
```
Yndira-E, line 31: @ZJvandeWeg is the plan to add the tests to this PR or in a follow up one?

nuxt/server/api/__sitemap__/content-urls.get.ts
```diff
@@ -0,0 +1,115 @@
...
+import { queryCollection } from '@nuxt/content/server'
+import { getGitLastmod } from '../../../lib/git-lastmod.mjs'
+
+// docs/handbook/changelog/blog/ebooks/whitepapers deliberately carry no `sitemap` schema
+// field in content.config.ts - @nuxtjs/sitemap's @nuxt/content integration re-splices
+// each collection's onUrl/filter as raw source text into a generated file with no closure
+// over this module's imports or helpers, which git-based lastmod needs. This route does
+// the same job (loc/lastmod/images) as a normal Nitro handler instead.
+
+const SITE_URL = 'https://flowfuse.com'
```
Yndira-E: We already define `site.url` in `nuxt.config.ts`, could we pull `SITE_URL` from there instead, so it stays a single source of truth?

Yndira-E:
I checked https://deploy-preview-5546--flowforge-website.netlify.app/sitemap.xml and the "last updated" column only shows dates for the docs. There's no `[sitemap] git log failed` warning in the build logs, but there is this warning in the [functions log](https://app.netlify.com/projects/flowforge-website/logs-and-metrics/functions/server?scope=deploypreview:5546):

`WARN [sitemap] git log failed in /var/task, lastmod will be omitted: spawnSync git ENOENT`

I did some digging with Claude and this is the diagnosis:

> `content-urls.get.ts` gets bundled into the deployed `server` function (not just run at prerender time), and `/var/task` (the Function's sandbox) has no `git` binary at all. So right now, `lastmod` for handbook/blog/ebooks/whitepapers silently never resolves whenever this route is served from the live function rather than baked into a static file at build — only `docs` gets a real date, because `updated` is computed once during docs-sync (also via `git log`, see `docs-sync.mjs:117`) and written into frontmatter, so no runtime git call is needed for it.

This reminded me that we have a field for this on the blog too, but unlike the docs, this one is set manually:

>`blog` already has a `lastUpdated` frontmatter field (`content.config.ts:117`) that the blog page itself uses for its "Updated" date and JSON-LD `dateModified` (`pages/blog/[...slug].vue:142-143`) — but the `blog` entry in `CONTENT_SOURCES` here doesn't reference it at all, so an author-provided update date gets silently dropped in favor of the git-only path instead of preferring `lastUpdated` (with git as a fallback, same as the page does with `page.lastUpdated || page.date`).
>

Would it make sense to compute `lastmod` the same way docs does, resolve it once at build/sync time and store it, rather than depending on `git` being available wherever this route ends up running?

Or, would forcing `/sitemap.xml` itself into the prerender crawl `(routeRules: { '/sitemap.xml': { prerender: true } })` fix this, since right now it looks like the whole sitemap route is being regenerated dynamically per-request rather than baked at build time (where git is available)?

nuxt/lib/git-lastmod.mjs
ZJvandeWeg: I had a test case, but the repo to test from is this one. So it's a moving target, would fail after a day or so.

nuxt/server/api/__sitemap__/content-urls.get.ts
ZJvandeWeg: Done in 7d107425d — now pulling from site.url via nuxt-site-config.

Yndira-E (approved):

PR #5584, opened 2026-08-12
Handbook: require AE to set company Lifecycle Stage to Customer on close

ZJvandeWeg:
## Summary
- When closing a deal, the AE must set the `Lifecycle Stage` on the HubSpot **company** record to `Customer`, not just the contact.

## Test plan
- [ ] Docs-only change; verify rendered handbook page for `/handbook/sales/engagements/` looks correct


MichaelBDavis:
@ZJvandeWeg We can already achieve this today for all enterprise customers (non-self serve) without any manual intervention. All accounts with one or more active subscriptions have the property 'Active subscription count' set accordingly. If you'd like, we can have the account lifecycle updated to 'customer' as well.

ZJvandeWeg:
Ok, that's a better way to do it. THanks

PR #5587, opened 2026-08-13
Include remote instances in hub feature table

knolleary:
This updates the pricing page to include edge instances in Hub - but without the more advanced features of Device Groups etc.

Rationale:

1. This is what had been documented internally in both the working doc and pricing spreadsheet following previous discussions. This PR brings the page inline with what had been agreed.
2. With trial accounts coming in at hub, *not* having edge instances available makes the edge onboarding harder for first-time users evaluating things

### DIFF
--- a/nuxt/content/feature-catalog.yml
+++ b/nuxt/content/feature-catalog.yml
@@ -21,7 +21,7 @@ sections:
       - id: edge-development
         title: Edge Development
         description: "Develop and test Node-RED flows directly on edge devices with a remote editor proxy."
-        tiers: { edge: true, hub: false, fleet: true }
+        tiers: { edge: true, hub: true, fleet: true }
       - id: private-npm-registry
         title: Custom Node-RED Nodes
         description: "Create and manage your own private npm registry for Node-RED nodes, so you can share custom nodes across your team and devices without publishing them publicly."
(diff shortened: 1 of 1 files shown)


allthedoll (approved):

nuxt/content/feature-catalog.yml
```diff
@@ -21,7 +21,7 @@ sections:
       - id: edge-development
         title: Edge Development
         description: "Develop and test Node-RED flows directly on edge devices with a remote editor proxy."
-        tiers: { edge: true, hub: false, fleet: true }
+        tiers: { edge: true, hub: true, fleet: true }
```
ZJvandeWeg, line 24: @knolleary How many? I would limit this.

nuxt/content/feature-catalog.yml
```diff
@@ -54,7 +54,7 @@ sections:
       - id: edge-devices
         title: Edge Instances
         description: "Deploy and mange your Node-RED instances on edge PLCs and gateways, with full visibility and control from the cloud."
-        tiers: { edge: true, hub: false, fleet: true }
+        tiers: { edge: true, hub: true, fleet: true }
```
ZJvandeWeg, line 57: @knolleary How many? I would limit this.

nuxt/content/feature-catalog.yml
knolleary: What had been agreed was the license includes a set number of runtimes - that could be either hosted or remote - rather than individual limits.

ZJvandeWeg: @knolleary So what's the difference between Hub and Edge?

ZJvandeWeg:
@knolleary I think a better fix for allowing users to experience edge instances is onboarding them to Edge instead of Hub. 
Is that a good middle ground here?

knolleary:
Yes, a trial at Edge level resolves that part.

The goal of this PR was to fix what I saw as a discrepancy between what had been agreed in previous discussions and what had been published to the pricing page, rather than reopen what I thought was a resolved discussion. As long as all everyone else is aware of this change, then we can close this PR.

PR #5588, opened 2026-08-13
Fix infinite redirect loops on two blog posts

Yndira-E:
## Description

- Two blog posts were rendering empty (or 404ing), with the browser reporting an infinite redirect loop:
  - `/blog/2023/05/integrating-modbus-with-node-red/`
  - `/blog/2026/03/rethinking-edge-ais-core-orchestration/`
- Root cause: PR #5540 moved legacy 11ty redirects for old, differently-cased URLs into `nuxt/redirects.ts`. Nitro's route matching is case-insensitive by default, so each of these rules also matched its own lowercase canonical URL, redirecting it to itself — an infinite loop that aborts navigation before the page ever renders.
- Removed both stale entries from `nuxt/redirects.ts` and re-added them to `netlify.toml` instead. Netlify's edge redirects match exact case, so the old capitalized URLs still 301 to the canonical lowercase URLs, without colliding with them.
- Audited the rest of `nuxt/redirects.ts` for the same pattern (a key case-insensitively equal to its own `to`, or colliding with another key) — no other instances found.

## Test plan
- [x] Visit both post URLs locally (`npm run dev`), confirm they render normally now.
- [x] Visit both post URLs in the deploy preview, confirm they render normally.
- [x] Visit the old capitalized-casing URLs in the **deploy preview** (`netlify.toml` redirects only run through Netlify's own layer, not plain `nuxt dev`, can't be checked locally without `netlify dev`, confirm they 301 to the canonical URL instead of looping or 404ing.

## Related Issue(s)

https://app.asana.com/1/1213818720452348/project/1215165115614867/task/1217418319610123?focus=true

## Checklist


sumitshinde-84 (approved):
Thank Yndira !

PR #5590, opened 2026-08-13
Changelog: group entries by release, replace pagination with scroll and search

dimitrieh:
## Description

Reframes the changelog around releases, so a reader can see what shipped in each one and self-hosted users can tell which release a change reaches them in.

Each release labels its own entries from a sticky column beside them, so the label lines up with the entries it belongs to and stays in view while you read through that release. The listing is a single page now: entries reveal as you scroll instead of paging, and search runs on the Algolia index the blog, docs and handbook already use, filtered to `category:changelog`, so it reaches entry bodies rather than just frontmatter. Each release label is an anchor for its own release, and a link to a release below the initial slice renders through to it before scrolling. Numbered `/changelog/<n>/` routes redirect back to the listing.

Entries carry a new `release` frontmatter field. It was backfilled across the archive from each entry's own date against the release milestone calendar, on the rule that an entry belongs to the first release dated on or after it, since a post goes up once the change is live on Cloud and self-hosted receives it in the next release. Nothing needs to be recorded in `featureCatalog.yaml` any more, and the handbook's frontmatter table and template are updated.

A `node:test` check validates every entry's frontmatter, including an invariant that release numbers never decrease as dates increase, which catches a mistyped release without needing a release calendar in the repo.

Also adds the `link rel=alternate` that advertises the existing Atom feed at `/changelog/index.xml`, which nothing pointed at before.

Two incidental fixes the listing exposed: entry descriptions that were empty rendered as blank rows, and one entry's date was not zero-padded.

## Related Issue(s)

No tracking issue. Came out of a Product and Marketing discussion about framing the changelog page.

## Checklist


dimitrieh:
Related slack thread: https://flowfuse.slack.com/archives/C03FAUBPPJ7/p1786557212500649

@allthedoll this includes a change to every changelog entry so that a release is noted in the front matter (therefore so many files) and an adjustment of the handbook page instructing about creating changelog entries, plus tests for it.

dimitrieh:
open question: this page now has its own rss. Do we want the merge this into the blog rss?

dimitrieh:
> make the heading of "what's new" a little larger 

adjusted

nuxt/composables/useChangelogList.ts
```diff
@@ -1,17 +1,64 @@
...
+    const entries = computed(() => (allEntries.value || []) as unknown as ChangelogEntry[])
 
-    const entries = computed(() => {
-        const start = (pageNumber - 1) * CHANGELOG_PAGE_SIZE
-        return (allEntries.value || []).slice(start, start + CHANGELOG_PAGE_SIZE)
+    const matching = computed(() => {
+        const term = search.value.trim().toLowerCase()
+        if (!term) return entries.value
+        return entries.value.filter(e =>
+            e.title?.toLowerCase().includes(term) || e.description?.toLowerCase().includes(term))
```
Yndira-E: @dimitrieh the Algolia search was already implemented before migrating to Nuxt, iirc, ZJ intentionally removed it. If we want to bring the search back, could we use Algolia like the rest of the site instead? That way we don't lose the full-text search, fuzzy matching and keep it consistent across the site.

nuxt/components/ChangelogListing.vue
```diff
@@ -1,30 +1,113 @@
...
+        Nothing matches "{{ search }}".
+        <button class="text-indigo-600 hover:underline" @click="search = ''">Clear the search</button>
+      </p>
+
+      <!-- The release label lives in a column beside its own entries rather than in a
+           separate rail, so it lines up with the entries it belongs to, and sticks while
+           you read through them. Two columns per release, not one list plus one nav. -->
+      <section
+        v-for="group in visibleGroups"
+        :id="anchorId(group.release)"
```
Yndira-E, line 106: While going through the deploy preview, I was expecting these to be interactive, actual shareable anchors. Was it deliberate to make them static? Is it related to the infinite scroll choice?

nuxt/composables/useChangelogList.ts
dimitrieh: Switched to `<AlgoliaSearch index-filter="category:changelog" />`, the same component the blog, docs and handbook use.

Checked the index before changing it: `/changelog/` is already in `ALLOWED_ROOT_SCOPES` in `lib/search-index.js`, so every entry is in `prod_netlify` with its body text under `category:changelog`. My filter only matched title and description, so this reaches the bodies too and tolerates typos.

nuxt/components/ChangelogListing.vue
dimitrieh: Not deliberate. The ids were there but nothing linked to them. The sticky label and the mobile heading are both anchors now.

It was related to the scroll choice though. A link like `/changelog/#release-2-20` had nothing to scroll to on arrival, since only the newest entries are rendered. The listing now renders through that release first, then scrolls to it. Tested back to `#release-1-12`, the oldest one.

Yndira-E: Tested this on the deploy preview with `/changelog/#release-1-12`, loading it fresh doesn't scroll anywhere, it stays at the top (newest release).

dimitrieh: You were right, cold load was broken. My test only covered a page that was already loaded, which is why it passed.

At hydration `window.location.hash` is still empty. Nuxt keeps the hash on the route and restores the address bar only once the initial navigation settles, so `revealFromHash` gave up before it found the release, and nothing changed afterwards to make it retry.

One line: read whichever of the two is populated.

Verified on a local production build. Cold `#release-1-12` and `#release-2-20` both render through and land on the release, no hash is unchanged, and label clicks, in-page hash changes and browser back all still work.

Could you re-test on the preview once it redeploys?

dimitrieh:
@Yndira-E up for review again

Yndira-E (approved):

PR #5601, opened 2026-08-14
Cancel superseded test runs

dimitrieh:
## Description

`on: push` with no concurrency group means a branch pushed to repeatedly runs a full site build per commit, and none of the obsolete ones are cancelled. Over the last two weeks of runs, roughly a fifth of all workflow minutes went to builds whose commit had already been replaced.

Group by ref and cancel the outdated run. `main` is exempt, so every commit there still gets its own run for the ruleset and for history.

This does not make any single build faster.

## Related Issue(s)

None.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,6 +2,13 @@ name: Tests
 
 on: push
 
+# Pushing again to a branch replaces the commit the running build was checking, so cancel
+# that build rather than pay for a result nobody will look at. main is exempt: every commit
+# there keeps its own run, since those are the ones the ruleset and the history rely on.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
 jobs:
   # The build itself lives in FlowFuse/github-actions-workflows so this repository and
   # FlowFuse/flowfuse documentation pull requests share one definition. The check this


dimitrieh:
See https://flowfuse.slack.com/archives/C032Q63FGG1/p1786707866678429?thread_ts=1786703349.609919&cid=C032Q63FGG1

.github/workflows/test.yml
```diff
@@ -2,6 +2,13 @@ name: Tests
 
 on: push
 
+# Pushing again to a branch replaces the commit the running build was checking, so cancel
+# that build rather than pay for a result nobody will look at. main is exempt: every commit
+# there keeps its own run, since those are the ones the ruleset and the history rely on.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
```
ppawlowski, line 10: Just for better understanding - why do we not want to cancel the old production build if a new change is merged into the `main` branch?

ZJvandeWeg: @ppawlowski There's times in the month where we have a merge 'hotspot', and we merge within 20 minutes of one another. That does create a situation where if we do so well timed, we won't deploy any change in an hour or so, just as we cancel a deploy just before it is done with a build.

ppawlowski: Fair enough.

ppawlowski (approved):

PR #5602, opened 2026-08-14
Blog: restore features lost on the Nuxt migration

Yndira-E:
## Description

Restores a few things that got lost when the blog migrated from 11ty to Nuxt, and fixes some post-migration bugs:
- **Related/Recommended Articles** — restored the sidebar list linking to related posts (by shared tags) or, as a fallback, recent posts. This is a meaningful interlinking mechanism between blog pages and shouldn't have been dropped.
- **"Updated on" date** — restored on the byline when a post has a `lastUpdated` value.
- **Author avatar** — added a circular headshot next to the byline (indigo-300 background), matching how presenters are shown on webinar pages. Falls back to the FlowFuse icon for legacy author slugs that have no data file in `team`/`guests`, same as the old Eleventy `renderTeamMember` behavior.
- **Default CTA changed from `contact` to `sign-up`** — sign-up is the lowest-friction action, which makes sense as a safe default for now. Longer-term we should be more intentional and pick the CTA (and accompanying copy) based on the post's content rather than relying on a single default for everything.
- **Responsive video embeds** — posts with a raw `<lite-youtube>` embed in their markdown body had a fixed inline `width`, so the video didn't shrink on small screens. Fixed with a CSS override.
- **Reading time always showing "1 min read"** — `extractText` assumed a hast-like `{type, children}` body shape; `@nuxt/content` v3 actually returns a minimark tree (nodes are either strings or `[tag, attrs, ...children]`). Rewrote the walker for the real shape.

## Checklist

### DIFF
--- a/nuxt/components/BlogPostCta.vue
+++ b/nuxt/components/BlogPostCta.vue
@@ -27,13 +27,12 @@ const CTA_VARIANTS: Record<string, { title: string, description: string }> = {
     },
 }
 
-// Missing type defaults to 'contact'; an unrecognised one (e.g. typo'd
-// `type: signup`) falls back to 'sign-up' - same two fallbacks as before.
+// Missing type, or an unrecognised one (e.g. typo'd `type: signup`), falls
+// back to 'sign-up'.
 const KNOWN_TYPES = new Set(['sign-up', 'demo', 'contact', 'pricing'])
 const ctaType = computed(() => {
     const type = props.cta?.type
-    if (!type) return 'contact'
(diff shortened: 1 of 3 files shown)


sumitshinde-84:
Thanks @Yndira-E! Everything is perfect except the Related/Recommended Articles section isn’t working

Yndira-E:
Thanks for flagging @sumitshinde-84

A typical case of "it worked on local", we'll see if it's fixed when the preview is available

sumitshinde-84 (approved):

PR #5605, opened 2026-08-14
Draft ROI calculator page

KristopherLeads:
Add ROI calculator page with input fields and calculations for savings categories.

## Checklist


KristopherLeads:
Pending review and iteration.

Yndira-E:
@KristopherLeads tests are not passing, could you please check?

nuxt/pages/roi-calculator.vue
Yndira-E, line 1: @KristopherLeads looking at the redirects I'm assuming this file is no longer needed, right?

nuxt/redirects.ts
```diff
@@ -144,6 +144,8 @@ export const redirects: Record<string, NitroRouteRules> = {
     '/handbook/development/releases': { redirect: { to: '/handbook/engineering/releases/', statusCode: 301 } },
     '/handbook/development/releases/': { redirect: { to: '/handbook/engineering/releases/', statusCode: 301 } },
     '/handbook/peopleops/job-descriptions/dashboard-engineer/': { redirect: { to: '/handbook/peopleops/job-descriptions/', statusCode: 301 } },
+    '/roi-calculator/': { redirect: { to: '/pricing/roi-calculator/', statusCode: 301 } },
+    '/roi-calculator': { redirect: { to: '/pricing/roi-calculator/', statusCode: 301 } },
```
Yndira-E: @KristopherLeads since this page hasn't been live before, the redirects are not needed

```suggestion
```

src/_data/chrome.json
```diff
@@ -42,6 +42,12 @@
                                 "label": "Pricing",
                                 "href": "/pricing/",
                                 "icon": "dollar"
+                            },
+                            {
+                                "label": "ROI Calculator",
+                                "href": "/pricing/roi-calculator/",
+                                "icon": "arrow-trending-up",
+                                "indent": true
```
Yndira-E: An ROI calculator isn’t a platform component or a pricing option, it’s an evaluation asset. I think it would fit better under Resources > Customers, alongside Customer Stories. It’s also already embedded on the Pricing page, which is one of our highest-traffic pages, with a link through to the full calculator, so moving it wouldn’t mean losing discoverability.

KristopherLeads:
> @KristopherLeads tests are not passing, could you please check?

I am looking into this right now.

src/_data/chrome.json
KristopherLeads: This was discussed with Joyce - the intent was to have it as a standalone component that could be shared as a decision-making element. Having it under customer stories imo doesn't make sense as you're not yet a customer.

Yndira-E:
It still needs a full style review, some things don't have enough contrast, or are duplicating patterns or classes. I'll take care of those things and push the changes when it's done.

src/_data/chrome.json
Yndira-E: I think we might be talking about two different things here:

I wasn’t suggesting placing the ROI calculator inside Customer Stories or implying that it’s a resource for existing customers. I meant grouping it in the same navigation section. Customer Stories themselves aren’t meant for existing customers either — they’re resources for prospects evaluating FlowFuse.

Also, moving it to a different section of the navigation wouldn’t make it any less of a standalone component. It would still have its own dedicated page and remain fully shareable as a decision-making tool. My point was purely about whether `Platform >  Overview > Pricing > ROI calculator` is the right place for it in the main nav.

Screenshot 2026-09-02 at 15 16 59

Yndira-E (approved):

Yndira-E:
Style-wise, this should be ready for merging now.

PR #5610, opened 2026-08-17
handbook: raise coworking space allowance to $250 a month

dimitrieh:
## Description

Raises the coworking space allowance from $150 to $400 a month. Coworking desks
across Europe cost well above $150, and because the allowance is set in USD, part
of it is lost again to the USD to EUR conversion.

## Related Issue(s)

None.

## Checklist

### DIFF
--- a/nuxt/content/handbook/peopleops/expenses.md
+++ b/nuxt/content/handbook/peopleops/expenses.md
@@ -250,7 +250,7 @@ before making such arrangements.
 ### Coworking space allowance
 
 If you want to occasionally use a separate workspace to your home, the company
-will cover up to $150 a month in charges for renting a desk in a co-working
+will cover up to $400 a month in charges for renting a desk in a co-working
 space after manager approval.
 
 ### Mileage
--- a/nuxt/content/handbook/peopleops/expenses.md
+++ b/nuxt/content/handbook/peopleops/expenses.md
@@ -250,7 +250,7 @@ before making such arrangements.
 ### Coworking space allowance
 
 If you want to occasionally use a separate workspace to your home, the company
-will cover up to $400 a month in charges for renting a desk in a co-working
+will cover up to $250 a month in charges for renting a desk in a co-working
 space after manager approval.
 
 ### Mileage


nuxt/content/handbook/peopleops/expenses.md
```diff
@@ -250,7 +250,7 @@ before making such arrangements.
 ### Coworking space allowance
 
 If you want to occasionally use a separate workspace to your home, the company
-will cover up to $150 a month in charges for renting a desk in a co-working
+will cover up to $400 a month in charges for renting a desk in a co-working
```
ZJvandeWeg: @dimitrieh Can you include more details on why $400? Why does this serve you better?

It also feels like it's enough to have a full time office, and that seems excessive?

dimitrieh: @ZJvandeWeg I raised this on request of @allthedoll as so it can thereafter be discussed.
The proposal I put forward here covers my current co-working give or take a few eur. I rounded to a whole number in usd.

From the description:

> Coworking desks across Europe cost well above $150, and because the allowance is set in USD, part
of it is lost again to the USD to EUR conversion.

A quick search with Claude results in the following if taking Amsterdam as an example. I think this falls a little lower in reality, but not by much. Note the conversion to USD.

CleanShot 2026-08-17 at 20 05 03@2x

PS: My coworking is a hot desk seat (and not even for a full week).

ZJvandeWeg: @dimitrieh I paid Euro 150 per month for 3 days of access, hot desk too. I'm conflicted about the delta and think it's rather large. So not sure.

To move forward I suggest we up the number to $250/mo and move on.

dimitrieh: @ZJvandeWeg 

```suggestion
will cover up to $250 a month in charges for renting a desk in a co-working
```

👍. 250 usd will cover my co-working for 63%. I think depending on where you are, the costs can be higher or lower. Some places in europe are just more expensive, plus you have the 14% difference in usd-to-eur conversion.

PR #5551, opened 2026-08-09
Drop docs redirect stubs from sidebar nav, redirect renamed doc paths

ZJvandeWeg:
## Summary
- Depends on flowfuse/flowfuse#8154, which renames three underscore-named doc files (`email_providers.md`, `user_management.md`, `aws_terraform.md`) to fix nuxt-link-checker's `no-underscores` warning.
- `nuxt/lib/docs-nav.mjs` now excludes any docs page with `redirect` frontmatter (e.g. `admin/licensing.md`, `community-support.md`) from the sidebar entirely — linking to them just bounces through a redirect, and since the sidebar renders on every docs page, that warning was repeating ~194 times in link-checker output.
- `nuxt/redirects.ts` gets 301s for the three renamed doc paths so existing bookmarks/links don't 404 once flowfuse/flowfuse#8154 ships.
- `.claude/CLAUDE.md` documents that redirects belong in `nuxt/redirects.ts` (Nitro route rules), not `netlify.toml` (legacy, reserved for domain-alias redirects).

## Test plan
- [ ] `npm run build:nuxt` locally and confirm `/docs/admin/licensing/` and `/docs/community-support/` no longer appear in the docs sidebar
- [ ] Confirm the sidebar still renders correctly for a normal section (e.g. Admin, Support)
- [ ] Once flowfuse/flowfuse#8154 merges and this branch is rebuilt, confirm `/docs/install/email_providers/`, `/docs/admin/user_management/`, `/docs/install/kubernetes/aws_terraform/` 301 to their new paths

PR #5617, opened 2026-08-17
Make docs section URLs actually redirect

dimitrieh:
## Description

Docs section URLs do not redirect. `/docs/install/` answers 200 with a `<meta http-equiv="refresh">` page instead of a 301. A reader still lands on the right page. A crawler or an inbound link gets nothing, because no redirect status is ever sent. This is live on flowfuse.com today.

**Why it happens.** The docs page component redirects while rendering. The prerenderer writes that render out as a static HTML file. Netlify serves a matching static file before it consults a redirect rule, so the rule never runs.

**What this changes.** `docs-source` now reads the `layout: redirect` frontmatter at build time and turns it into a Nitro route rule. Those routes also stay out of the prerender list, so no stub file is written that could win over the rule. The netlify preset compiles the rules into `dist/_redirects`, and they resolve at the edge.

The prerender filter matches any rule that redirects, not only the generated ones. `redirects.ts` maps a few underscored docs slugs to hyphenated names, and a docs tree still carrying the old filenames used to prerender a stub that beat the rule.

The URLs this affects:

| URL | Redirects to |
|---|---|
| `/docs/user/` | `/docs/user/introduction` |
| `/docs/cloud/` | `/docs/cloud/introduction` |
| `/docs/admin/` | `/docs/admin/introduction` |
| `/docs/install/` | `/docs/install/introduction` |
| `/docs/hardware/` | `/docs/hardware/introduction` |
| `/docs/migration/` | `/docs/migration/introduction` |
| `/docs/contribute/` | `/docs/contribute/introduction` |
| `/docs/device-agent/` | `/docs/device-agent/introduction` |
| `/docs/device-agent/install/` | `/docs/device-agent/install/overview` |
| `/docs/admin/hardening/` | `/docs/admin/hardening/kubernetes/` |
| `/docs/admin/licensing/` | `/pricing/` |
| `/docs/community-support/` | Node-RED forum |

**The stub pages have to stay.** Deleting them is not an alternative fix. There is no Netlify rule for any of these paths, so the frontmatter is the only place the redirect is declared. The same frontmatter is also the docs nav definition: each stub carries the `navGroup`, `navGroupOrder`, `navTitle` and `navOrder` its whole section is grouped and labelled by, and the leaf pages do not. #5551 dropped the stubs from the sidebar and d1a2528ef had to put that metadata back.

## Related Issue(s)

None.

## Checklist

### DIFF
--- a/nuxt/lib/docs-redirects.mjs
+++ b/nuxt/lib/docs-redirects.mjs
@@ -1,0 +1,105 @@
+// Turns the `layout: redirect` pages synced from FlowFuse/flowfuse into Nitro route
+// rules. Kept free of Nuxt and filesystem imports so it can be unit tested with
+// `node --test`; modules/docs-source.ts is the thin filesystem wrapper around this.
+//
+// Before this existed the pages redirected from inside the docs page component, which
+// the prerenderer turns into an HTML file holding a `<meta http-equiv="refresh">`. That
+// answers the URL with a 200 and no redirect status at all, so an inbound link to
+// /docs/install/ passes nothing on to /docs/install/introduction. A route rule gives
+// Netlify a real 301, the same way nuxt/redirects.ts does for the retired 11ty paths.
+
+// js-yaml rather than yaml: it is the declared root dependency, and `node --test`
+// resolves it as raw ESM without Vite's CommonJS interop in the way. Same reasoning as
+// feature-catalog.test.mjs.
+import jsYaml from 'js-yaml'
+
+// Only the leading block, so a `---` used as a horizontal rule in the body is not read as
+// the closing fence. Docs synced from another repo may end at `---` with no trailing
+// newline, hence the `$` alternative.
+const FRONTMATTER = /^---[ \t]*\r?\n([\s\S]*?)\r?\n?---[ \t]*(?:\r?\n|$)/
+
+function frontmatterOf (source) {
+    const block = FRONTMATTER.exec(source ?? '')
+    if (!block) return null
+
+    let data
+    try {
+        data = jsYaml.load(block[1])
+    } catch {
+        // The docs are authored in another repo, so one malformed page must not fail the
+        // build. It keeps prerendering as an ordinary page instead.
+        return null
+    }
+
+    return data && typeof data === 'object' && !Array.isArray(data) ? data : null
+}
+
+/**
+ * @param {Array<{route: string, source: string}>} pages route docs-source would prerender,
+ *   plus the raw markdown behind it
+ * @returns {Record<string, {redirect: {to: string, statusCode: number}}>} keyed by route,
+ *   ready to spread into `nuxt.options.routeRules`
+ */
+export function docsRedirectRules (pages) {
+    const rules = {}
+
+    for (const { route, source } of pages) {
+        const data = frontmatterOf(source)
+
+        // Both keys, the same pair the page component required: a `to` on a page that
+        // still renders content of its own is not a redirect.
+        if (data?.layout !== 'redirect') continue
+
+        const to = data.redirect?.to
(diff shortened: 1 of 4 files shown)


ZJvandeWeg:
He @dimitrieh, if we remove the frontmatter, would that also resolve this issue? As we have a redirect on Netlify in that case, we can just remove those files altogether?

dimitrieh:
> if we remove the frontmatter, would that also resolve this issue?

No. That frontmatter is also what builds the docs sidebar. Each of these stub pages carries navGroup, navGroupOrder, navTitle and navOrder. The leaf pages don't: docs/user/introduction.md only has navTitle and navOrder. So docs/user/README.md is the only thing saying /docs/user belongs under "FlowFuse User Manuals", first in the list.

We tried this already. #5551 dropped the stubs from the sidebar. d1a2528ef had to put the metadata back, because sections fell into "Other", got named after their raw path ("user", "admin"), and three groups disappeared.

> As we have a redirect on Netlify in that case, we can just remove those files altogether?

There is no Netlify redirect for these paths. I checked every one. None is in nuxt/redirects.ts, and netlify.toml only has the flowforge.com to flowfuse.com rules. The frontmatter is the only place these redirects exist. Delete the files and the URLs 404.

Regardless, the bug is still live. Every one of those paths answers 200 with a meta refresh on flowfuse.com right now. Rebased on main and rewrote the description, so this is ready for another look.

dimitrieh:
## Before and after

Probed both sides with the same command on 2026-09-10, after the rebase onto main:

```bash
curl -s -o /dev/null -w '%{http_code} %{redirect_url}' <host><path>
```

`Before` is live flowfuse.com. `After` is the deploy preview for this branch. Every path targets the same place on both sides, so the target is listed once.

| URL | Before | After | Target |
|---|---|---|---|
| `/docs/user/` | 200, meta refresh | **301** | `/docs/user/introduction` |
| `/docs/cloud/` | 200, meta refresh | **301** | `/docs/cloud/introduction` |
| `/docs/admin/` | 200, meta refresh | **301** | `/docs/admin/introduction` |
| `/docs/install/` | 200, meta refresh | **301** | `/docs/install/introduction` |
| `/docs/hardware/` | 200, meta refresh | **301** | `/docs/hardware/introduction` |
| `/docs/migration/` | 200, meta refresh | **301** | `/docs/migration/introduction` |
| `/docs/contribute/` | 200, meta refresh | **301** | `/docs/contribute/introduction` |
| `/docs/device-agent/` | 200, meta refresh | **301** | `/docs/device-agent/introduction` |
| `/docs/device-agent/install/` | 200, meta refresh | **301** | `/docs/device-agent/install/overview` |
| `/docs/admin/hardening/` | 200, meta refresh | **301** | `/docs/admin/hardening/kubernetes/` |
| `/docs/admin/licensing/` | 200, meta refresh | **301** | `/pricing/` |
| `/docs/community-support/` | 200, meta refresh | **301** | Node-RED forum (off-site) |

Before, no path sent a `Location` header at all. The redirect only existed as `<meta http-equiv="refresh" content="0; url=...">` inside a 200 response. After, every path sends a real `Location` and no HTML body.

## Sidebar regression check

The same stub pages define the docs nav, so I checked the sidebar still builds on the preview at `/docs/user/introduction/`. All group headings render (FlowFuse User Manuals, FlowFuse Cloud, FlowFuse Self-Hosted, Device Agent, Contributing, Support) and no section fell into an `Other` group.

## Checks

`test_website / Build and check` passes, including the link checker. The new `nuxt/lib/docs-redirects.test.mjs` covers the frontmatter parsing and the prerender filter, and passes.

KristopherLeads (approved):

PR #5627, opened 2026-08-18
prose-pre: render the filename and highlighted lines a fence asks for

dimitrieh:
## TL;DR

Code blocks in docs and the handbook can now say which file they are, and point at the lines that matter. Writing this in a markdown fence:

````
```js [server.js]{2,4-6}
````

gives the reader a `server.js` label on top of the block and a highlight band across lines 2, 4, 5 and 6. Both of those were already being parsed out of the fence and then thrown away, so this is turning on something the stack was doing anyway.

Nothing changes on the live site the moment this merges: no fence in the synced docs uses the syntax yet. It becomes visible when docs start using it, which is a follow-up in FlowFuse/flowfuse.

## Description

`nuxt/components/prose/ProsePre.vue` declared `filename` and `highlights` and rendered neither, so a fence like ```` ```js [server.js]{2,4-6} ```` silently lost both the file name and the called-out lines, everywhere in docs and the handbook.

- `filename` now renders as a header flush on top of the block, in the same slate tones as the copy button. The header takes the block's top margin and the block squares off its top corners, so the two read as one object. The copy button lands inside the header when there is one.
- Highlighted lines needed no new markup. The parser already expands `{2,4-6}` into line numbers and the shiki highlighter already tags those lines `highlight`, so this is a band plus a left accent rule.

A block with neither prop renders exactly as before, and the mermaid branch is untouched.

Checked in the Nuxt dev server against fences covering: filename only, a single highlighted line, a highlighted range, both props together, a plain fence, a fence with highlights and no language, and a mermaid fence. Copy button and mermaid rendering both re-verified.

## Related Issue(s)

None.

## Checklist

### DIFF
--- a/nuxt/components/prose/ProsePre.vue
+++ b/nuxt/components/prose/ProsePre.vue
@@ -67,7 +67,23 @@ async function copyCode () {
 
 <template>
   <div v-if="language === 'mermaid'" ref="mermaidRef" class="mermaid-diagram overflow-x-auto py-4" />
-  <div v-else class="code-block group relative">
+  <div
+    v-else
+    class="code-block group relative"
+    :class="{ 'code-block--titled': filename, 'code-block--highlighted': highlights?.length }"
+  >
+    <!--
+      A fence can name the file it shows (```js [server.js]). The header sits flush on
+      top of the block, in the same slate tones as the copy button, and takes the
+      block's top margin so the two read as one object. It leaves room on the right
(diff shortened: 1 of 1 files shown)


dimitrieh:
Companion PR created that makes use of this: https://github.com/FlowFuse/flowfuse/pull/8216

KristopherLeads (approved):

PR #5631, opened 2026-08-18
Fix: restore full-width prose images on blog, changelog, stories & whitepaper

Yndira-E:
## Description

- `.prose img` in `src/css/style.css` only had `max-width: 100%`, so images render at their native pixel size instead of filling the prose column.
- Before the changelog's move to Nuxt, 11ty's `eleventy-img` pipeline resized every markdown image to the prose column's width (650px), which is why images always looked full-width. The new `nuxt/components/content/ProseImg.vue` just passes the image through with no resizing, so screenshots narrower than the column now render smaller than expected.
- Added `.prose:not(.prose-natural-size-images) img { width: 100%; height: auto; }` to restore the fill-the-column look for blog, changelog, customer-stories, and whitepaper pages.
- Added the new `prose-natural-size-images` class to handbook, docs, and the `/node-red/**` layout (`layouts/documentation.njk`), since those keep inline screenshots at their natural size.

## Checklist

### DIFF
--- a/src/_includes/layouts/documentation.njk
+++ b/src/_includes/layouts/documentation.njk
@@ -29,7 +29,7 @@ date: git Last Modified
         <div class="w-full">
             <!-- Main Content -->
             <div class="order-last md:order-first">
-                <div class="mt-6 mb-4 prose prose-blue main-content">
+                <div class="mt-6 mb-4 prose prose-blue main-content prose-natural-size-images">
                     {{ content | rewriteHandbookLinks(page) | safe }}
                 </div>
             </div>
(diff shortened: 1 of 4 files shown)


src/_includes/layouts/documentation.njk
```diff
@@ -29,7 +29,7 @@ date: git Last Modified
         <div class="w-full">
             <!-- Main Content -->
             <div class="order-last md:order-first">
-                <div class="mt-6 mb-4 prose prose-blue main-content">
+                <div class="mt-6 mb-4 prose prose-blue main-content prose-natural-size-images">
```
ZJvandeWeg, line 32: @Yndira-E Where's this file still in use?

ZJvandeWeg: It is, thanks

PR #5637, opened 2026-08-18
nuxt: disable nitro server sourcemaps

dimitrieh:
## Description

Sets `nitro.sourceMap: false` in `nuxt/nuxt.config.ts`. Nitro's default is `true`, which emits a `.mjs.map` next to every server chunk. On run 32107141754 that was 186 map files in the Netlify functions bundle.

The win is marginal and worth stating plainly: it removes those files from the output and takes a few seconds off the nitro server build, which was 21.6s of a 24m15s job. It is not a meaningful change to overall build time.

The cost: sourcemaps are what make a server-side stack trace in the Netlify function logs readable. Without them, runtime errors in the SSR fallback are harder to trace back to source. If you would rather keep that debuggability, closing this PR is the right call and costs nothing.

## Related Issue(s)

None

## Checklist


dimitrieh:
I think these are mostly relevant when developing locally, not for when previewing builds on netlify 🤔

PR #5649, opened 2026-08-20
Fix PostHog pageview capture on Nuxt route changes

Yndira-E:
## Description

- Since the blog/changelog/docs migration to Nuxt, client-side route changes (Vue Router) weren't being captured as `$pageview` events in PostHog — only the initial full page load was. This silently dropped roughly a quarter of site-wide pageviews and over half of the docs section's, starting July 30.
- Root cause: the PostHog snippet in `src/_includes/analytics/body.html` is injected once per document load via `nuxt/server/plugins/analytics.ts`, and no route hook re-fired `$pageview` on SPA navigation.
- Fix: enable `capture_pageview: 'history_change'` in `posthog.init()`, PostHog's documented approach for SPA frameworks (Vue included) — it listens to the browser History API and captures navigation automatically, without duplicating capture logic per migrated section.

Confirmed via HogQL query against production PostHog data: pageviews-per-session on `/docs/**` dropped from ~2.5–3.5 to ~1.0–1.5 starting July 30, consistent with SPA navigations no longer being tracked.

Tests:
- [x] Confirm this doesn't change injection gating/behavior for 11ty-served pages — diff scoped to a single line inside `body.html`'s `posthog.init()` config; `nuxt/server/plugins/analytics.ts` untouched
- [x] Confirm the config change matches PostHog's documented approach for SPA pageview tracking (https://posthog.com/tutorials/single-page-app-pageviews)
- [x] Verified end-to-end against a real PostHog project (Test), temporarily bypassing the `CONTEXT === 'production'` gate: navigated `/blog` → post, and `/docs` → `/docs/quick-start/` → `/docs/install/docker`. PostHog recorded exactly one `$pageview` per URL, no duplicates, confirmed both via debug console and the PostHog UI. Each navigation was verified as client-side (not a hard reload) via a `window` marker that survived across navigations.

## Checklist

### DIFF
--- a/src/_includes/analytics/body.html
+++ b/src/_includes/analytics/body.html
@@ -46,6 +46,7 @@
     posthog.init('{{ POSTHOG_APIKEY }}', {
         api_host: 'https://eu.i.posthog.com',
         person_profiles: 'always',
+        capture_pageview: 'history_change',
         capture_pageleave: false,
         disable_surveys: true
     });


dimitrieh (commented):
Checked the mechanism rather than just the diff, and it holds up.

The one line lands in an 11ty include but the broken pages are Nuxt, so the part worth confirming is that it reaches them: `nuxt.config.ts` maps `serverAssets` baseName `analytics` to `../../src/_includes/analytics`, and the Nitro plugin reads that same `body.html` via `useStorage('assets:analytics')`. One file, both renderers. And in posthog-js 1.418.5 (what `array.js` serves right now) `historyAutocapture` wraps `pushState`/`replaceState` plus `popstate`, while the initial pageview still fires on load, so there's no duplicate on hydration.

Four things it doesn't cover. None are blockers for merging:

- `$title` will be stale on route changes. PostHog captures synchronously inside the pushState wrapper, `useHead` sets `document.title` a tick later. A Vue Router `afterEach` hook wouldn't have this. Worth a check if we break pageviews down by title anywhere.
- Dedupe is on `location.pathname` only, so query-string-only navigations stay uncounted.
- The `$pageleave` scroll-depth block right above this still fires once per document, and `MAX_CONTENT_VIEWED` never resets across routes. So `content_viewed` and `content_scrolled` stay wrong for the Nuxt sections, and will now diverge visibly from pageview counts. Separate follow-up rather than something for this PR.
- Nothing is recoverable for 30 Jul through merge date, and merge day adds a second discontinuity: pageviews step back up and will read as a traffic spike in marketing reporting. Worth a PostHog annotation on the merge date.

One practical note: the Nitro plugin bails unless `CONTEXT === 'production'`, so the deploy preview can't demonstrate this. After merge, re-running the pageviews-per-session query on `/docs/**` should put it back at roughly 2.5 to 3.5.

dimitrieh:
For consideration/context if following up is needed. I'd say we're good for now :)

PR #5662, opened 2026-08-21
Device Agent page: the install commands need the run step

dimitrieh:
## Description

commands on the marketing page were missing part of the command.

## Related Issue(s)

None.

## Checklist

Notes on the checklist: no performance impact, this is two strings and a sentence. No test was added because the repo has no test that asserts page copy; the check I ran was to decode both `<code>` blocks and diff them against the quickstart doc. Docs need no update, the docs were already correct and are the source these strings were copied from.

## Follow-up, not in this PR

The GIF beside the commands (`src/images/device-agent/terminal-install.gif`) opens on `node index.js --dir /tmp/device-agent-demo -p 1881`, a command run from a source checkout that no visitor would type. The rest of it correctly demonstrates the agent's interactive registration prompts. Re-recording it needs a real machine and a real registration, so it is worth its own change.


dimitrieh:
Tentatively merging this, as this is a clear error. @knolleary assigned you as reviewer regardless.

PR #5668, opened 2026-08-24
Add breadcrumbs to blog, changelog, and customer stories

sumitshinde-84:
Docs, handbook, and application-guide already rendered a visual breadcrumb trail with matching BreadcrumbList schema via the shared Breadcrumbs component. Blog, changelog, and customer stories had neither, despite being some of the most active content on the site. Wires all three into the same component for consistent navigation and structured data.

## Checklist


Yndira-E:
@sumitshinde-84 The breadcrumbs look odd in their current position, and they’re redundant with the “back to blog/changelog/stories” link since they already provide navigation back to the parent page.

Could you replace the “back to…” link with the breadcrumbs, keeping them in the same position as the “back to…” link was before? I think that would give us the navigation context without having two links serving the same purpose.

PR #5681, opened 2026-08-26
/ai: rebuild on Nuxt, with a shared command block

dimitrieh:
## Description

Rebuilds `/ai` on Nuxt and deletes `src/ai.njk`, so one renderer owns the route. Same page, same copy, same classes.

Opened for discussion rather than to merge as-is. It overlaps https://github.com/FlowFuse/website/pull/5670, which rebuilds the same page on 11ty and is where the content is being reviewed. Whichever approach we take, the other closes.

The part worth keeping either way is `nuxt/components/content/FfCommand.vue`. The copyable command block is currently markup plus a copy helper hand-pasted per page, and putting it in `components/content/` means markdown can use it too. https://github.com/FlowFuse/website/pull/5682 is stacked on this and does that for the changelog entry.

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/8192

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5692, opened 2026-08-27
/ai: take a self-hosted address in the command block

dimitrieh:
## Description

Stacked on https://github.com/FlowFuse/website/pull/5681. Base is that branch, not `main`.

The MCP address on `/ai` is the Cloud one. A self-hosted platform answers on its own
domain, so a self-hosted reader copies an address that is not theirs and has to know
to edit it. The docs say "substitute your own platform address" in prose; this makes
the block do it.

`FfCommand` gets an opt-in `host-swap`. Closed, it is one line of link text under the
block. Opened, it takes a domain and the address shown and copied becomes
`https://<that domain>/mcp`. It is shared state, so the address survives switching
agent tabs, and it is remembered per browser. The copy event now carries
`self_hosted`, so we can see which of the two people actually copy.

Not in here: swapping the two `/docs/user/expert/` links to the new guide. That waits
for https://github.com/FlowFuse/flowfuse/pull/8241, because the link checker fails on
a docs page that is not live yet.

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/8192

## Checklist


dimitrieh:
CleanShot 2026-08-27 at 14 36 59

@KristopherLeads

KristopherLeads (approved):

PR #5694, opened 2026-08-27
Pull blueprints at build time; drop the live branch deploy path

ZJvandeWeg:
## Summary

Blueprints were previously fetched by the `Build Site` Actions workflow, committed onto a force-pushed `live` branch, and Netlify watched that branch for production deploys. Product docs already resolve at build time (`nuxt/lib/docs-sync.mjs`); this applies the same pattern to blueprints so Netlify can build straight from `main`, and the `live` branch can go away.

This ships as two logical steps:

1. **Pull blueprints at build time.** `nuxt/lib/blueprints-sync.mjs` resolves `FlowFuse/blueprint-library` the same local → sibling → clone precedence as docs, writing into `src/blueprints`. `blueprint-library` is private (unlike the public `flowfuse/flowfuse` docs source), so the clone step authenticates with a GitHub App installation token (`nuxt/lib/github-app-token.mjs`, via `@octokit/auth-app`) minted from the existing `GH_BOT_APP_ID`/`GH_BOT_APP_KEY` credentials, instead of cloning anonymously.
2. **Remove the `live` branch requirement.** `.github/workflows/build.yml` no longer checks out `blueprint-library`, commits, or force-pushes to `live` — it's now a thin trigger that hits a Netlify build hook. It's kept (not deleted) because `flowfuse/flowfuse`'s `Publish Documentation` workflow dispatches it **by name** after a docs PR merges to trigger a website rebuild, and it also runs on the existing cron schedule to pick up `blueprint-library` changes — neither of those pushes anything to this repo, so Netlify wouldn't otherwise notice.

## Manual setup required before `live` can be retired

This PR is the code side only. None of it takes effect in production until the following is done, in order:

1. **Add Netlify env vars** `GH_BOT_APP_ID` and `GH_BOT_APP_KEY` — same values as the existing GitHub Actions secrets of the same name. This is what lets Netlify's build clone the private `blueprint-library` repo.
2. DONE ZJ:  Add the build hook Netlify URL as a `NETLIFY_BUILD_HOOK` secret in this GitHub repo
3. **Switch Netlify's production branch** from `live` to `main` in site settings. Post-merge this is
5. **Verify a production deploy from `main`** actually renders blueprints correctly (image paths, `flow.json` passthrough, etc.) before relying on it.
6. **Delete the `live` branch** once step 4 is confirmed working.

`GH_BOT_APP_ID`/`GH_BOT_APP_KEY` are a shared org-wide GitHub App credential also used by `project-automation.yml` in this repo and by `flowfuse/flowfuse`'s `docs.yml` — adding them as a Netlify env var is read-only reuse, it doesn't move or affect those existing usages.

## Test plan
- [x] `npm test` — all 111 tests pass
- [x] `npm run blueprints` locally against the real `../blueprint-library` sibling checkout — verified `image:` frontmatter paths, README→index renaming, and `flow.json` passthrough all match the previous script's output
- [ ] Manual setup steps above completed
- [ ] A real Netlify deploy from `main` (with the clone-fallback path, not the sibling path) verified to produce working blueprint pages


ZJvandeWeg:
@Yndira-E This PR can be reviewed, we just need @ppawlowski to help out to config Netlify before merging and to validate the flow

ppawlowski:
@ZJvandeWeg please elaborate on what I need to do on my end, as it is not clear what actions I should take.

nuxt/lib/blueprints-sync.mjs
```diff
@@ -0,0 +1,215 @@
...
+async function cloneBlueprints (ref, env, logger) {
+    const { mintInstallationToken } = await import('./github-app-token.mjs')
+    const token = await mintInstallationToken({
+        appId: env.GH_BOT_APP_ID,
+        privateKey: env.GH_BOT_APP_KEY,
+        owner: REPO_OWNER,
+        repo: REPO_NAME,
+    })
+    const authedUrl = REPO_URL.replace('https://', `https://x-access-token:${token}@`)
+    const redact = (text) => text.split(token).join('***')
```
Yndira-E: `mintInstallationToken()` runs before the retry loop in `cloneBlueprints`, so a transient GitHub API failure while minting the token isn't retried (unlike the git clone/checkout, which get 3 attempts with backoff), and its errors bypass the `redact()` call, currently harmless since octokit redacts its own auth header, but not something this code guarantees. Suggest moving the mint call inside the loop so minting gets the same retry + redaction as the clone.

Yndira-E:
Made two small local-dev fixes while testing this PR:

- Added `dev:blueprints` to `npm run dev` (it was previously only wired into `npm start`), using the existing `scripts/watch_blueprints.js`, so blueprints now auto-reload under the combined Nuxt+11ty dev command too.
- Fixed stale files: a blueprint renamed or removed in `blueprint-library` used to linger forever in `src/blueprints`. Sync now prunes only what's actually gone, without touching unaffected blueprints.

Verified locally against a real sibling checkout of `blueprint-library`, works as expected. Still need the Netlify env vars in place to test the rest.

dimitrieh:
@ZJvandeWeg can we move this forward?

ZJvandeWeg:
@ppawlowski Have you implemented the required changes from your side? We're waiting on your progress report.

ppawlowski:
> @ppawlowski Have you implemented the required changes from your side? We're waiting on your progress report.

@ZJvandeWeg  https://github.com/FlowFuse/website/pull/5694#issuecomment-5493214929

ppawlowski:
Environmental variables `GH_BOT_APP_ID` and `GH_BOT_APP_KEY` have been added to the Netlify configuration (thank you @Yndira-E for your help!).
A dedicated GitHub App, with read-only access to the `FlowFuse/blueprint-library` repository, has been created and installed in the FlowFuse GitHub organisation.

Important note: Since Netlify does not accept multiline environment variables, the application's key, which is originally in PEM format, has been added in Base64 format. To use it properly, it must first be decoded.

dimitrieh:
@ppawlowski thanks! Will you update this PR based on the changes? Some of the PRs for https://github.com/FlowFuse/website/issues/5777 rely on this PR being merged ideally.

ppawlowski:
> Will you update this PR based on the changes?

WDYM?

Yndira-E:
I'm on it

Yndira-E:
I can confirm blueprints are now available in the preview: https://deploy-preview-5694--flowforge-website.netlify.app/blueprints/ 

@ZJvandeWeg step 3: `Switch Netlify's production branch from live to main in site settings. Post-merge this is` seems incomplete, so I'll leave that one to you.

ZJvandeWeg:
@Yndira-E Updated Netlify to use `main` as production branch. Won't delete live until later today

Yndira-E:
Ok, could that be what's causing the test/deploy to try and run twice?

Screenshot 2026-09-14 at 18 56 28

Yndira-E:
Found the issue. The site on Netlify was renamed from flowforge to flowfuse and it needed to be updated on the [repo settings.](https://github.com/FlowFuse/website/settings/rules/12085742) to match it.

Screenshot 2026-09-14 at 19 14 02

I already fixed it.

PR #5697, opened 2026-08-27
Replace back-links with breadcrumbs, fix truncation and remove duplication

sumitshinde-84:
## Summary
- The breadcrumb and the "Back to X" link both navigated to the same parent page, dropped the redundant link and moved the breadcrumb into its old position (top of the article body) instead of the hero title block.
- That move also fixes a mobile overflow bug from #5668: a long post title in the breadcrumb, when placed in the hero's `flex` row, pushed the page wider than the viewport (confirmed live on production: 520px rendered in a 375px viewport).
- Fixed `Breadcrumbs.vue` so only the active (last) crumb truncates on overflow, instead of every crumb, including short ones like "Blog",  shrinking evenly.

## Checklist

### DIFF
--- a/nuxt/components/Breadcrumbs.vue
+++ b/nuxt/components/Breadcrumbs.vue
@@ -21,7 +21,7 @@ const displayItems = computed(() => props.items.map((item, index) => ({
     ...item,
     ui: {
         ...(item.to ? { link: 'hover:text-indigo-600' } : {}),
-        ...(index < props.items.length - 1 ? { item: 'shrink-0' } : {}),
+        ...(index === 0 && props.items.length > 1 ? { item: 'shrink-0' } : {}),
     },
 })))
 </script>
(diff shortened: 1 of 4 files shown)


n-lark (approved):
Hey! I pushed up a fix for the breadcrumbs. The `shrink-0` change introduced an overflow on docs pages, where we can be nested 4 layers deep, ex `docs/install/kubernetes/aws-terraform/`. The leading crumbs couldn't shrink, so they ran past the viewport and pushed the page title crumb off-screen. Made it so all but the first item can truncate.

Def could do another iteration where we collapse the middle crumbs — like `Docs / … / … / AWS EKS Installation` instead of shrinking every layer, but this will get the job done.

sumitshinde-84:
Thanks @n-lark  🙌 !

PR #5641, opened 2026-08-19
Add OPC UA landing page and interlink existing OPC UA content

sumitshinde-84:
## Description

Adds a new OPC UA pillar page and interlinks all 13 existing OPC UA blog posts back to it.

Since we already have a lot of content on OPC UA, and most of it is bringing in a good amount of traffic, it’s worth taking advantage of that existing content and building a pillar page around it.

The page is available at `/integrations/opcua`. The OPC UA item in the navigation bar previously linked directly to the docs, which can be too technical for users who are just taking a look or trying to understand what FlowFuse offers. Linking it to this pillar page gives users a better overview of what FlowFuse can do with OPC UA, along with links to all the relevant resources and content.

Planning to build similar pages for MQTT and Modbus in the upcoming weeks.

## Checklist

 - [ ] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [ ] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5703, opened 2026-08-28
Block new pages from being added to 11ty

Yndira-E:
## Summary

Following up on [FlowFuse/website#5641 (review comment)](https://github.com/FlowFuse/website/pull/5641#discussion_r3830628292): a new landing page (`src/landing/opcua.njk`) was added directly to the legacy 11ty stack instead of Nuxt, and there was no mechanism to catch that before merge. This adds one.

- `nuxt/lib/legacy-pages.test.mjs`: fetches `main` and fails if any `.njk` file under `src/` doesn't already exist there (at its current tip or within the last 30 commits, so an unrelated rename/delete elsewhere on `main` doesn't retroactively fail an untouched PR). No allowlist to maintain — any new `.njk` file, on any route, fails; editing an existing one is unaffected.
- Picked up by the existing `npm test` glob, so a failure here fails the already-required `test_website / Build and check` PR status check — this blocks merging with no change to branch protection / rulesets.
- **Scope is deliberately `.njk` only.** It does not catch a new page built as a bare `.md` file against an existing layout (11ty makes this possible without touching any `.njk`). Content directories that legitimately keep growing with new `.md` entries — `src/blog/`, `src/changelog/`, `src/customer-stories/`, `src/webinars/`, `src/blueprints/` — need to stay unaffected, and reliably telling "new content" apart from "a new page wearing an existing layout" isn't something this check does. `.claude/CLAUDE.md` and the handbook call this gap out explicitly; avoiding it relies on following the written rule, not automated enforcement.
- Documented in `.claude/CLAUDE.md` (engineers/AI agents) and the handbook (`/handbook/marketing/website#new-pages-must-be-built-in-nuxt`, marketing).
- If a page genuinely must ship on 11ty before its Nuxt equivalent exists, a member of the GitHub `admin` team can merge anyway via "Merge without waiting for requirements to be met".

## Test plan

- [x] `npm test` passes (112/112)
- [x] New `.njk` in an existing legacy directory (e.g. `src/landing/`) fails the check
- [x] New `.njk` in a brand-new, previously-unknown top-level `src/` directory fails the check
- [x] Editing an existing `.njk` page does not fail
- [x] New file under `src/_includes/` (a layout partial, not a page) does not fail
- [x] A file renamed away by an unrelated commit on `main` after the PR branched is not retroactively flagged (reproduced with a real rename in an isolated repo, confirmed the 30-commit lookback catches it)

## Checklist


ZJvandeWeg:
Fun idea @Yndira-E! Thanks!

PR #5671, opened 2026-08-24
blog: connect your industrial AI agent to FlowFuse

dimitrieh:
## Description

A how-to blog post for the 3.0 third-party agent feature, sitting between the `/ai` page and the documentation in depth. Copy now follows the reviewed doc, authored by @dimitrieh, and the slug moves to `industrial-ai-agent`.

It leads with the policy angle rather than the capability: teams under approved-AI-only rules were prompting an external tool for flow JSON and hand-importing it, so the AI was already doing the work through the clipboard. Then the mechanism, what you grant, and the conversation with IT. The connect widget now closes the post, after a new starter-prompts carousel, with a link down to it from where it used to interrupt the argument.

Adds `::prompt-carousel` (`nuxt/components/content/PromptCarousel.vue`) for the nine starter prompts, in `components/content/` so `/ai` and the docs page can reuse it rather than re-copying nine paragraphs. Two shared-component changes come with it: each connect step is a card while the picker is stacked, and the step number now sits on its heading's line in both layouts. Those also affect the changelog entry and the docs page.

No `image:` in the frontmatter yet, matching the handful of 2026 posts that also omit it, so the build is fine. **An Art Request is still needed** before this publishes.

One open item for review: Joyce's comment on the doc asked for the connector-directory news to get its own section to build on later. It is currently a titled note inside "The conversation with IT" rather than a section of its own.

## Related Issue(s)

https://github.com/FlowFuse/flowfuse/issues/8192

https://github.com/FlowFuse/website/issues/5612 (blog post deliverable, sub-issue of https://github.com/FlowFuse/engineering/issues/272)

## Depends on

Links to `/docs/user/expert/third-party-agents/`, which lands with https://github.com/FlowFuse/flowfuse/pull/8241. A link check against that path fails until it merges. Pairs with https://github.com/FlowFuse/website/pull/5670 for the `/ai` page and the changelog.

Rebased onto main to pick up https://github.com/FlowFuse/website/pull/5719, which is what keeps the picker's brand marks at 16px in prose. https://github.com/FlowFuse/website/pull/5714 was my duplicate of that fix and is closed; the commit adding it here is reverted in this branch.

[...]

PR #5714, opened 2026-08-31
css: stop prose stretching the agent picker brand marks

dimitrieh:
## Description

The agent picker's brand marks rendered at full column width on any surface that wraps it in a plain `.prose`. On the changelog entry the ChatGPT square came out 780px.

`.prose:not(.prose-natural-size-images) img` is `(0,2,1)` and forces `width:100%`. `.ff-agent-tab img` was only `(0,1,1)`, so it lost regardless of source order. The component's styles are deliberately unlayered to beat Tailwind's `@layer utilities`, but unlayered does not beat a *more specific* unlayered rule, which is the case that was missed.

Scoping the rule through the card and the tab strip takes it to `(0,3,1)`.

Affected:
- changelog entry: broken, fixed here
- blog post in https://github.com/FlowFuse/website/pull/5671: broken, fixed here
- docs page: was already fine, its container carries `prose-natural-size-images`
- `/ai`: was already fine, renders outside `.prose`

Verified all four surfaces at 375, 414, 768, 1024, 1280 and 1536px. Every mark measures 16x16 in all 24 combinations, and the tab strip still wraps correctly at 375px.

## Related Issue(s)

Found while reviewing https://github.com/FlowFuse/website/pull/5671, which adds a fourth surface for this component.

## Checklist

### DIFF
--- a/src/css/style.css
+++ b/src/css/style.css
@@ -3379,8 +3379,15 @@ lite-youtube::before {
     outline-offset: 2px;
 }
 /* The brand marks are <img>, so they keep their own colours; our own glyphs are
-   masked spans that inherit currentColor and turn white with the label. */
-.ff-agent-tab img {
+   masked spans that inherit currentColor and turn white with the label.
+
+   Three classes on purpose. Unlayered beats layered, but it does not beat a more
+   specific unlayered rule, and .prose:not(.prose-natural-size-images) img is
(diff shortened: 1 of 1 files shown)


Yndira-E:
@dimitrieh I used a more general approach to fix this in https://github.com/FlowFuse/website/pull/5719

PR #5731, opened 2026-09-02
Handbook: fix article errors in Subscription Agreement 1.5

MichaelBDavis:
Follow-up to [#5729](https://github.com/FlowFuse/website/pull/5729), which standardized quote naming and surfaced these errors.

## What

[`subscription-agreement-1.5.md`](https://github.com/FlowFuse/website/blob/handbook-subscription-agreement-article-fix/nuxt/content/handbook/sales/subscription-agreement-1.5.md) contains **12 instances of "an Quote" across 11 lines**. The agreement was drafted with a defined term taking "an" and later globally replaced with "Quote", without correcting the articles.

Section 2 (*Acceptance*) also carries a doubled replacement:

> (b) reference to an Quote Quote No. within a purchase order or similar document

## Changes

- `an Quote` → `a Quote` — 11 instances, on lines 17 (×2), 75, 77, 79, 121, 175, 177, 201, 255
- `AN QUOTE` → `A QUOTE` — 1 instance, in the all-caps warranty disclaimer in 3.4 (line 105)
- **Line 27, Acceptance definition** — removes the duplicated word so it reads `reference to a Quote No.`

## Please review line 27 specifically

Every other edit is a bare article correction and mechanically safe. Line 27 is the one place a **word is removed** rather than corrected. The reading it produces matches how the number is used in practice — a customer citing the quote number on their PO, per the [Company Quote Reference Number](https://github.com/FlowFuse/website/blob/main/nuxt/content/handbook/sales/engagements.md) rule in Engagements & Pricing — but it is worth a deliberate look rather than being waved through with the articles.

## Verification

Diffed word-by-word against `main`: line count unchanged, and every changed line differs only by `an`→`a` (or `AN`→`A`), except line 27's removed duplicate. No defined terms, obligations, amounts, or substantive wording touched.

## Note on scope

This file is live customer-facing legal text — the header warns it is *still referenced in contracts and quotes*, and it is the document customers sign per the closing line of Appendix 1. It was deliberately kept out of #5729 so a wording change to an executed-contract template is not buried in a process-documentation PR. Per [Legal](https://github.com/FlowFuse/website/blob/main/nuxt/content/handbook/sales/legal.md), agreement changes route through Deal Desk — flagging that in case this needs that sign-off before merge, even though the change is grammatical.


MichaelBDavis:
@ZJvandeWeg I suspect we can ignore this is 1.6 is coming soon.

ZJvandeWeg:
I'll merge it just because we should, but 1.6 is the leading agreement now indeed.

PR #5740, opened 2026-09-03
Fix internal links and breadcrumbs missing trailing slashes

sumitshinde-84:
Internal links across blog, changelog, handbook, docs, and application-guide pages,  including breadcrumbs,  were missing a trailing slash. This didn't 404 (Netlify silently 301s to the correct URL), so it went unnoticed while quietly costing every click and crawl an extra redirect hop.

Added a withTrailingSlash() helper and applied it wherever a link is built from a collection or nav-tree path, including the shared Breadcrumbs component.

### DIFF
--- a/nuxt/composables/useChangelogList.ts
+++ b/nuxt/composables/useChangelogList.ts
@@ -1,3 +1,5 @@
+import { withTrailingSlash } from '~/utils/withTrailingSlash'
+
 // How many entries are in the markup before any scrolling. The whole archive is
 // fetched in one query either way (it always was - the old paginated page fetched
 // everything and sliced it), so this only governs how much is rendered up front.
@@ -21,7 +23,10 @@ export interface ChangelogReleaseGroup {
 export function useChangelogList () {
     const { data: allEntries } = useAsyncData(
         'changelog-all',
-        () => queryCollection('changelog').order('date', 'DESC').all()
+        async () => {
+            const entries = await queryCollection('changelog').order('date', 'DESC').all()
+            return entries.map(entry => ({ ...entry, path: withTrailingSlash(entry.path) }))
+        }
     )
 
     const visibleCount = ref(CHANGELOG_INITIAL_VISIBLE)
--- a/nuxt/components/BlogListItem.vue
+++ b/nuxt/components/BlogListItem.vue
@@ -3,6 +3,8 @@ const props = defineProps<{
     entry: { path: string, title: string, description?: string, meta?: { description?: string }, date: string | Date, authors?: string[], image?: string }
 }>()
 
+const linkPath = computed(() => withTrailingSlash(props.entry.path))
+
 const authorNames = computed(() => useAuthorNames(props.entry.authors))
 const summary = computed(() => props.entry.description || props.entry.meta?.description || '')
 const formattedDate = computed(() => new Date(props.entry.date).toLocaleDateString('en-US', { year: 'numeric', month: 'short', day: 'numeric' }))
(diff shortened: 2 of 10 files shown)


sumitshinde-84:
@Yndira-E, could you please review this PR when you have a moment?

Yndira-E (changes requested):
One gap: the breadcrumbs in Handbook, Docs, and Application Guide still lack the trailing slash and redirect, even though those sections are called out as fixed. E.g. on this page the "Front End" breadcrumb links to /handbook/engineering/frontend (no slash), screenshot attached.

Image

Yndira-E:
@sumitshinde-84 Heads up — [#5744](https://github.com/FlowFuse/website/pull/5744) just got merged, fixing the sanitize-html/htmlparser2 crash (and the /robots.txt 502) via a different approach: inlining sanitize-html at build time via nitro.externals.inline, plus prerendering /robots.txt so it doesn't depend on the fallback function at all. It also turned out the crash was taking down more than just robots.txt, any 404 and /api/* route was 502ing too, all now confirmed fixed on main.

That makes the nuxt/package.json / package.json / package-lock.json changes here (moving sanitize-html out of the workspace + the htmlparser2 override) redundant, could you rebase on main and drop that part, keeping this PR scoped to the trailing-slash fix?

That also still leaves the breadcrumb gap I flagged earlier open, worth fixing in the same pass since it's the same class of bug.

nuxt/composables/useChangelogList.ts
```diff
@@ -1,3 +1,5 @@
+import { withTrailingSlash } from '~/utils/withTrailingSlash'
```
ZJvandeWeg, line 1: I find it hard to believe that Nuxt gets this wrong by default, and thus that this is the fix.

@sumitshinde-84 do you yourself understand this PR?

ZJvandeWeg (changes requested):

nuxt/composables/useChangelogList.ts
sumitshinde-84: @ZJvandeWeg `trailingSlash: true` (nuxt.config.ts:156) only drives the sitemap/canonical tags; it doesn’t rewrite links. `queryCollection().path` always returns paths without a trailing slash, so links built from it were missing one. They didn’t 404 because Netlify auto-redirects to the slash version, but this added an extra hop on every click/crawl, which is why it went unnoticed.

This PR adds a small `withTrailingSlash()` helper and applies it everywhere we build a link from a collection path (blog, changelog, docs, handbook, breadcrumbs), so links go directly to the real page instead of through a redirect. There were **771 URLs without a trailing slash**.

sumitshinde-84:
> @sumitshinde-84 Heads up — [#5744](https://github.com/FlowFuse/website/pull/5744) just got merged, fixing the sanitize-html/htmlparser2 crash (and the /robots.txt 502) via a different approach: inlining sanitize-html at build time via nitro.externals.inline, plus prerendering /robots.txt so it doesn't depend on the fallback function at all. It also turned out the crash was taking down more than just robots.txt, any 404 and /api/* route was 502ing too, all now confirmed fixed on main.
> 
> That makes the nuxt/package.json / package.json / package-lock.json changes here (moving sanitize-html out of the workspace + the htmlparser2 override) redundant, could you rebase on main and drop that part, keeping this PR scoped to the trailing-slash fix?
> 
> That also still leaves the breadcrumb gap I flagged earlier open, worth fixing in the same pass since it's the same class of bug.

Done, I have seprated the trailing slash fix, now all the urls has trailing slash at the end

Yndira-E (approved):

PR #5747, opened 2026-09-04
hmi: a page for HMI (conceptual)

dimitrieh:
## Description

**Conceptual for now.** Opened to discuss the idea and the framing, not to publish as it stands. Copy, positioning and the screenshot on the panel are all still open.

A page at `/hmi/` for the question `/platform/dashboard/` does not answer: "can this replace the panel on my machine?" That one is about where the screen runs, how it reaches the PLC, and how one design gets onto every station.

- The hero projects an OEE dashboard onto a photographed panel in CSS rather than as a flattened composite, so it stays sharp at any width and the screen contents can be swapped without re-compositing.
- Claim discipline follows `/ai/`: the capability and how-it-works sections are shipped behaviour only. Tags and Canvas are fenced into an "in design" block that says neither has shipped, worded to match `/dashboard/tags-and-canvas-feedback/`, and its CTA is the feedback form.
- Adds an HMI entry to the Products nav and the footer. Easy to drop if the page should stay unlinked while it is conceptual.
- Built in Nuxt, since `legacy-pages.test.mjs` fails the build for any new `.njk` under `src/`.

## Related Issue(s)

None.

## Checklist


ZJvandeWeg:
@dimitrieh @JoycePlaysFootball if we're a solution focussed company, the path should not include Platform in my mind. What do you think.

dimitrieh:
> the path should not include Platform in my mind. What do you think.

Agreed and adjusted. 

> if we're a solution focussed company

This needs more discussion, but not in this PR. Will get back on this.

PR #5745, opened 2026-09-03
Make the Application Guide part of the documentation

dimitrieh:
## Description

`/docs` is now assembled from two repos. FlowFuse/flowfuse keeps the product docs (how-to and reference, versioned with the code); the Application Guide keeps being authored here (explanation, which changes when our understanding changes, not when a release ships). `nuxt/lib/guides-sync.mjs` overlays `nuxt/content-guides/` onto the docs content tree after `docs-sync` populates it, so @nuxt/content sees one `docs` collection and the sidebar, breadcrumbs, TOC, search, sitemap and prerendering treat both sources identically. Nothing changes in FlowFuse/flowfuse.

The guide moves from `/application-guide/{guide}/{slug}/` to `/docs/application-guide/**` and `/docs/node-red-guide/**`, every old URL 301s, and "Edit this page" resolves to whichever repo owns the page.

One call for the reviewer: the new sidebar group sits **first**, above "FlowFuse User Manuals" (`navGroupOrder: 0`, chosen so no existing group needed renumbering in the other repo). Happy to drop it lower if the manuals should stay top.

## Related Issue(s)

None.

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [x] Suitable unit/system level tests have been added and they pass
 - [x] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5750, opened 2026-09-07
Make the Node-RED library part of the documentation, and convert it to Nuxt

dimitrieh:
## Description

Moves the Node-RED library from `/node-red/**` into `/docs/`, off Eleventy. Same mechanism as #5745.

**Stacked on #5745** (needs `guides-sync.mjs`).

| | |
|---|---|
| Moves | 81 pages + 433 assets → `nuxt/content-guides/`, Nunjucks → MDC |
| Also moves | `/node-red/flowfuse/**` → `/docs/flowfuse-nodes/` (our node reference, under FlowFuse User Manuals) |
| Stays put | `/node-red/` itself. It's a marketing page, not docs |
| Retires | `/node-red/learn/` ("100+ Tutorials (2026)") → becomes the group index |
| New sidebar group | "Node-RED", ranked **last**, after Contributing |
| Removes | 9 header nav slots + their footer mirrors |

### Worth a look

- **`library-markdown.mjs`** — reuses `docs-markdown`'s callout/HTML repair. Two orderings are pinned by tests because both fail silently: `{% renderFlow %}` must convert before leftover Nunjucks is stripped (77 diagrams), and Node-RED's `{{ msg.payload }}` must be code-spanned so MDC doesn't bind it.
- **`core-nodes-sync.mjs`** — the 39 core-node pages were never files. Still fetched from the Node-RED repo per build, but a missing help name now **fails the build** instead of rendering an empty section. 5xx is retried, 404 isn't.
- **Fixed 4 pages that have been shipping empty help.** `coreNodes.json` asked for `mqtt-in`/`mqtt-out` (upstream says `mqtt in`/`mqtt out`) and `upd` for both UDP nodes, a typo.
- **Redirects are explicit paths, not a splat** — a splat would swallow `/node-red/`.

### Not in scope

Whether any of these pages should be removed rather than moved is a separate decision.

## Related Issue(s)

None.

## Checklist


dimitrieh:
### Code review

Found 4 issues:

1. `protectMustaches` splits an existing code span in two and leaves the mustache exposed, which is the exact failure it exists to prevent. The backtick check only looks at the character immediately either side of the match, so a mustache in the *middle* of a longer span fails it and gets re-wrapped. Verified against real content: `src/_includes/core-nodes/template-use-case.md` goes from `{% raw %}\`<p>Hello {{payload.name}}!</p>\`{% endraw %}` to `\`<p>Hello \`{{payload.name}}\`!</p>\``, so `{{payload.name}}` reaches MDC as bare text. `http-request-use-case.md` cascades worse and also wraps two runs of ordinary prose as bogus inline code. Both feed the generated Template and HTTP request pages via `core-nodes-sync.mjs`.

https://github.com/FlowFuse/website/blob/04bc888854a48fc385655e12d43c5f893f80f844/nuxt/lib/library-markdown.mjs#L74-L79

2. The Function node's page and the Function *category* index resolve to the same URL. `slugFor('Function')` is `function`, which is also its category directory, so the generator writes both `core-nodes/function.md` (the node's mirrored help, ~13KB) and `core-nodes/function/index.md` (the category listing, ~650B), and both become `/docs/node-red/core-nodes/function/`. Whichever @nuxt/content resolves last wins. The legacy redirect for that path pointed at the node page, so a visitor following it may land on the category list instead. No build error, so CI is green on it.

https://github.com/FlowFuse/website/blob/04bc888854a48fc385655e12d43c5f893f80f844/nuxt/lib/core-nodes-sync.mjs#L250-L254

3. `{% caution %}` callouts are silently downgraded to plain paragraphs. `convertCallouts` in `docs-markdown.mjs` handles only `note`, `warning` and `critical`, then strips every remaining Nunjucks tag, but `.eleventy.js` defines `caution` as a first-class callout with its own icon. The OPC UA guide used it for "Server hosting is **not supported on FlowFuse Cloud**", which now renders with no visual distinction from body text.

https://github.com/FlowFuse/website/blob/04bc888854a48fc385655e12d43c5f893f80f844/nuxt/lib/library-markdown.mjs#L105-L108

4. Two redirects still point at moved paths, so they now double-hop. Their targets became redirect sources in the new map: `/node-red/core-nodes/mqtt/` → `/node-red/core-nodes/mqtt-in/` → `/docs/node-red/core-nodes/mqtt-in/`, and `/education/` → `/node-red/learn/` → `/docs/node-red/`.

https://github.com/FlowFuse/website/blob/04bc888854a48fc385655e12d43c5f893f80f844/nuxt/redirects.ts#L51-L53

One thing that is not a bug but is decision-relevant: this extends the sync-script pattern that was objected to on #5495 and has not been revisited since ("Why a hacky script? I would've expected this to use the Nuxt collection primitive?", https://github.com/FlowFuse/website/pull/5495#issuecomment-5358122597). Worth settling before adding two more generators to it.

dimitrieh:
All four resolved in 02d28260b.

1. `protectMustaches` now splits on fences and code spans and only rewrites what falls between them, instead of checking adjacent backticks. Two committed pages carried the damage and are regenerated.
2. Per-category pages are gone, so nothing shadows the Function node's URL; their grouping is now headings on the section index. The five bare category paths come out of the redirect map too, since those were never pages under Eleventy either.
3. `{% caution %}` is converted locally, in the same markup shape, before the tag-stripping pass.
4. `/node-red/core-nodes/mqtt/` and `/education/` go straight to `/docs/`.

Each of the three code defects has a regression test naming the failure, because all three were silent: none produced a build error and CI was green on every one of them.

dimitrieh:
Verified every redirect against the deploy preview, not just the source: **121 of 121 old URLs 301 to the correct new URL**, no failures, and every one lands on a 200.

| | |
|---|---|
| `/node-red/learn/` | `/docs/node-red/` |
| `/node-red/protocol/modbus/` | `/docs/node-red/protocol/modbus/` |
| `/node-red/core-nodes/function/` | `/docs/node-red/core-nodes/function/` |
| `/node-red/flowfuse/mcp/` | `/docs/flowfuse-nodes/mcp/` |
| `/node-red/` | 200, unmoved: it is the marketing page |

Two notes:

- The five bare category paths (`/node-red/core-nodes/common/` and siblings) 404 rather than redirect. That is intended: they were never pages, since the Eleventy templates were pagination-only.
- These are Nitro route rules, so they are served by the app rather than at the edge. The existing TODO in `nuxt/redirects.ts` about moving redirects to `netlify.toml` still stands and predates this PR.

dimitrieh:
Follow-up opened: #5752, which answers the sync-script objection from #5495 (https://github.com/FlowFuse/website/pull/5495#issuecomment-5358122597) for the part where it actually applies.

It gives the `docs` collection a second source pointing straight at `nuxt/content-guides/`, so the guides are read rather than copied in. `guides-sync.mjs` shrinks to copying non-markdown assets plus a collision check, and the `editUrl`/`updated` stamping moves into `content:file:beforeParse`. Small change: 7 code files, plus 27 `README.md` to `index.md` renames that `@nuxt/content` requires.

The other two generators stay, for reasons rather than convenience:

- `docs-sync` keeps its own clone. `@nuxt/content`'s native git source is shallow, which would flatten every page's `updated` date, which is already the documented reason that clone is not shallow.
- `core-nodes-sync` cannot be a collection source at all: its input is an HTTP fetch, not files on disk.

Draft, and stacked on this PR. Merge order is #5745, then this, then #5752.

dimitrieh:
FYI: this PR stackes on https://github.com/FlowFuse/website/pull/5745 which when merged, will reduce file count in this PR.

sumitshinde-84:
Hey @dimitrieh, this looks good. There are two things I noticed, though:

* The **Getting Started** nav item is duplicated several times in the navbar.

gettting-started-duplicated

* The page this PR removes was getting traffic. I'm not against removing it, but I think it's worth making sure the new **"Using Node-RED"** page is also optimized for the old keyword so we don't lose that traffic. I'll add a suggestion for this in the PR.

Also, one question: apart from `node-red/learn`, is there any other page being removed by this PR? And have any titles or meta titles been changed?

Also, I find this PR really big. Would it be possible to split it into smaller chunks? If not, that's fine.

nuxt/content-guides/node-red/README.md
```diff
@@ -0,0 +1,80 @@
...
+## Reference
+
+[Terminology](/docs/node-red/terminology/) for the words that come up in Node-RED
+discussions, and [keyboard shortcuts](/docs/node-red/keyboard/) for the editor.
+
+::callout{icon="i-lucide-arrow-right"}
+**Running Node-RED for an organisation?** Access control, version history, deployment
+across environments and remote instance management are what FlowFuse adds on top.
+See [Using FlowFuse](/docs/user/).
+::
```
sumitshinde-84, line 81:
```suggestion
---
title: "Using Node-RED"
navTitle: "Using Node-RED"
metaTitle: "Node-RED Documentation | Using Node-RED"
navOrder: 1
navGroup: "Node-RED"
navGroupOrder: 7
meta:
    description: "Reference documentation for using Node-RED with databases, communication protocols, hardware, peripheral devices, notification services and integration technologies."
    keywords: node-red, node red, node-red documentation, node-red core nodes, node-red database, node-red protocols, node-red hardware, industrial iot
---

# Using Node-RED

Node-RED is the open-source runtime FlowFuse runs, governs and scales. This section provides
reference documentation for working with Node-RED itself: connecting it to databases,
communication protocols, hardware and notification services, and understanding what each
core node does.

It is here to answer a question you already have. It is not a path into FlowFuse, and it
does not need to be read in order. If you are setting up FlowFuse, start with
[Using FlowFuse](/docs/user/) instead.

::callout{icon="i-lucide-book-open"}
**The Node-RED project's own documentation is the primary source.** Installing Node-RED,
the editor, the message model, writing Functions and the API reference all live at
[nodered.org/docs](https://nodered.org/docs/). This section covers the ground that
documentation does not: per-node reference on the web, and connecting Node-RED to
specific databases, protocols, hardware and services.
::

## Getting started

Installing Node-RED, finding your way around the editor, and the basics of shaping a
message. [Getting started](/docs/node-red/getting-started/)

## Core nodes

Reference documentation for each node in the default Node-RED palette, with a worked
reason to reach for it. [Core nodes](/docs/node-red/core-nodes/)

## Communication protocols

Modbus, OPC UA, MQTT, AMQP, WebSocket and LwM2M, for connecting Node-RED to controllers
and edge equipment. [Communication protocols](/docs/node-red/protocol/)

## Databases

Reading and writing SQL, NoSQL and time-series databases with Node-RED, with one guide
per database. [Databases](/docs/node-red/database/)

## Integration technologies

Webhooks, REST APIs and GraphQL, for connecting Node-RED to the rest of your systems.
[Integration technologies](/docs/node-red/integration-technologies/)

## Notification services

Sending alerts from Node-RED by email, Telegram and Discord.
[Notification services](/docs/node-red/notification/)

## Hardware

Running Node-RED on a Raspberry Pi, a Siemens IoT2050, and other industrial gateways.
[Hardware](/docs/node-red/hardware/)

## Peripheral devices

Webcams, barcode scanners and other devices connected to the machine Node-RED runs on.
[Peripheral devices](/docs/node-red/peripheral/)

## Reference

[Terminology](/docs/node-red/terminology/) for the words used in Node-RED
documentation, and [keyboard shortcuts](/docs/node-red/keyboard/) for the editor.

::callout{icon="i-lucide-arrow-right"}
**Running Node-RED for an organisation?** Access control, version history, deployment
across environments and remote instance management are what FlowFuse adds on top.
See [Using FlowFuse](/docs/user/).
::
```

sumitshinde-84: optimized for old keyword "node-red documentation"

nuxt/content-guides/flowfuse-nodes/ai/image-classification.md
```diff
@@ -1,13 +1,12 @@
...
-  parent: AI
+title: Image Classification
+navTitle: Image Classification
+navOrder: 50
 meta:
-  title: Image Classification
   description: Classify images using ONNX models directly in Node-RED. Supports pre-trained and custom models for tasks like labeling, content moderation, and object recognition.
 ---
 
-# {{ meta.title }}
```
sumitshinde-84, line 10: I think using the template expression way was better, so when needed, we could just change the title variable, and it would be updated automatically.

sumitshinde-84:
@dimitrieh, regarding the Node-RED Core Nodes section, it was previously categorized into sections such as Function, Network, Common, Parser, etc. Now, those categories have been removed.

nuxt/content-guides/node-red/README.md
dimitrieh: Taken, with one change, because the suggestion as written would not have moved the browser title.

`meta.title` never reaches a docs page's `<title>`. The page titles itself `navTitle || title` (`nuxt/pages/docs/[...slug].vue`), and the `docs` collection schema does not declare `meta.title`, so @nuxt/content strips the key before the page ever sees it. The tab would still have read "Using Node-RED".

So the keyword goes in `metaTitle` instead, which this commit teaches the docs page to prefer and declares in the schema. The page now titles as **Node-RED Documentation • FlowFuse Docs**, which is the phrase itself rather than a phrase plus a qualifier. Your `meta.description` and `meta.keywords` are in verbatim, and the body wording too.

One thing to know separately: no docs page emits `meta.keywords` into the markup, on this branch or on main. Nothing is lost by keeping it in frontmatter, but it is not doing SEO work there today.

sumitshinde-84:
@dimitrieh, the MQTT In and MQTT Out nodes have duplicated H1 headings. I’ve attached a screenshot from the MQTT Out node as an example.
mqtt-core-node-title-duplication

nuxt/content-guides/flowfuse-nodes/ai/image-classification.md
dimitrieh: Kept as a literal, because `{{ }}` cannot come along. In MDC it is a Vue expression, and this section is full of literal Node-RED mustaches (`{{ msg.payload }}`) that the migration has to code-span specifically so MDC does *not* bind them. Nothing else in `nuxt/content/` or `nuxt/content-guides/` interpolates frontmatter either, so it would be a new convention for one page.

The H1s are reproduced exactly as Eleventy rendered them, including the three pages where the H1 was deliberately shorter than the SEO title (`notification/`, `peripheral/`, `getting-started/library/`). Those would have been flattened by a `{{ meta.title }}` heading.

dimitrieh:
All four fixed, in 35d1a80f6 and 88b3b982c.

- **Duplicated "Getting Started"**: every nested directory index took its *parent's* `eleventyNavigation` key instead of its own. Editor, Node-RED Library and Programming are back.
- **Core node categories**: back as a directory each, so the sidebar nests the way the editor's palette does. That also resolves the URL collision which removed them last round, the other way round: the category owns `core-nodes/function/` and the node is `core-nodes/function/function/`. The old flat URLs still reach a node in one hop.
- **Titles: yes, and on every moved page, not only `/node-red/learn/`.** Eleventy titled a page `metaTitle || navTitle || meta.title`; the migration dropped `metaTitle`, and the docs page titles `navTitle || title`, so each title collapsed to its sidebar label. "Using MySQL with Node-RED (2026 Updated)" became "MySQL", "Node-RED - Inject Node" became "Inject". Every page now carries its Eleventy title in `metaTitle`, which the docs page prefers. Product docs titles are untouched, nothing from FlowFuse/flowfuse sets the field.
- **Other pages removed**: only `/node-red/learn/`. `/node-red/` itself stays put, it is the marketing page.

The nav and title defects each have a check over the guides tree now, because neither broke a build, a link or a test.

dimitrieh:
Fixed in 11cde53c7. Both includes open with `# {{ meta.title }}`, and the page is assembled with its own `# <node name>`, so resolving that interpolation put two identical headings next to each other.

Eleventy had the same duplicate, but empty (`meta.title` was undefined on a paginated core-node page, so it rendered as a bare `<h1 tabindex="-1">`), which is why it went unnoticed. A heading that just repeats the node name is now dropped, and any other H1 in an include becomes an H2. Three other includes carried a second H1 and fold the same way. Checked over every committed include, not a fixture.

nuxt/content-guides/node-red/README.md
sumitshinde-84: yes, it is fine to drop the meta.keywords since they are depreciated now

sumitshinde-84: I have added metaTitle property in suggestion

dimitrieh: Taken as written, in 480216b6d. 39 characters, so it renders as "Node-RED Documentation | Using Node-RED • FlowFuse Docs" and still fits a search result at 55.

Keywords stay, since your suggestion keeps them: harmless as source data even though nothing emits them.

And yes, applied. The plumbing was the part that was missing: `metaTitle` is now declared in the `docs` collection schema and preferred by `nuxt/pages/docs/[...slug].vue`, so the field actually reaches `<title>`. That went in for every moved page, not just this one, because they had all silently lost their titles to the sidebar label.

sumitshinde-84:
@dimitrieh cta looks weird

cta-node-red

dimitrieh:
Good catch, fixed in e583b3d9e. That CTA is a hand-written wrapper `<div>`, and its contents only parse as markdown because of the blank lines around them. One of the migration transforms joins adjacent raw-HTML blocks by removing blank lines, which turned the whole thing into a single HTML block and made the markdown inside render as literal text. Blank lines restored, and I scanned every migrated page for markdown wedged against an HTML tag: this was the only one.

While in there, a review pass turned up five more of the same shape, all of which built green:

- The `link` and `websocket` pages were showing a fraction of their help. Those catalogue entries name a whole node family, and Eleventy joined every matching help block where the new code took only the first, while still captioning it as the node's complete built-in help.
- `/docs/node-red/core-nodes/read-file/` has been showing the **Watch** node's help, in production too. The catalogue pointed it at the wrong file.
- The Template node's page had backticks injected into the code sample that teaches Mustache syntax, because the fence was indented under a list item and the transform only recognised fences at column 0.
- The Function category page and the Function node inside it both read "Function" in the sidebar, one directly above the other.
- Two guides linked the Function node at what is now the category's URL.

dimitrieh:
Heads-up before you look again: **the deploy preview on this PR is stale and will stay stale.**

This PR is now stacked (it targets `docs/application-guide-into-docs` rather than `main`, so its diff is its own 8 commits instead of the whole move). Netlify only builds previews for PRs targeting `main`, so it has stopped building this one. The link at the top still resolves, but it is serving 02d28260b, from before any of today's fixes: the duplicated "Getting Started", the raw-markdown CTA and the flat core-nodes list are all still in it. Anything you screenshot from there now is already fixed.

Merge order is #5745, then this, then #5752. The preview comes back on its own once #5745 merges, because GitHub retargets this PR to `main` at that point.

Until then, the honest status of the fixes is: green CI, and asserted by unit tests over the real content tree, but not visually confirmed on a rendered page. If you would rather have the preview back than the smaller diff, say so and I will point this PR at `main` again.

dimitrieh:
Since this PR's deploy preview is not being built while it is stacked, here are the three fixes rendered from the branch itself, in a dev container running the same Nuxt app.

**The duplicated "Getting Started" is gone.** Editor, Node-RED Library and Programming now carry their own nav labels, which is what their Eleventy nav entries had before the migration flattened them into their parent's.

Docs sidebar with the Getting Started subtree correctly labelled

**The MQTT Out page has one H1.** The use-case include opened with `# {{ meta.title }}`, which resolved to the node's own name next to the heading the generator adds. Note the breadcrumb too: the palette categories are back, so this page sits under Core Nodes / Network Nodes.

MQTT Out docs page with a single H1 heading

**The CTA renders as markup again.** Heading, paragraphs and links, rather than the literal `###` and `[text](url)` you screenshotted.

Firebase page CTA rendering as a heading, paragraphs and links

Two things these also confirm, which I could not check on the stale preview: the browser titles are back (`Node-RED - MQTT Out Node` and `Using Firebase with Node-RED (2026 Updated)` respectively, rather than `MQTT Out` and `Firebase`), and the nested core-node URLs resolve.

sumitshinde-84 (approved):

dimitrieh:
The core node pages were written by a build script (`core-nodes-sync.mjs`) out of three things that were not content: a JSON catalogue, 41 markdown fragments in an Eleventy includes folder, and a fetch of Node-RED's help. They are now ordinary content files, one per node, each carrying its own frontmatter and prose and ending with `::node-red-help{category="network" file="10-mqtt" node="mqtt out"}`. Same URLs, same titles, same help: the fetch still runs during the build, so the help is in the static HTML, and it is sanitised before rendering now.

Most of the defects found on this PR were in that script rather than in the move, so `core-nodes-sync.mjs` and `library-markdown.mjs` are both deleted and the change is net smaller than what it replaces. A new check reports help that Node-RED offers but nothing asks for, and it found three: `http-response`, `tcp-out` and `mqtt-broker` have prose written and real help upstream but have never been published, one because its file was named `tpc-out`. Those three fragments stay in place, since publishing them is a content decision rather than part of this move.

KristopherLeads (approved):

PR #5751, opened 2026-09-07
test: pick the newest handbook commit by author date, not traversal order

dimitrieh:
## Description

Fixes a latent flaky assertion in `handbookChanges.test.mjs`. **Test-only; the module is unchanged and was never at fault.**

The test asked git for `log -1` and asserted `weeks[0]` was that commit's week. But:

- git orders traversal by **commit** date
- `getHandbookChanges` buckets weeks by **author** date, and says so in a comment

Those agree only while author dates happen to be monotonic with merge order. Any rebase breaks it, so does a PR that sat open while a later-authored one merged ahead of it.

Caught while rebasing an unrelated branch: tip authored 09-03, newest handbook change authored 09-07, two different weeks.

The test now takes the maximum author date, which is what `weeks[0]` is built from. Verified on `main` and on the history that exposed it.

### Side note

This could not have failed CI. git isn't on the PATH in `test_website`, so `getHandbookChanges` returns `[]` and every integration assertion in this file silently doesn't run. Left alone here, since fixing it is a workflow change.

## Related Issue(s)

None.

## Checklist

### DIFF
--- a/nuxt/server/lib/handbookChanges.test.mjs
+++ b/nuxt/server/lib/handbookChanges.test.mjs
@@ -147,18 +147,39 @@ test('getHandbookChanges produces well-formed, newest-first weekly data', () => 
 })
 
 test('the actual latest handbook commit appears in the most recent week', () => {
-    // Ask git directly (independent of handbookChanges.mjs) for the single most
-    // recent commit that touched a tracked handbook file, and confirm it shows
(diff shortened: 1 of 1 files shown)


dimitrieh:
came across this while developing https://github.com/FlowFuse/website/pull/5750

Yndira-E (approved):

PR #5752, opened 2026-09-07
docs: source guides from a content-collection instead of copying them in

dimitrieh:
## Description

A reviewer asked on https://github.com/FlowFuse/website/pull/5495#issuecomment-5358122597 why the docs pipeline uses a custom sync script instead of `@nuxt/content`'s own collection primitives. This checks that against the installed version (3.15.2) and applies it where it genuinely helps.

The guides overlay (`nuxt/content-guides/`) is now a second source on the `docs` collection, read directly from disk instead of being copied into `nuxt/content/docs` first. The frontmatter it used to inject during that copy now happens in the existing `content:file:beforeParse` hook.

| Source | Change |
|---|---|
| Guides (`nuxt/content-guides/`) | Native second collection source, no copy step |
| FlowFuse/flowfuse docs (`docs-sync.mjs`) | Unchanged |
| Core nodes (`core-nodes-sync.mjs`) | Unchanged |

The other two sources stay as they are:

- `docs-sync.mjs` needs the full git history of each file it clones (for the "updated" date). `@nuxt/content`'s own git source only does a shallow clone, which would date every page the same.
- `core-nodes-sync.mjs` generates pages from an HTTP fetch, not files on disk, so there is nothing to point a collection source at.

Also renamed the guides' `README.md` section-index files to `index.md`, since `@nuxt/content` only recognizes the latter.

**What this does not solve:** two of the three sync scripts are unchanged. `scripts/sync_docs.mjs` (used outside a Nuxt build) no longer copies guide pages into `nuxt/content/docs`, only their non-markdown assets.

Stacks on #5745 and #5750.

## Related Issue(s)

https://github.com/FlowFuse/website/pull/5495#issuecomment-5358122597

## Checklist


dimitrieh:
has a lot of changes as this stacks on top of other PRs. If those get merged, this one can follow and it will have less changes. Until then this stays a draft PR

dimitrieh:
@ZJvandeWeg can you have a look here? This essentially fixes your request at https://github.com/FlowFuse/website/pull/5495#issuecomment-5358122597

Issue #5777, opened 2026-09-11
Complete conversion to Nuxt


dimitrieh:
The site is served by two generators: Eleventy renders most pages from `src/`, Nuxt renders the rest, and `nuxt/server/middleware/legacy.ts` proxies between them in dev. This is the last stretch of that migration: every remaining page moves to Nuxt, then Eleventy comes out.

## Order

1-17 are independent and target `main`, but they all touch the same two registry files (`legacy.ts`, `nuxt.config.ts`), so expect a small rebase after each merge. Merge top to bottom.

1. https://github.com/FlowFuse/website/pull/5778
2. https://github.com/FlowFuse/website/pull/5779
3. https://github.com/FlowFuse/website/pull/5780
4. https://github.com/FlowFuse/website/pull/5781
5. https://github.com/FlowFuse/website/pull/5782
6. https://github.com/FlowFuse/website/pull/5783
7. https://github.com/FlowFuse/website/pull/5784
8. https://github.com/FlowFuse/website/pull/5785
9. https://github.com/FlowFuse/website/pull/5786
10. https://github.com/FlowFuse/website/pull/5787
11. https://github.com/FlowFuse/website/pull/5788
12. https://github.com/FlowFuse/website/pull/5789
13. https://github.com/FlowFuse/website/pull/5790
14. https://github.com/FlowFuse/website/pull/5791
15. https://github.com/FlowFuse/website/pull/5792
16. https://github.com/FlowFuse/website/pull/5793
17. https://github.com/FlowFuse/website/pull/5794
18. https://github.com/FlowFuse/website/pull/5795
19. https://github.com/FlowFuse/website/pull/5796

18 and 19 are drafts because 18 deletes `src/` and the dev proxy, which against today's `main` would delete pages 2-17 are still migrating. They are based on `nuxt/eleventy-teardown-base`, a scaffolding branch holding the combined state of 1-17, so their diffs show only their own change. Retarget both to `main` once 1-17 have landed, then delete that branch.

## Notes

- https://github.com/FlowFuse/website/pull/5694 overlaps 17: it adds its own `nuxt/lib/blueprints-sync.mjs` and rewrites the same workflow. Land it first and rebase 17 onto it.
- Every route was checked against the published page before this was raised.

dimitrieh:
@Yndira-E @KristopherLeads FYI. To complete the efforts for the Nuxt migration so there is no longer needing to deal with two architectures at the same time.

dimitrieh:
[...]

PR #5780, opened 2026-09-11
use-cases: serve /use-cases/ from Nuxt

dimitrieh:
## Description

Moves `/use-cases/` to Nuxt. The six architecture pages become `.vue` pages; the two operational use-cases stay data-driven and render through `pages/use-cases/[slug].vue`.

FAQ answers now go through an escaping renderer rather than being printed raw.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5780--flowforge-website.netlify.app/use-cases/
- https://deploy-preview-5780--flowforge-website.netlify.app/use-cases/data-integration/
- https://deploy-preview-5780--flowforge-website.netlify.app/use-cases/edge-connectivity/
- https://deploy-preview-5780--flowforge-website.netlify.app/use-cases/it-ot-middleware/
- https://deploy-preview-5780--flowforge-website.netlify.app/use-cases/mes/
- https://deploy-preview-5780--flowforge-website.netlify.app/use-cases/scada/
- https://deploy-preview-5780--flowforge-website.netlify.app/use-cases/uns/

dimitrieh:
Rebased on `main` (which now carries the six merged PRs from this set).

A review pass over the whole migration found three things here:

- `/use-cases/production-monitoring/` and `/use-cases/shop-floor-communication/` render through `[slug].vue` over a `data` collection, so @nuxtjs/sitemap's static-route discovery cannot see them and `content-urls.get.ts` cannot either (it keys on `path`). They had also left `sitemap-legacy.xml` when their `.njk` files were deleted, so they were in neither sitemap. Added to `sitemap.urls`, reusing the same collector the prerender list uses.
- `/use-cases/it-ot-middleware/` lost `px-3 md:px-12` and `md:px-6` on two containers. The `.njk` carried both those and `md:px-0`; Tailwind emits padding utilities in ascending scale order, so the larger value was the winning rule and desktop side padding had gone to zero. Restored verbatim.
- The `<LiteYoutube>` on `/use-cases/edge-connectivity/` passed a title already ending in "- YouTube video", which the component appends again, so the accessible name read "... - YouTube video - YouTube video".

Also tidied the stale list in `legacy.ts`, which still named `/webinars` and `/free-consultation` as 11ty routes after they migrated, and the `llms` note that named sections this set moves.

dimitrieh:

nuxt/composables/useScrollToAnchor.ts
Yndira-E, line 1: https://github.com/FlowFuse/website/pull/5786#discussion_r4017410455

nuxt/server/middleware/legacy.ts
```diff
@@ -21,7 +21,7 @@ const NUXT_PREFIXES = ['/handbook', '/ebooks', '/whitepaper', '/pricing', '/docs
 // already falls through to the 11ty proxy by default). Remove entries here as they migrate:
 // /about, /blueprints, /careers, /community, /events
 // /free-consultation, /industries, /landing, /node-red, /partners, /platform,
-// /use-cases, /webinars
```
Yndira-E, line 24: @dimitrieh this needs to be added to the `NUXT_PREFIXES` or it won't render the pages on dev mode

src/use-cases/use-cases.json
```diff
@@ -1,6 +0,0 @@
-{
-    "layout": "use-case",
```
Yndira-E, line 2: @dimitrieh remember to remove the unused layouts

src/use-cases/it-ot-middleware.njk
```diff
@@ -1,289 +0,0 @@
----
-layout: solution
```
Yndira-E, line 2: @dimitrieh I'd say this layout is dead code too

nuxt/pages/use-cases/it-ot-middleware.vue
```diff
@@ -0,0 +1,241 @@
...
+]
+
+const CASE_STUDIES = [
+    { title: 'Revolutionizing Precision Manufacturing with Node-RED', image: '/images/stories/abrasive_tech.jpg', alt: 'Photo of helicopters and a precision manufacturing machine', link: '/customer-stories/leveraging-node-red-and-flowfuse-to-automate-precision-manufacturing/' },
+    { titleLead: 'The Future of Textile Manufacturing Powered with ', titleTail: 'Node-RED', image: '/images/stories/stfi-future-textile.jpg', alt: 'Photo of the STFI floor plant', link: '/customer-stories/stfi-future-of-textile-powered-by-node-red/' },
+    { title: 'Node-RED Enables Digital Transformation of a Large US Manufacturing Company', image: '/images/stories/large-us-manufacturing.jpg', alt: 'Photo of a manufacturing company with a worker holding a laptop', link: '/customer-stories/manufacturing-digital-transformation/' },
+]
+
+const WHITEPAPERS = [
+    { title: 'Accelerating Innovation in Manufacturing with FlowFuse', image: '/whitepaper/images/whitepaper-manufacturing.png', alt: 'Photo of a manufacturing company with a worker holding a laptop', link: '/whitepaper/accelerating-innovation-in-manufacturing-with-flowfuse/' },
```
Yndira-E, line 54: @dimitrieh The whitepaper images are not rendering on local

dimitrieh:
Pushed:

- `use-case.njk` and `solution.njk` deleted with their aliases in `.eleventy.js`. Nothing referenced either once `src/use-cases/` went, checked across all the open migration branches, not just this one.
- The scroll composable is gone and the anchor is plain again. The site-wide `scroll-behavior: smooth` in `style.css` already animates it, and that file survives the teardown as `nuxt/assets/css/style.css`.

Two things from this review moved to https://github.com/FlowFuse/website/pull/5815 instead, because they are edits to `legacy.ts` and doing them per branch would just conflict eight ways: the `/use-cases` registration, and the whitepaper covers not rendering locally. That second one is the same cause, `/whitepaper` is a Nuxt prefix while the images are still in `src/whitepaper/images/`, so dev needs the carve-out `/vs` and `/resources` already have. Merge 5815 after this.

dimitrieh:
Correction to my earlier comment: #5815 is closed, the registration belongs in each migration PR instead. Pushed here. It also carries the `/whitepaper/images/` carve-out, which is what makes the whitepaper covers render locally.

nuxt/composables/useScrollToAnchor.ts
dimitrieh: Removed. The anchor keeps its `href="#whitepapers"` and the site-wide `scroll-behavior: smooth` in `style.css` animates the jump, so the composable was only restating the default. That file survives the teardown as `nuxt/assets/css/style.css`, so the behaviour does not go away with `src/`.

nuxt/server/middleware/legacy.ts
dimitrieh: Added `/use-cases` to `NUXT_PREFIXES`. It also needed a `/whitepaper/images/` carve-out, which is the cause of the broken covers you flagged separately.

src/use-cases/use-cases.json
dimitrieh: Deleted `use-case.njk` along with its alias in `.eleventy.js`. Checked across every open migration branch rather than just this one, since each was cut from main independently: the only consumers were in `src/use-cases/`, which this PR removes.

src/use-cases/it-ot-middleware.njk
dimitrieh: Agreed, `solution.njk` is dead too. It and its `.eleventy.js` alias are gone, same check as the other layout.

nuxt/pages/use-cases/it-ot-middleware.vue
dimitrieh: Fixed, and it was a dev-only fault rather than a bad path. The covers are at `/whitepaper/images/...`, `/whitepaper` is in `NUXT_PREFIXES`, but the files are still in `src/whitepaper/images/` and only reach `nuxt/public` through the 11ty passthrough in a production build. So dev answered from Nuxt and 404d while the deploy preview was fine. Added the carve-out `/resources/images/` and `/vs/images/` already have.

nuxt/composables/useCaseIcon.ts
Yndira-E, line 1: @dimitrieh why not assign the real Heroicons name directly in each .yml instead of transforming the old 11ty icon names through a new useCaseIcon.ts? Feels like we're carrying over an 11ty pattern we're trying to get away from.

dimitrieh: Agreed, done. The `.yml` now names the glyph in full (`i-heroicons-chart-bar-square`) and `<UIcon>` resolves it directly; `useCaseIcon.ts` is gone.

It was also the convention already on `main`: every `icon:` key under `nuxt/content/` is a full name. And this PR was inconsistent with itself, since `pages/use-cases/uns.vue` already passed `feature.icon` straight to `<UIcon>`.

Worth noting what the mapper was buying, since dropping it is a real trade: it fell back to a neutral dot for an unknown key. Now an unresolvable name renders nothing. That is the better failure, because it is visible in review instead of looking deliberate. `nuxt/utils/navIcons.ts` is the same pattern one layer over, and it fails silently in exactly that way: `/vs/kepware/` and `/vs/litmus/` name `arrows-pointing-out` and `link`, neither is in the registry, and `NavIcon` renders nothing, so both pages are missing an icon on live right now. Both keys are added in #5781.

Yndira-E:
@dimitrieh let's hide the `Common in these industries` section for now, new pages are coming and we can bring it back when those are live

dimitrieh:
Hidden in 507eefb. The component and the `industries:` field in each use-case stay where they are, so bringing it back is putting one line back in `pages/use-cases/[slug].vue`.

Same commit drops `src/_includes/components/use-case-chips.njk`, the 11ty copy of that section: its only caller was `layouts/use-case.njk`, which this PR deletes.

Yndira-E (approved):

PR #5781, opened 2026-09-11
industries: serve /industries/ from Nuxt

dimitrieh:
## Description

Moves `/industries/` to Nuxt. The seven industry pages were pure frontmatter read by a Nunjucks layout, so they become a content collection plus one `[slug].vue`; `/industries/automotive/` keeps its own page because it carries more than the others.

Shares `components/use-case/Card.vue` with the `/use-cases/` PR, whichever lands first.

## Related Issue(s)

Part of https://github.com/FlowFuse/website/issues/5777

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5784, opened 2026-09-11
partners: serve /partners/ from Nuxt

dimitrieh:
## Description

Moves `/partners/` to Nuxt: the listing, certify-hardware, ctrlx and the referral sign-up.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5784--flowforge-website.netlify.app/partners/
- https://deploy-preview-5784--flowforge-website.netlify.app/partners/certify-hardware/
- https://deploy-preview-5784--flowforge-website.netlify.app/partners/ctrlx/
- https://deploy-preview-5784--flowforge-website.netlify.app/partners/referral-sign-up/

nuxt/server/middleware/legacy.ts
```diff
@@ -13,14 +13,16 @@ const NUXT_ROUTE_PREFIXES = ['/integrations/', '/raw/']
...
+
 
 // Top-level routes still on 11ty, not yet ported to Nuxt (everything not listed above
 // already falls through to the 11ty proxy by default). Remove entries here as they migrate:
 // / (homepage), /about, /blueprints, /careers, /community, /events,
-// /free-consultation, /industries, /landing, /node-red, /partners, /platform,
+// /free-consultation, /industries, /landing, /node-red, /platform,
 // /use-cases, /vs, /webinars
 
+
```
ZJvandeWeg: Why?

nuxt/nuxt.config.ts
```diff
@@ -383,6 +383,11 @@ export default defineNuxtConfig({
                     '/pricing',
                     '/resources/roi-calculator',
                     '/product',
+                    // The partner pages are .vue files; only the listing is linked from the nav.
+                    '/partners/',
+                    '/partners/certify-hardware/',
+                    '/partners/ctrlx/',
+                    '/partners/referral-sign-up/',
```
ZJvandeWeg, line 435: @dgatti0213 Do you handle the inputs on that page? Or should it be removed?

scripts/gen-site-art.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python3
```
ZJvandeWeg: @dimitrieh Let's keep this script in `js` or `ts` too, because requiring more components is adding a maintenance burden.

nuxt/pages/partners/index.vue
```diff
@@ -0,0 +1,114 @@
+<script setup lang="ts">
+// Ported from src/partners/index.njk (11ty), which this replaces. Same page, same copy,
+// same classes from src/css/style.css.
+//
+// What the port changes on purpose:
+//  - The section headings carried <span class="text-indigo-600"> in frontmatter, printed
+//    with `| safe`. They go through nuxt/lib/rich-text.mjs, which escapes the string and
```
ZJvandeWeg: I've reviewed `rich-text` before, and the key of requiring it is when you're exposing user input to the world. However, all strings that we include are already trusted and safe.

Let's keep things simple and trust our static strings, remove this.

dimitrieh:
Rebased on `main`, with all four review points addressed.

- `legacy.ts:25` — two stray blank lines the diff added, nothing more. Gone.
- `scripts/gen-site-art.py` — rewritten as `scripts/gen-site-art.mjs`, so the repo still needs nothing but node. Verified it emits a byte-identical `nuxt/utils/siteArt.ts` apart from its own filename in the generated header. Ported the same way on every branch that carried it.
- `rich-text` — removed, call sites now plain `v-html`. Agreed on the reasoning: these are repo strings, and the 11ty templates they replace printed them with `| safe`, so the allowlist was not a trust boundary. Done on every branch that added it.
- `/partners/referral-sign-up/` — the port is 1:1 and the form is still HubSpot form `aad1cda7`, so submissions land where they always did. Whether the page stays is a content call, not a migration one; if it should go, a deletion plus a redirect is cleaner as its own change than dropping it from this port.

One note on `rich-text` worth a second look: the `/blueprints/` descriptions in #5794 come from READMEs in `FlowFuse/blueprint-library` rather than from this repo. They are still `v-html` there, matching the `| safe` they had, on the basis that library submissions are reviewed before they land. Flagging it because it is the one call site whose content is not authored in this repo.

Also from the review pass: a bullet on `certify-hardware.vue` described a `</h1>` after an `<h2>` in the `.njk`. That file has no `<h2>`; the defect is `referral-sign-up.njk`'s and is already documented on `referral-sign-up.vue`. Removed. Same for the `ctrlx.vue` bullet claiming the layout emitted an empty second `<h1>`: `layouts/page.njk` guards the hero on `{% if (title and not hideTitle) %}` and that `.njk` sets no `title`, so no heading was emitted at all.

nuxt/server/middleware/legacy.ts
dimitrieh: No reason. Two stray blank lines the diff added. Gone.

nuxt/nuxt.config.ts
dimitrieh: The port is 1:1 and the form is still HubSpot form `aad1cda7`, so submissions land where they always did. Whether the page stays is a content call rather than a migration one. If it should go, a deletion plus a redirect is cleaner as its own change than dropping it from this port.

scripts/gen-site-art.py
dimitrieh: Done, it is `scripts/gen-site-art.mjs` now. Verified it emits a byte-identical `nuxt/utils/siteArt.ts` apart from its own filename in the generated header. Ported the same way on every branch that carried it, so the repo needs nothing but node.

nuxt/pages/partners/index.vue
dimitrieh: Agreed, removed. The call sites use plain `v-html`, and the 11ty templates they replace printed those strings with `| safe`, so the allowlist was not a trust boundary. Done on every branch that added it.

One call site is worth a second look: the `/blueprints/` descriptions in #5794 come from READMEs in `FlowFuse/blueprint-library` rather than from this repo. They are `v-html` too, on the basis that library submissions get reviewed before they land, but that is the one place the content is not authored here.

nuxt/utils/siteArt.ts
Yndira-E, line 1: @dimitrieh Most of these keys are literal Heroicons v2 names (`bolt`, `shield-check`, `check-badge`, …), and `@iconify-json/heroicons` is already a dependency. Could we use `<UIcon name="i-heroicons-x" />` directly for those instead of inlining raw SVG? It already handles `currentColor` and sizing, which is the reason given here for not using ``/`NavIcon`. Only the genuinely custom icons (`uns`, `wind-turbine`, `role-based-access`, …) would need a real local solution.

dimitrieh: Done in 1690d0a51, rebased on main.

Since #5792 removed `SiteArt` and `gen-site-art.mjs` from main, this PR no longer adds them either. `certify-hardware.vue` uses `<Icon name="i-heroicons-…">` at the call sites, same as the homepage. All five keys here were stock Heroicons, so nothing on /partners/ needed a local fallback.

One detail worth carrying to the other pages: `.ff-certify-list-icon` is `w-[24px]` with no height of its own. The `<svg>` filled it through its intrinsic ratio, but `<Icon>` renders a masked `<span>`, which has none, so `w-full h-full` collapses to zero height there. These are sized `w-6 h-6` instead. Checked before and after: same glyphs, same 24x24, same colour.

Yndira-E: Please check out https://github.com/FlowFuse/website/pull/5803, let's use `UIcon` instead.

nuxt/nuxt.config.ts
dgatti0213: Content on the pages was before me, It probably needs some review from me. Port for now is fine tho.

dimitrieh:

nuxt/pages/partners/certify-hardware.vue
```diff
@@ -0,0 +1,96 @@
...
+            </ul>
+          </div>
+        </div>
+        <hr class="w-full md:w-[768px] mx-auto">
+        <div class="container max-w-screen-md m-auto flex flex-col gap-4">
+          <h3 class="m-auto text-center w-full mb-6">Marketing and Publications</h3>
+          <div class="ff-certify-list">
+            <div class="ff-item">
+              <div class="ff-line-container sm:invisible"><div class="ff-line" /></div>
+              <div class="ff-certify-list-icon"><Icon name="i-heroicons-chat-bubble-bottom-center-text" class="w-6 h-6" /></div>
```
Yndira-E: https://github.com/FlowFuse/website/pull/5784#discussion_r4007059302
```suggestion
              <div class="ff-certify-list-icon"><UIcon name="i-heroicons-chat-bubble-bottom-center-text" class="w-6 h-6" /></div>
```

Yndira-E: @dimitrieh please apply this to all the `<Icon>` elements

dimitrieh:
Nothing to change here, one note. `/partners` was never added to `NUXT_PREFIXES` in `nuxt/server/middleware/legacy.ts`, so `npm run dev` keeps proxying it to 11ty and serves the old page. That middleware returns early unless `NODE_ENV === 'development'`, which is why CI and the deploy preview both look correct and this went unnoticed.

dimitrieh:
Correction to my earlier comment: #5815 is closed, the registration belongs in each migration PR instead. Pushed here.

nuxt/nuxt.config.ts
dimitrieh: Leaving this one open, it needs an answer from whoever owns that form rather than from me. Nothing in this PR changes that page beyond the port; the only thing pushed here since review is the `/partners` registration in `legacy.ts`.

Yndira-E:
@dimitrieh this hasn't been implemented yet https://github.com/FlowFuse/website/pull/5784#discussion_r4018487477

dimitrieh:
Done. The two remaining `<Icon>` call sites in `certify-hardware.vue` are `<UIcon>` now, matching the suggestion you had already applied to the third. Same component underneath, so nothing changes visually.

Each keeps its `w-6 h-6`: `.ff-certify-list-icon` sets a width and no height, and the masked span has no intrinsic ratio, so without it they collapse.

That was the last `<Icon>` in the PR, and there is no inline `<svg>` left in the partner pages either.

nuxt/pages/partners/certify-hardware.vue
dimitrieh: Done here, and I swept the rest of the series rather than just this PR.

`<Icon>`, `<NuxtIcon>` and `SiteArt` are all at zero on `main` and on every open migration branch: #5780, #5781, #5782, #5784, #5785, #5786, #5793, #5794.

The only `<Icon>` left anywhere is `components/DifferentiatorCards.vue` on #5795. It needs nothing: #5795 does not touch that file, it inherits it from `nuxt/eleventy-teardown-base`, which is 55 commits behind `main` and predates `Standardize on <UIcon>, drop stray <Icon> usages`. Retargeting #5795 to `main` once 1-17 have landed, which the plan in #5777 already calls for, picks up the `<UIcon>` version on its own.

Inline `<svg>` counts are unchanged from `main` on every branch, so none of these PRs introduces a raw glyph that should have been an icon.

Yndira-E (approved):

dimitrieh:

PR #5785, opened 2026-09-11
events: serve /events/ from Nuxt

dimitrieh:
## Description

Moves `/events/` to Nuxt: ProveIt 2026 and both Hannover Messe pages.

Their art is now a tracked asset rather than something Eleventy copied at build time.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5785--flowforge-website.netlify.app/events/proveit-2026/
- https://deploy-preview-5785--flowforge-website.netlify.app/events/hannover-messe-2025/
- https://deploy-preview-5785--flowforge-website.netlify.app/events/hannover-messe-2026/

ZJvandeWeg:
@Yndira-E I recommend we merge this, when @dimitrieh fixed the tests, but then afterwards merge `/webinars` and `/events`. Webinars are just virtual events IMO.

dimitrieh:
Rebased on `main` and fixed the failing build.

The prerender 500 on `/events/hannover-messe-2026/` was a missing auto-import: the page calls `useScrollToAnchor()`, and `nuxt/composables/useScrollToAnchor.ts` only exists on the `/use-cases/` branch. Nuxt auto-imports mean there is no import line to break, so it compiles and throws at prerender. The composable is carried here too, byte-identical to the copy in #5780, so whichever lands first the other merges cleanly.

From the review pass over the whole migration:

- Adding `/events` to `NUXT_PREFIXES` stopped `src/events/images/**` and `/events/hm25-invite.ics` resolving in dev: those only reach `nuxt/public/` through the 11ty passthrough in a production build. Added a carve-out beside the two this file already has for `/resources/images/` and the blog and changelog image paths. Production was never affected.
- A header comment on `hannover-messe-2025.vue` still described `rich-text.mjs` escaping the copy strings. It is gone, and the strings are page-local consts rendered with `v-html`; the comment said so.

`nuxt/lib/rich-text.mjs` is dropped and `scripts/gen-site-art.py` is now `scripts/gen-site-art.mjs`, both per the review on #5784.

On merging `/webinars` and `/events`: agreed they are the same thing. Worth doing as its own change once both are on Nuxt, since it is an IA decision with redirects attached rather than part of the port.

dimitrieh:

dimitrieh:

dimitrieh:
Nothing to change here, one note. `/events` was never added to `NUXT_PREFIXES` in `nuxt/server/middleware/legacy.ts`, so `npm run dev` keeps proxying it to 11ty and serves the old page. That middleware returns early unless `NODE_ENV === 'development'`, which is why CI and the deploy preview both look correct and this went unnoticed.

dimitrieh:
Correction to my earlier comment: #5815 is closed, the registration belongs in each migration PR instead. Pushed here. It also carries the carve-out for `/events/images/` and the Hannover `.ics`, which stay in `src/`.

dimitrieh:
Took the cross-PR pass over the four patterns here. Three of them applied:

- **Custom icon files.** `arrow-small-down`, `calendar` and `arrow-top-right-on-square` are stock Heroicons, so they are `<UIcon>` now and `SiteArt.vue`, `siteArt.ts` and `gen-site-art.mjs` are gone. Each call site carries its own size: `UIcon` renders a mask span with no intrinsic ratio, so the two width-only wrappers (`<div class="w-5">`) would have collapsed it.
- **`useScrollToAnchor`.** Both anchors already carry the href they scrolled to (`#special-sessions`, and `secondaryCtaTarget` is `#flowfuse-talks-agenda`), so the handler only preventDefault'd to do what the browser does anyway. Removed, the global `scroll-behavior: smooth` animates it.
- **`legacy.ts`.** `/events` registered, plus the carve-out for `/events/images/` and `hm25-invite.ics`, which stay in `src/` and only reach `nuxt/public` via the passthrough in a production build.

**Leftover 11ty layouts**: none here. Checked which layouts this branch makes dead relative to main rather than which are dead on it, since `blueprint`, `common-js`, `mql-contact` and `toc` already have no consumers on main and are not this PR's to remove.

Worth a look on the preview: `arrow-small-down` is the 20x20 solid Heroicon (`-20-solid`), not the 24 outline, so it is the mini glyph.

dimitrieh:

PR #5783, opened 2026-09-11
vs: serve /vs/ from Nuxt

dimitrieh:
## Description

Moves `/vs/` to Nuxt. The comparison pages were frontmatter read by one layout, so they become a content collection and a single `[slug].vue`.

## Related Issue(s)

Part of https://github.com/FlowFuse/website/issues/5777

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [ ] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5786, opened 2026-09-11
landing: serve /landing/ from Nuxt

dimitrieh:
## Description

Moves `/landing/` to Nuxt. The campaign pages that were pure frontmatter become a content collection; the four with their own layout become `.vue` pages.

The HubSpot meeting embed stays behind the analytics-consent gate: it registers the same global the consent banner calls, rather than loading on mount.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/tulip/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/plc/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/factory-efficiency/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/building-and-scaling-industrial-applications/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/coordinating-industrial-systems-at-scale/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/edge-connectivity/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/enterprise-integration/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/line-control/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/plant-orchestration/
- https://deploy-preview-5786--flowfuse-website.netlify.app/landing/unified-real-time-data-platform/

dimitrieh:
Rebased on `main` and fixed the failing build.

The prerender 500s on `/landing/tulip/` and `/landing/factory-efficiency/` were a missing auto-import: both call `useScrollToAnchor()`, and `nuxt/composables/useScrollToAnchor.ts` only exists on the `/use-cases/` branch. Nuxt auto-imports mean there is no import line to break, so it compiles and throws at prerender, on exactly the pages that call it. The composable is carried here too, byte-identical to the copy in #5780, so whichever lands first the other merges cleanly.

From the review pass over the whole migration:

- Adding `/landing` to `NUXT_PREFIXES` stopped `src/landing/images/**` resolving in dev: those 13 files only reach `nuxt/public/` through the 11ty passthrough in a production build. Added a carve-out beside the two this file already has for `/resources/images/` and the blog and changelog image paths. Production was never affected.
- The seven `[slug].vue` pages were in neither sitemap (a `data` collection is invisible to the module's static-route discovery and to `content-urls.get.ts`), having left `sitemap-legacy.xml` when their `.njk` files were deleted. Added to `sitemap.urls`, minus the `skipIndex` entries, which are noindex.
- Three `<Cta* ... icon />` rendered no icon: `icon` is typed `string`, so a value-less attribute is `""`, not `true`.
- `HubSpotMeetings.vue` had replaced `<CtaCustom destination-key="hubspotMeeting">` with a hand-written `<a>` that re-typed the URL and the event name already held in `nuxt/lib/custom-cta-destinations.ts`. That registry exists so one destination cannot drift between call sites; `CtaCustom` restored.

`nuxt/lib/rich-text.mjs` is dropped and `scripts/gen-site-art.py` is now `scripts/gen-site-art.mjs`, both per the review on #5784.

dimitrieh:

nuxt/utils/siteArt.ts
Yndira-E, line 1: @dimitrieh please remove these siteArt files, let's keep it consistent with the approaches we've taken on previous PRs

dimitrieh: Removed: `SiteArt.vue`, `siteArt.ts` and `gen-site-art.mjs` are all gone, call sites are `<UIcon>`.

Of the 59 keys these pages use, 48 are stock Heroicons. Three were the same glyph under another name, checked against the set path by path: `chip` is `cpu-chip`, `chart` is `chart-bar-square`, `chevron-right-sm` is `chevron-right` at `w-4`.

The other eight have no Heroicons equivalent (`uns`, `layers`, `pin`, `pin-slash`, `pulse`, `snowflake`, `target-view`, `arrows`) and nearly all arrive as strings from the `.yml`, so they are a local `ff` collection in `nuxt/assets/icons/` that `UIcon` reads by name. The content stays uniform that way, `i-heroicons-x` next to `i-ff-x`, with nothing left to resolve a bare key.

Each call site carries its own `w-`/`h-` now, per your note on #5784: the raw `<svg>` filled its wrapper through the intrinsic ratio, and one wrapper here set only `w-6`.

dimitrieh: Verified on the preview, CI green.

All 59 glyphs render at the same size and position as live. One visible difference, and it is a fix rather than a regression: the `uns` icon in the right-hand card renders as a solid red blob on live, because the raw inline SVG carried a `clip-path` referencing an internal id. Iconify normalises that, so the preview shows the actual glyph.

Compared `/landing/building-and-scaling-industrial-applications/` against live: same 976x706 box, same 14 glyphs, otherwise pixel-identical. Four of the custom ones (`snowflake`, `pin-slash`, `target-view`, `layers`) paint via `background-image` rather than `mask-image` because their SVGs have hardcoded colours, which is the same reason they never followed `currentColor` under `SiteArt` either.

Also corrected the preview links in my earlier comment: they pointed at `flowforge-website`, which now 404s after the Netlify project rename.

nuxt/composables/useScrollToAnchor.ts
Yndira-E, line 1: @dimitrieh I'd say we don't need this either, `scroll-behavior: smooth` is already global, defined in style.css

nuxt/public/images/walter-icon.png
Yndira-E, line 1: @dimitrieh it seems like the customer stories assets are being duplicated here... I don't see them removed from the 11ty folder, and if we did that, then it would also affect the customer stories pages. Moving them now seems out of scope for the landing pages migration.

dimitrieh:
Pushed:

- The three CTAs already carry `href="#form"` and the targets exist, so the handler only had to `preventDefault` to do what the browser does anyway. They keep their `capture()` call and lose the scroll, which retires the composable.
- On the duplicated assets: seven of the images copied into `nuxt/public` here are referenced by nothing in the Nuxt tree, and `src/` already has all seven, which 11ty passthrough-copies to the same URLs. So the copies just go. Nothing moves out of the 11ty tree and the customer-stories pages are untouched, which keeps it in scope. The seven are `arch-icon.png`, `home/home-arch.png`, `home/pidd.png`, `home/power.png`, `pidd-icon.png`, `stories/walter.png` and `walter-icon.png`.

That matches what an already-merged Nuxt page does: `/images/stories/un-wmo.jpg` is referenced from `nuxt/` on main and is not in `nuxt/public`.

The `/landing` registration and the `/landing/images/` dev carve-out those pages need are in https://github.com/FlowFuse/website/pull/5815.

dimitrieh:
Correction to my earlier comment: #5815 is closed, the registration belongs in each migration PR instead. Pushed here. It also carries the `/landing/images/` carve-out, since those images stay in `src/`.

nuxt/composables/useScrollToAnchor.ts
dimitrieh: Removed. All three CTAs already carry `href="#form"` with a matching target, so the handler was only calling preventDefault to do what the browser does anyway. They keep their `capture()` call and the global `scroll-behavior: smooth` animates the jump.

nuxt/public/images/walter-icon.png
dimitrieh: Agreed that moving them is out of scope, so nothing moved out of the 11ty tree. The copies go instead: seven images this PR added under `nuxt/public` are referenced by nothing in the Nuxt tree, and `src/` already has all seven at the same URLs via the passthrough. That is `arch-icon.png`, `home/home-arch.png`, `home/pidd.png`, `home/power.png`, `pidd-icon.png`, `stories/walter.png` and `walter-icon.png`. Customer-stories pages are untouched. It also matches what an already-merged Nuxt page does: `/images/stories/un-wmo.jpg` is referenced from `nuxt/` on main and is not in `nuxt/public`.

Yndira-E:
The icon for the secondary cta is not properly aligned and on mobile it becomes stacked

Screenshot 2026-09-16 at 12 32 38

dimitrieh:
Fixed, and both symptoms had one cause.

The `.njk`'s `extraClass` came over verbatim, but it was written against a different DOM. `cta-button.njk` wrapped the label and the arrow in a single `<span class="inline-flex items-center gap-2">`, so the `<a>` had one child and `inline` / `flex flex-col` changed nothing that was drawn. `UButton` renders the label and its `trailing-icon` as siblings, so `flex flex-col` stacks them and `inline` drops the flex context that centres the arrow. Copying the strings across was faithful to the source and wrong on the page.

Dropping the two classes restores `UButton`'s own `inline-flex`. The mobile CTA keeps `w-full` and its margins.

`pages/vs/[slug].vue` carries the same string from #5783 and is already on `main`, so `/vs/ignition/`, `/vs/kepware/` and `/vs/litmus/` stack their mobile CTA on live today. Same one-line fix, included here because it is the same cause.

I checked the other migration branches for the pattern. The rest are `flex flex-col` on a CTA with no icon, where there is only one child and nothing stacks, so I left those alone. The tell is `icon` plus `flex flex-col` together.

Yndira-E (approved):

PR #5790, opened 2026-09-11
careers: redirect /careers/ to the job board

dimitrieh:
## Description

`src/careers.njk` was a page whose only content was a `<script>` calling `location.replace(site.jobBoard)`. This replaces it with a Nitro route rule: no blank page while the script runs, and it works with JavaScript disabled. 302 rather than 301, because the board's address is not a permanent property of this URL.

`src/404.njk` goes with it. `nuxt/error.vue` already renders that page, down to the same pictogram and copy.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5790--flowforge-website.netlify.app/careers/

This one redirects off-site to the job board, so the check is that it 302s rather than rendering.

src/404.njk
Yndira-E, line 1: @dimitrieh I haven't gone through all the PRs yet, but why is this page removed on this unrelated PR?

dimitrieh: `nuxt/error.vue` renders it, and has since #5715. Same pictogram, same copy, same button.

Production confirms it: an unknown URL returns 404 with that page.

It was in this PR because both files were 11ty pages Nuxt had already replaced, so they were dead for the same reason. The description says that, the title does not. It should have been split out.

PR #5791, opened 2026-09-11
about: serve /about/ from Nuxt

dimitrieh:
## Description

Moves `/about/` to Nuxt: the team grid, values, benefits and investors.

The team grid reads the `team/` data directory only. The merged lookup used for blog bylines also includes `guests/`, and three guest files are former team members still carrying the `order` they had as staff, so reading the merged one puts them back on the page.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5791--flowforge-website.netlify.app/about/

dimitrieh:
Rebased on `main`. No content change.

One comment correction: the page comment claimed the 11ty layout rendered an empty second `<h1>`. It did not. `layouts/page.njk` guards the whole hero on `{% if (title and not hideTitle) %}` and `src/about.njk` sets no `title`, so the layout emitted no heading at all. Bullet removed rather than left crediting the port with fixing a defect that never existed.

dimitrieh:

dimitrieh:

Yndira-E:
@dimitrieh the pictograms are the wrong size, they're too big. Compare the preview against the live version and you'll see the difference.

Also, the email icon under each team member's picture should be filled and it has a weird spacing right now.

nuxt/server/middleware/legacy.ts
```diff
@@ -19,7 +19,7 @@ const NUXT_PREFIXES = ['/handbook', '/ebooks', '/whitepaper', '/pricing', '/docs
 
 // Top-level routes still on 11ty, not yet ported to Nuxt (everything not listed above
 // already falls through to the 11ty proxy by default). Remove entries here as they migrate:
-// /about, /blueprints, /careers, /community, /events
+// /blueprints, /careers, /community, /events
```
Yndira-E, line 22: @dimitrieh /about needs to be added to the NUXT_ROUTES, otherwise, it won't load on local dev

nuxt/pages/about/index.vue
```diff
@@ -0,0 +1,272 @@
...
+                  scalable platform.
+                </p>
+                <p>
+                  <span class="font-medium">FlowFuse seamlessly integrates into existing industrial environments,</span>
+                  offering an adaptable foundation for building bespoke applications. Our customers rely on FlowFuse to
+                  develop innovative applications that modernize their industrial operations creating the opportunity to
+                  improve their processes continuously.
+                </p>
+                <p>
+                  <NuxtLink to="/platform/security/#certifications" class="text-indigo-600 hover:text-indigo-800 font-medium">See our security certifications</NuxtLink>
```
Yndira-E: @dimitrieh this link is broken

nuxt/pages/about/index.vue
```diff
@@ -0,0 +1,272 @@
...
+
+useSeoMeta({
+    title: 'About',
+    description: META_DESCRIPTION,
+    ogDescription: META_DESCRIPTION,
+    ogUrl: 'https://flowfuse.com/about/',
+    twitterSite: '@FlowFuseinc',
+})
+
+useSchemaOrg([
```
Yndira-E, line 67: @dimitrieh `app.vue` already declares a sitewide `defineOrganization()`. Some fields here overlap with it so the rendered JSON-LD on `/about`/ ends up with duplicate entries. Worth trimming this block to just the fields app.vue doesn't already have.

nuxt/server/middleware/legacy.ts
dimitrieh: Added `/about` to `NUXT_ROUTES`. It had only been dropped from the informational list in the comment block below, which the middleware does not read.

nuxt/pages/about/index.vue
dimitrieh: Fixed. It was a `<NuxtLink>`, so the click was resolved client-side by vue-router, matched `pages/[...slug].vue`, found no content entry and 404'd. `/platform/security/` is still 11ty, so it needs a plain `<a href>`.

nuxt/pages/about/index.vue
dimitrieh: Trimmed to the fields `app.vue` does not carry: `legalName`, `description`, `foundingDate`, `founder`, `address`.

One correction on the mechanism: it resolved to a single `#identity` node, not duplicate entries, because both calls dedupe on the same `@id`. The duplication was inside it. `url` was overwritten with the `www` host, and `sameAs` concatenated into a list holding GitHub twice and LinkedIn and Twitter in two spellings each. The logo block also declared 396x215 for a file that is 529x287, so that goes too.

Facebook and YouTube were only in the page list, so they move to `app.vue` rather than being dropped.

dimitrieh:
Both fixed.

**Pictograms.** They carried `w-full h-auto`. Tailwind v4 resolves `utilities` after `components`, where `style.page.css` lives, so `w-full` beat `.company-value img { width: 128px }` and `.about-values .company-value img { width: 60px }` despite being less specific. Measured against live: benefits were 299px instead of 128px, values 121-171px (each tracking its own PNG aspect ratio) instead of a uniform 60px. The 11ty markup carried no classes, so dropping them restores both.

**Email icon.** It was a `<UIcon>`, which renders a masked `<span>` with no intrinsic aspect ratio, so the link laid out 36x20 against the 20x20 the siblings get. That is the spacing. Lucide's mail is also an outline where the partial was solid. Mail and RSS now have components next to the other social icons, carrying the same paths the partials had.

Yndira-E (approved):
I’m approving this. There’s only one small issue, which is also present in the live version: when one card is taller than the others, the other cards don’t adjust, and on hover, the hidden card slides behind the empty space at the bottom.

Image

If you could fix it here, that would be great. Otherwise, we can address it in a follow-up PR. This isn’t a blocker, so I’ll leave it up to you.

dimitrieh:
Checked the three open threads against the branch and all three are already addressed, so nothing pushed here. They just need resolving.

- `/about` is in `NUXT_ROUTES`.
- The security link is a plain `<a href="/platform/security/#certifications">`.
- The `defineOrganization()` block is trimmed to `legalName`, `description`, `foundingDate`, `founder` and `address`.

This is the one migration PR that registered its own route, so it is not part of https://github.com/FlowFuse/website/pull/5815.

dimitrieh:
Fixed it here rather than a follow-up, since it turned out to be two lines.

The grid stretches every card to the tallest in its row, but the card's children do not: the headshot holds its aspect ratio and `.member-info` its natural height. So a card whose role fits on one line stopped about 24px short and left the bottom of the box open, and on hover the facts panel filled only that shorter box while the card outline carried on below it. That is the sliding-behind you saw.

Laying `.team-headshot` out as a column and letting `.member-info` take up the slack closes it. 4 of the 22 cards were affected. Headshot dimensions are unchanged, which I checked rather than eyeballed.

Before, with the middle card hovered. Yndira's card has a two-line role so it sets the row height; the other two stop short:

before

After:

after

The rule lives in `src/css/style.page.css`, which the teardown moves to `nuxt/assets/css/style.page.css`, so it survives that PR.

PR #5792, opened 2026-09-11
homepage: serve / from Nuxt

dimitrieh:
## Description

Moves the homepage to Nuxt. The cross-fading hero, the social proof marquee, the testimonial carousel and the FAQ become components.

The hero's second button keeps its white treatment, so it stays legible against the hero image.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5792--flowforge-website.netlify.app/

scripts/gen-site-art.py
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env python3
```
ZJvandeWeg: He dude, if we could do this in `node`, that would be much better. Now we're not depending on python3 for the website. After this PR, we are?

Would be better to keep it contained.

dimitrieh:
Rebased on `main`. `scripts/gen-site-art.py` is now `scripts/gen-site-art.mjs`, so the repo still needs nothing but node. Verified it produces a byte-identical `nuxt/utils/siteArt.ts` apart from its own filename in the generated header, and ported the same way on every branch that carried the script.

One fix from the review pass over the whole migration: the two `<CtaSignUp ... icon />` on the homepage rendered no icon. `icon` is typed `string`, so a value-less attribute resolves to `""`, not `true`, and `UButton` gets an empty `trailing-icon`. Every other call site in the repo writes `icon="i-lucide-arrow-right"`.

scripts/gen-site-art.py
dimitrieh: Right, and it is node now: `scripts/gen-site-art.mjs`. Verified it emits a byte-identical `nuxt/utils/siteArt.ts` apart from its own filename in the generated header. Ported the same way on every branch that carried the script, so nothing here depends on python3.

Yndira-E:
Pushed a commit with two fixes:

- **Hero CTA arrow wrapping to a 2nd line**: `flex flex-col mb-6` was copied from the 11ty `<a>` wrapper, but on `CtaButton` that class lands on the same element that lays out icon+label in a row — `flex-col` flipped it to a column. Removed (it was vestigial anyway).
- **Dropped `SiteArt`**: the 6 icons it rendered here are byte-identical to Heroicons, so swapped to `<Icon i-heroicons-*>` (`arrow-long-right` → `-solid` variant) and removed `SiteArt.vue`/`siteArt.ts`/`gen-site-art.mjs` since nothing else used them. A truly custom icon can still get a one-off SFC in `nuxt/components/icons/`same as `GithubIcon.vue`.

Yndira-E (approved):
This is the current state of the homepage bottom section since both blog articles and webinars are now in nuxt
Image

I'm merging now to fix this. If the other reviewers have feedback, that can be address in a follow up pr

dimitrieh:
Following up on the note about other reviewers' feedback going to a later PR. The one thread still open here asked for `scripts/gen-site-art.py` to be node rather than python3, so the site does not depend on python3.

It is resolved across the board now, not just renamed: `gen-site-art.mjs` is deleted outright on every remaining migration PR, along with `SiteArt.vue` and `siteArt.ts`, with the glyphs going through `<UIcon>` instead. That is #5780, #5781, #5785, #5786, #5793 and #5794.

So nothing depends on python3 for the website, and there is no generator script left to depend on either.

PR #5793, opened 2026-09-11
node-red: serve /node-red/ from Nuxt

dimitrieh:
## Description

Moves `/node-red/` to Nuxt.

The certified nodes grid is derived from the two certified catalogue feeds the integrations code already fetches, instead of a second copy of that fetch. Names come from the package id, so acronyms print as `OPC UA` and `RTSP` rather than lowercase slugs.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5793--flowforge-website.netlify.app/node-red/

dimitrieh:
Rebased on `main`. `scripts/gen-site-art.py` is now `scripts/gen-site-art.mjs` and `nuxt/lib/rich-text.mjs` is gone, both per the review on #5784.

Three fixes from the review pass over the whole migration:

- A four-line Nunjucks `{# ... #}` comment survived inside the Vue `<template>`. Vue has no `{# #}` syntax, so it was a static text node and would have printed verbatim above the certified-nodes section on the live page. Moved into `<script setup>` beside the fetch it describes (it was also stale: it named `src/_data/certifiedNodes.js`, which this PR retires).
- `legacy.ts` still carried `normalised !== '/node-red'` and a comment saying `/node-red` "keeps falling through to 11ty: it is still a marketing page there". This PR puts `/node-red` in `NUXT_PREFIXES` and deletes `src/node-red/index.njk`, so the guard was dead and the comment inverted. Both gone.
- `fetchCatalogueFeed`'s bare `catch { return [] }` said nothing. `/node-red/` is prerendered now, so a transient failure on either certified-nodes feed at build time bakes a page with that whole section missing, and it stays missing until the next deploy. Still degrades to a hidden section rather than a broken page, but the build log now names the feed that failed.

Also dropped a page-comment bullet claiming the 11ty layout emitted an empty second `<h1>`: `layouts/page.njk` guards the hero on `{% if (title and not hideTitle) %}` and `src/node-red/index.njk` sets no `title`, so no heading was emitted at all.

dimitrieh:

dimitrieh:

dimitrieh:
Nothing to change here, one note. `/node-red` was never added to `NUXT_PREFIXES` in `nuxt/server/middleware/legacy.ts`, so `npm run dev` keeps proxying it to 11ty and serves the old page. That middleware returns early unless `NODE_ENV === 'development'`, which is why CI and the deploy preview both look correct and this went unnoticed.

dimitrieh:
Correction to my earlier comment: #5815 is closed, the registration belongs in each migration PR instead. Pushed here. It also drops the carve-out that kept `/node-red` itself on 11ty, now that its index is a Nuxt page.

dimitrieh:
Cross-PR pass over the four patterns. Two applied:

- **Custom icon files.** The nine glyphs are stock Heroicons, so the keys in `LOW_CODE` and `WHY` are iconify names now and the call sites are `<UIcon>`. `SiteArt.vue`, `siteArt.ts` and `gen-site-art.mjs` are gone. Sizes are explicit per call site, since the mask span has no intrinsic ratio and both wrappers are `w-8 h-8`.
- **`legacy.ts`.** `/node-red` registered. Its index is a Nuxt page now, so the carve-out that deliberately kept just `/node-red` on 11ty while `/node-red/*` went to Nitro is gone too.

One deliberate exception: `certified-node` does not become a `<UIcon>`. It is Heroicons `check-badge` solid, but `IntegrationsCertifiedIcon` already renders it and `.certified-icon` styles it with `fill` and `stroke` plus a size override inside `.certified-eyebrow`, none of which apply to a mask span. Reusing that component keeps it rendering as it does today.

No `useScrollToAnchor` here, and no layouts this PR makes dead relative to main.

dimitrieh:

ZJvandeWeg:
@macroscope-app review

dimitrieh:
@Yndira-E @KristopherLeads PR conflicts are resolved.

ZJvandeWeg:
@dimitrieh They are back :) But also check out: https://github.com/FlowFuse/website/pull/5793#issuecomment-5800730402

PR #5794, opened 2026-09-11
blueprints: serve /blueprints/ from Nuxt

dimitrieh:
## Description

Moves `/blueprints/` to Nuxt: the paginated library, the detail pages and the submission page.

The blueprints come from the separate `FlowFuse/blueprint-library` repository, so the build step that pulls them in moves too. `nuxt/lib/blueprints-sync.mjs` writes the markdown into `nuxt/content/blueprints/` and the screenshots into `nuxt/public/blueprints/`; a Nuxt module runs it during a build or dev start, the way the product docs already work. `npm run blueprints` keeps its name, so the Build Site workflow is unchanged apart from the path it commits.

The library is private, so the sync resolves its source in precedence order: `FLOWFUSE_BLUEPRINTS_LOCAL`, a sibling checkout, an authenticated sparse clone, then whatever is already on disk. Only a production deploy that resolves nothing is fatal.

Rebased onto https://github.com/FlowFuse/website/pull/5694, which landed on 14 September and owns the authenticated clone path.

The listing is ordered by path, descending. Eleventy ordered by date and reversed, but no blueprint README carries a date, so they all fell back to their checkout timestamp and the tie broke on path. The one blueprint that does set a date sorted to the very end there and now sits with its category.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

- https://deploy-preview-5794--flowforge-website.netlify.app/blueprints/
- https://deploy-preview-5794--flowforge-website.netlify.app/blueprints/2/
- https://deploy-preview-5794--flowforge-website.netlify.app/blueprints/submit/
- https://deploy-preview-5794--flowforge-website.netlify.app/blueprints/manufacturing/oee-calculator/
- https://deploy-preview-5794--flowforge-website.netlify.app/blueprints/flowfuse-dashboard/multi-user-dashboard/
- https://deploy-preview-5794--flowforge-website.netlify.app/blueprints/ai/llm-chat-agent/

The other 15 blueprint detail pages share the same template.

dimitrieh:
Rebased on `main`, squashed to one commit: the branch's first commit carried leftover conflict markers in `legacy.ts` that its follow-up removed, and replaying that through a rebase reproduced them. The tree is unchanged from the branch head.

`scripts/gen-site-art.py` is now `scripts/gen-site-art.mjs` and `nuxt/lib/rich-text.mjs` is gone, per the review on #5784. The blueprint descriptions are still rendered with `v-html`, matching the `| safe` they had, but they come from READMEs in `FlowFuse/blueprint-library` rather than from this repo, so they are the one call site whose content is authored elsewhere. Raised on #5784 as well.

Four fixes from the review pass over the whole migration, all in the sync:

- A library that resolves but holds no blueprints crashed with a raw `ENOENT` writing the manifest, after `writeBlueprints` had already deleted both published trees. Only the per-blueprint loop recreated `contentDir`, so the graceful "no blueprints" path in `modules/blueprints-source.ts` was unreachable. It now creates the directory after clearing it, with a test.
- The sync deletes `nuxt/content/blueprints` and `nuxt/public/blueprints` before reading the source, and zero entries is the only fatal case. A partially-resolved library (a README renamed upstream, a restructured category) therefore unpublished live pages on a green build with nothing in the log, and the Build Site workflow commits and force-pushes whatever it leaves. It now reads the published set first and names anything being dropped, and names any directory skipped for a missing README. Two tests.
- `injectUpdated` prepended `updated:` unconditionally, so a README that sets its own produced a duplicate YAML key, which fails the schema and drops the page silently.
- `stripLayout`'s `/^layout:.*$/m` ran over the whole file, so it could delete a body line beginning `layout:` when the frontmatter had no such key. It splits frontmatter first now, as `rewriteFrontmatterImage` does.

Also moved the `stories` collection comment back above `stories:` — inserting the `blueprints` collection had left it describing the wrong collection.

This still needs rebasing onto #5694 once that lands, per the note in #5777.

dimitrieh:

nuxt/components/ContactUsCtaLine.vue
```diff
@@ -0,0 +1,10 @@
+<script setup lang="ts">
```
ZJvandeWeg, line 1: I thought we already had a CTA component?

nuxt/components/CtaContactUsLine.vue
```diff
@@ -0,0 +1,10 @@
+<script setup lang="ts">
```
dimitrieh: We do, but they are all buttons through `cta/CtaButton.vue`. This one is the one-line text band with an inline link, ported from `contact-us-cta-line.njk`, so there is no button to share. Renamed it to `CtaContactUsLine.vue` so it sits with the rest of the family.

The wider point stands though: the teardown branch ends up with thirteen CTA-ish components. Consolidating them touches pages across five of the open migration PRs, so better as its own PR once 5777 is done.

dimitrieh:
There is `CtaContactUs`, but it is a button: it goes through `cta/CtaButton.vue`, takes a variant and a position, and fires a tracked event. This one is the band from `contact-us-cta-line.njk`, an inline link inside a sentence, so it stays separate.

Pushed a change so it at least reads its href from `CTA_DESTINATIONS` like the other `Cta*` components. It stays untracked because the `.njk` it ports was untracked too, so adding an event here would be a behaviour change rather than a port. https://github.com/FlowFuse/website/pull/5749 is the right place for that if we want it.

The `/blueprints` registration is in https://github.com/FlowFuse/website/pull/5815.

dimitrieh:
Correction to my earlier comment: #5815 is closed, the registration belongs in each migration PR instead. Pushed here.

dimitrieh:
Cross-PR pass over the four patterns. Two applied:

- **Custom icon files.** `chevron-left` and `rocket-launch` are byte-identical to the Heroicons outline glyphs, so they are `<UIcon>` now and `SiteArt.vue`, `siteArt.ts` and `gen-site-art.mjs` are gone. Each call site keeps the size its source SVG had (`w-5 h-5` and `size-6`), since the mask span has no intrinsic ratio.
- **`legacy.ts`.** `/blueprints` registered.

This also closes out the open thread on https://github.com/FlowFuse/website/pull/5792, which asked for `gen-site-art` not to be python3. It became `.mjs` there, and on this branch the script is gone entirely.

No `useScrollToAnchor` here, and no layouts this PR makes dead relative to main.

dimitrieh:
Separate finding from the preview, not from the icon change and not a blocker.

`/blueprints/other/mobile-alerting/` logs `Hydration completed but contains mismatches`. Traced it by diffing the SSR tree against the hydrated DOM: it is the `## Start now` heading, whose markdown is a link.

`@nuxt/content`'s prose heading wraps the heading text in its own `<a href="#start-now">` anchor. When the heading is itself a link, that nests an `<a>` inside an `<a>`, which HTML cannot represent, so the parser splits them when it reads the SSR response while Vue's client render builds the nested structure it intended. The two trees disagree:

- SSR: `<a href="#start-now"></a>` then `<a href="https://app.flowfuse.com/deploy/blueprint?...">Start now</a>`
- Client: `<a href="#start-now">Start now</a>` then the deploy link

11ty never did this, so it only shows up now that these render through `ContentRenderer`.

Sampled the 12 blueprints linked from the index and 1 is affected, so it is narrow: only blueprints whose markdown uses a linked heading. The fix belongs in the blueprint content rather than here (heading and link as separate lines), since the markdown is synced from `blueprint-library` at build time and is not in this repo.

Flagging rather than fixing, as it is pre-existing to my commits and outside this PR.

dimitrieh:
Fix for the hydration mismatch is up: https://github.com/FlowFuse/blueprint-library/pull/182

It was the only linked heading across all 19 blueprints, and the link was dead anyway (`blueprintId=` with no value) while the page already renders its own Deploy button from the frontmatter. Once that merges and the next build syncs, this page stops logging the mismatch. Nothing needed in this PR.

dimitrieh:

PR #5795, opened 2026-09-11
11ty: retire the second generator

dimitrieh:
## Description

Every page is served by Nuxt once PRs 1-17 land, so the Eleventy build has nothing left to generate. This removes it and moves what it still owned into the Nuxt app.

Moved, with no change to any URL:

- `src/{blog,changelog,customer-stories,webinars}` markdown into `nuxt/content/`, so the four collections that still carried `cwd: '../src'` are sourced like every other one.
- The images beside them into `nuxt/public/` at the path they are served from. `@nuxt/content` resolves a relative image against the post's own directory, so these mirror the content tree rather than sitting in one flat folder.
- `src/images/**` into `nuxt/public/images/`. 188 of those were already byte-identical copies there from earlier ports; those are dropped rather than moved.
- `src/public/*` to the root of `nuxt/public/`, which is where Eleventy was copying them.
- `src/_data/*` into `nuxt/data/`, `src/_includes/components/icons` into `nuxt/assets/icons`, `src/css` into `nuxt/assets/css`, `src/_includes/analytics` into `nuxt/server/assets/analytics`.

Also restores the art for `/events/`, `/landing/` and `/vs/` as tracked assets. Those pages were ported earlier in this series, but their images only reached `nuxt/public` through Eleventy's passthrough, so they would 404 on the first build without it.

`nuxt/public` is a tracked asset tree now rather than build output, so its `.gitignore` lists what a build writes instead of the reverse, and `clean:nuxt` removes exactly those paths.

`netlify.toml`'s build command and publish directory are unchanged, and `npm run blueprints` keeps its name, so the Build Site workflow and the shared CI workflow both still work.

**Draft until PRs 1-17 have merged.** Its base is `nuxt/eleventy-teardown-base`, a scaffolding branch holding their combined state so this diff shows only the teardown. Retarget to `main` once they land.

## Related Issue(s)

## Checklist


dimitrieh:
Preview affected pages:

This PR has no deploy preview of its own. Its base is `nuxt/eleventy-teardown-base`, a scaffolding branch, and Netlify builds previews against `main`-targeting PRs.

It changes no rendered page: every route it touches is already served by Nuxt after PRs 1-17. What it moves is where the files live. The check is that the build still produces them, which is what `test_website / Build and check` verifies here.

Once PRs 1-17 have merged and this is retargeted to `main`, its preview is the whole site, and the pages worth a second look are the ones whose assets moved:

- `/blog/2024/07/dashboard-new-charts/` (post images moved to `nuxt/public/blog/`)
- `/changelog/` (same, for changelog entries)
- `/events/hannover-messe-2026/` and `/landing/tulip/` (art that was previously copied by Eleventy)
- `/about/` (team data moved to `nuxt/data/`)
- `/handbook/company/organizational-structure/` (reads the same team data)

dimitrieh:
Root cause of the failing `test_website / Build and check`, for whenever this comes off draft. The build itself succeeds; the link check after it is what exits 1:

```
Checking 365041 links from 6562 files (1314 documents)
Found some broken links, reading source files
Error: IO error for operation on website/src: No such file or directory (os error 2)
```

The step runs `hyperlink website/nuxt/dist/ --check-anchors --sources website/src`, and this PR is the one that deletes `src/`. `--sources` only maps a broken link in built HTML back to the file that produced it, so dropping it does not weaken the check.

It is not fixable in this repo. The argument is hardcoded at line 165 of `.github/workflows/build_website.yml` in `FlowFuse/github-actions-workflows`, and `.github/workflows/test.yml` here pins that workflow at tag `build_website/v1`. `FlowFuse/flowfuse` consumes the same workflow and tag, so moving the tag changes that repo's CI at the same moment. Not touching it.

Two things to expect when it is fixed: `Found some broken links` printed before the crash, so there are real broken links underneath that the `--sources` error masked, and they will surface once the flag is dropped. https://github.com/FlowFuse/website/pull/5796 fails the same way, since it is based on this branch.

PR #5802, opened 2026-09-14
Move heroicons/lucide iconify packages to dependencies

ZJvandeWeg:
## Summary
- Netlify build logs for `main` show repeated `[Icon] failed to load icon` warnings (`heroicons:chevron-left` x31, `heroicons:chevron-down`) during the Nuxt build/prerender.
- `@iconify-json/heroicons` and `@iconify-json/lucide` were listed under `devDependencies` in `nuxt/package.json`; moved them to `dependencies` so the icon collections are reliably available wherever they're resolved at build time.
- Updated the root `package-lock.json` accordingly (only the `dev` flag/location changed, no version bumps).

## Test plan
- [x] Confirm the next Netlify build no longer emits the `[Icon] failed to load icon` warnings
- [x] Spot-check pages using chevron icons (e.g. accordions, changelog listing) render correctly


Yndira-E:
The dependency fix is good practice, but it wasn't what caused the `[Icon] failed to load icon` warnings, those still appeared unchanged (`heroicons:chevron-left` x31, same as reported) even in this PR's own preview before the second commit.

The real fix is `icon.clientBundle.scan` in `nuxt.config.ts` (plus a custom `globInclude` covering `.navigation.yml`, otherwise invisible to the default glob). Confirmed: this preview's build log shows zero `[Icon] failed to load icon` warnings vs. hundreds before.

Yndira-E (approved):

PR #5800, opened 2026-09-14
Add Claude Code and Codex tabs to the agent connector widget

dimitrieh:
## Description

Adds a **Claude Code** tab and a **Codex** tab to the agent connector widget.

A coding agent installs the connector into itself, so each tab is one prompt and nothing else:

> Add the FlowFuse MCP tool at https://app.flowfuse.com/mcp. Then ask me to complete the sign-in in the browser that opens.

No flags and no config file. That also keeps the tabs from going stale when a client changes how remote servers are added: Claude Code documents an `mcp add` subcommand, Codex documents only its config file and its UI, and asking works on both.

Two supporting changes were needed:

- **`FfCommand` host-swap was anchored to the start of the command.** Every command until now was a bare address, so `^` was harmless. These tabs pass a sentence with the address inside it, and the anchored match silently left a self-hosted reader copying `app.flowfuse.com`. Now unanchored and ungreedy.
- **`AgentSetupTabs` step one was hardcoded** for every client that is not Expert. It now takes the same per-client overrides that steps two and three already had.

Neither tab carries a brand mark; both use a lucide terminal glyph, so there is no trademark question to answer. The logo README records that.

The `/ai` coding note drops the two agents that now have their own tab.

**Conflicts with #5799** (the Gemini tab), which inserts into the same array. Whichever merges second needs a one-line rebase; happy to do it.

## Related Issue(s)

https://github.com/FlowFuse/engineering/issues/272

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [x] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5805, opened 2026-09-15
Let the prompt block wrap instead of overflowing the card

dimitrieh:
## Description

The Claude Code and Codex tabs from #5800 put a sentence into a block built for an address, and it overflows the card. Reported from the live page.

The stacked command block sets `white-space: nowrap` on purpose: a wrapped address is harder to read than one you scroll. But `nowrap` sizes the block to its content. Measured on the preview at 1440px:

| Tab | Block width | Position |
| --- | --- | --- |
| Claude (an address) | 291px | 698px inside the card |
| Claude Code (a sentence) | 1079px | **90px past the card's right edge** |

So the end of the prompt, the half that says what it does, was cut off.

Adds an explicit `wrap` prop rather than inferring it from the content's length, and passes it only from the two tabs that carry a prompt. `word-break` goes back to `normal` so words stay whole, overriding the base rule's `break-all`; `overflow-wrap: anywhere` still lets the address break if it has to.

The same defect exists in the platform's own copy of this block and is fixed in https://github.com/FlowFuse/flowfuse/pull/8499.

## Related Issue(s)

https://github.com/FlowFuse/engineering/issues/272

## Checklist


dimitrieh:
Fixes this problem: 
CleanShot 2026-09-15 at 12 14 06 PM@2x

PR #5799, opened 2026-09-14
Add the Gemini tab to the agent connector widget

dimitrieh:
## Description

Adds the Gemini tab to the agent connector widget. `gemini.svg` and its README row were already committed; the client entry was never added, so Gemini was the one mainstream assistant the widget did not offer.

Gemini reaches a custom MCP server through **Settings & help → Connected Apps → Add a custom app**. FlowFuse supports Dynamic Client Registration, so no client ID or secret is needed. Google currently limits this to Gemini Spark on personal accounts in the US, which the step says, the same way the ChatGPT tab already says developer mode needs a paid plan.

One entry covers every surface that embeds `AgentSetupTabs`: `/ai`, the changelog entry, the blog post, and the third-party-agents docs page.

Also repoints the logo README at `AgentSetupTabs.vue`; it still described `src/ai.njk`, which no longer exists on main.

The matching change for the in-platform modal is a separate PR on FlowFuse/flowfuse, which carries its own copy of this widget.

## Related Issue(s)

https://github.com/FlowFuse/engineering/issues/272

## Checklist

 - [x] I have read the [contribution guidelines](https://github.com/FlowFuse/flowfuse/blob/main/CONTRIBUTING.md)
 - [x] I have considered the performance impact of these changes
 - [ ] Suitable unit/system level tests have been added and they pass
 - [x] Documentation has been updated
 - [ ] For blog PRs, an Art Request has been created ([instructions](https://flowfuse.com/handbook/design/art-requests/#creating-an-art-request))

PR #5808, opened 2026-09-15
Revert "Add the Gemini tab to the agent connector widget"

dimitrieh:
Reverts FlowFuse/website#5799


dimitrieh:
Reverting as we have yet to make sure we cater towards gemini more specifically. As it is only available to a subset of their users with gemini spark. When that becomes more clear we can move forward with this.

PR #5812, opened 2026-09-15
ai: show the product in the capability wells

dimitrieh:
## Description

The capability list gives all eleven entries the same schematic. That consistency is
what the drawing was for, and it is also why the section never shows the product: to
see what an approval card actually looks like you have to leave the page.

Eight capabilities that have shipped now play a short clip of themselves in the media
well instead, each taken from the changelog entry that released it. The other three
keep the schematic, because there is nothing honest to show yet: agent blueprints have
never been captured, and the two Connect items are already demonstrated by the full
recording higher up the page.

Treatment matches that recording: autoplaying muted loop, WebM only, no MP4 fallback.
No controls, since these run four seconds rather than fifty-seven. 1.7MB for the whole
set at 720p, which is the ceiling the column width can show. The well switches to 16:9
when it holds a clip, because letterboxing a screen recording into the schematic's
4.8:1 band crops away the interface.

Draft: the clips are the right treatment but several are not yet the right shot for
the capability they sit under. Worth looking at on the preview before this goes
further.

## Related Issue(s)

None.

## Checklist


dimitrieh:
@KristopherLeads conceptual and imagery too big, but curious what you think about showcasing things like that for added dynamism: https://deploy-preview-5812--flowfuse-website.netlify.app/ai/

nuxt/pages/ai/index.vue
```diff
@@ -372,11 +372,38 @@ onUnmounted(() => {
...
+                   Three capabilities keep the schematic, because there is nothing
+                   honest to show yet: agent blueprints have never been captured, and
+                   the two Connect items are demonstrated by the full recording further
+                   up the page rather than by a second clip of the same flow.
+
+                   Same treatment as that recording: autoplaying muted loop, WebM only.
+                   No controls here, unlike the demo - these run 4 seconds rather than
+                   57, so there is nothing to pause. Browsers pause an offscreen muted
+                   autoplay by themselves, so no observer and no preload hint. The
+                   poster is the first paint before playback starts. -->
```
ZJvandeWeg:
```suggestion
```

dimitrieh: Are these automated? This PR is just to showcase a concept and not yet done by any means.

ZJvandeWeg: No, I hate AI comments, so let's remove the, :)

PR #5815, opened 2026-09-16
dev proxy: register the migrated sections so local dev serves them from Nuxt

dimitrieh:
## Description

`nuxt/server/middleware/legacy.ts` decides, per request, whether Nuxt answers or the request is proxied to the 11ty dev server. It is dev-only (`if (process.env.NODE_ENV !== 'development') return`), so nothing in CI or a Netlify deploy preview exercises it.

That is why this drifted: eight migrated sections were never added to it, and every reviewer checking a deploy preview saw a correct page while `npm run dev` kept serving the old 11ty one.

Registers `/use-cases`, `/industries`, `/platform`, `/partners`, `/events`, `/landing`, `/blueprints` and `/node-red`.

Three of them keep assets in the 11ty tree that only reach `nuxt/public` through the passthrough during a production build, so they need the carve-out `/vs` and `/resources` already have: `/events/images`, `/landing/images` and `/whitepaper/images`, plus the one `.ics` the Hannover page links. The `/whitepaper/images` one is what makes the whitepaper covers render on the use-case pages locally.

`/node-red` becomes a plain prefix now that its index is a Nuxt page, retiring the carve-out that used to keep just that one path on 11ty.

**Merge this last.** Registering a section before its migration PR lands points dev at a Nuxt page that does not exist yet. `/industries` is the live example: `nuxt/pages/industries/automotive.vue` is already on main while the other industry pages are still 11ty-only, so this is only correct once #5781 has merged. Same for the other seven.

No test: `legacy.ts` is deleted by #5795 in this same epic, so a regression test for it would be born dead.

Verified by simulating the middleware's decision against the real constants for 20 routes, including that the `/whitepaper/images` carve-out does not swallow real whitepaper pages, and that `/about` still falls through to 11ty since #5791 registers that itself.

## Related Issue(s)

https://github.com/FlowFuse/website/issues/5777

## Checklist


Yndira-E:
@dimitrieh I think this should be addressed individually on each pr you opened for migrating each route.

dimitrieh:
Agreed, closing. Moving each route's registration into its own migration PR.

One sequencing note that motivated collecting them, so it is not lost: registering a section only works once that section's pages exist, so each edit belongs with the PR that adds them, which is what you are asking for. The flip side is that all eight touch the same two lines of `legacy.ts`, so expect the rebase-after-each-merge the issue already calls out.

Carrying over with them: `/events`, `/landing` and `/whitepaper` keep assets in `src/` that only reach `nuxt/public` via the 11ty passthrough in a production build, so those three need the dev carve-out `/vs` and `/resources` already have. The `/whitepaper/images/` one is what makes the whitepaper covers render locally on #5780.

